PR #7, opened 2016-04-01
Update WebServer.cs

huenato:
My suggestion is to change "toscdn" to "toslive" path name as it in the actual client.xml file.


exectails:
It would certainly make the replacement a little easier. I see no reason not to do it. Thanks.

PR #10, opened 2016-04-01
CZ_REQ_NORMAL_TX_NUMARG

S-anasol:
unkShort is known


exectails:
Oh~ good to know. Haven't worked on skill point distribution yet. Thanks.

PR #18, opened 2016-04-03
Skill learn very base implement

S-anasol:
Cant handle tabs correctly in VS(default settings)
So few lines changed tabulation :(

Added:
* skills.txt - database with some dummy skills
* db table with learned skills
* loading/saving skill list for character in DB.
* skill learning in-game.
* modified ZC_ABILITY_LIST to read "real" skills data, but i have no research experience to get it work with client correctly(properly display learned skills in client).

P.S.
Still not a C# dev. 
С# experience is a few hours of copypasting :+1:

### DIFF
--- a/src/ChannelServer/Network/ChannelPacketHandler.cs
+++ b/src/ChannelServer/Network/ChannelPacketHandler.cs
@@ -806,49 +806,72 @@ namespace Melia.Channel.Network
 		public void CZ_REQ_NORMAL_TX_NUMARG(ChannelConnection conn, Packet packet)
 		{
 			var size = packet.GetShort();
-			var type = packet.GetShort(); // 1 - statpints, 2 - skillpoints
-			var count = packet.GetInt();
-			//Throws exception when trying distributing skill points (count 6)
-			if (count != 5)
-				throw new Exception("Unknown CZ_REQ_NORMAL_TX_NUMARG format, expected 5 stats.");
+			var type = packet.GetShort(); // 1 - statponts, 2 - skillpoints
+            var count = packet.GetInt();
 
-			var character = conn.SelectedCharacter;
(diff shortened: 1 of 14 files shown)


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -806,49 +806,72 @@ public void CZ_CLIENT_HIT_LIST(ChannelConnection conn, Packet packet)
 		public void CZ_REQ_NORMAL_TX_NUMARG(ChannelConnection conn, Packet packet)
 		{
 			var size = packet.GetShort();
-			var type = packet.GetShort(); // 1 - statpints, 2 - skillpoints
-			var count = packet.GetInt();
-			//Throws exception when trying distributing skill points (count 6)
-			if (count != 5)
-				throw new Exception("Unknown CZ_REQ_NORMAL_TX_NUMARG format, expected 5 stats.");
```
epicmiro: Not sure if you should throw new exceptions vs debug logging. Haven't looked trough aura source lately to confirm though.. @exectails ?

src/Shared/Data/Database/Skills.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Aura development team - Licensed under GNU GPL
+// For more information, see license file in the main folder
```
epicmiro: Skill and Skills cause confusion. Does melia also use SkillManager like Aura or is there no such system present?

epicmiro:
Use tabs instead of spaces, Aura projects follow this indentation rule.

src/Shared/Data/Database/Skills.cs
S-anasol: There are no skills system, @exectails didn't code it yet :)

Issue #13, opened 2016-04-02
Closing dialog with Escape


exectails:
When pressing escape, an open dialog should be closed. Right now we're resuming and continuing the dialog, which is not official behavior and can get the player stuck in the dialog until relog.

The unknown int in `CZ_DIALOG_ACK` seems to indicate what happened, with 0 or -1 being sent when Escape was pressed and 1 otherwise.

exectails:
I'm not sure how to do this right now, as simply closing the dialog on the client when pressing escape doesn't stop the yielding Lua thread. If you talk to the NPC again it will continue where it left off. We have to cancel the yield. We could create a new thread for the client to use, but then we have yield hanging around. I don't know if Lua would collect that.

exectails:
A potential solution could be to not use lua_yield anymore, but C#'s async features, to suspend the operation from within the ScriptManager, similar to how we did stateless NPCs in Aura.

https://github.com/aura-project/aura/blob/master/src/ChannelServer/Scripting/Scripts/NpcScript.cs#L2048

That way we don't leave the suspending function and might be able to stop the execution, and be it by returning an error.

exectails:
I did some research on how the Lua threads work and figured out how to solve all my Lua problems. When a new thread is created with `lua_newthread` it's put on the global state's stack. It stays there until it's removed, and only after that it's collected. This means we simply have to remove the thread from the stack to clean everything up.

For more flexibility I also changed the way threads are created, so it's one thread per *dialog sessions*, instead of play session.

PR #19, opened 2016-04-03
Close dialog with escape #13

S-anasol:
Now escape properly closes dialog with npc - #13 

And tabs again rewrited

### DIFF
--- a/src/ChannelServer/Network/ChannelPacketHandler.cs
+++ b/src/ChannelServer/Network/ChannelPacketHandler.cs
@@ -678,10 +678,10 @@ namespace Melia.Channel.Network
 		[PacketHandler(Op.CZ_DIALOG_ACK)]
 		public void CZ_DIALOG_ACK(ChannelConnection conn, Packet packet)
 		{
-			var unkInt = packet.GetInt();
+			var type = packet.GetInt(); // 1 - Close from dialog, 0 - Close with escape
 
-			// Check state
-			if (conn.ScriptState.CurrentNpc == null)
+            // Check state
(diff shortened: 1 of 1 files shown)


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -678,10 +678,10 @@ public void CZ_DIALOG_SELECT(ChannelConnection conn, Packet packet)
 		[PacketHandler(Op.CZ_DIALOG_ACK)]
 		public void CZ_DIALOG_ACK(ChannelConnection conn, Packet packet)
 		{
-			var unkInt = packet.GetInt();
+			var type = packet.GetInt(); // 1 - Close from dialog, 0 - Close with escape
```
exectails: -1 is another type I've seen, when closing normal dialogues with escape.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -678,10 +678,10 @@ public void CZ_DIALOG_SELECT(ChannelConnection conn, Packet packet)
...
 
-			// Check state
-			if (conn.ScriptState.CurrentNpc == null)
+            // Check state
```
exectails: Please set your IDE to use tabs, as outlined in our [coding conventions](https://github.com/aura-project/melia#coding-conventions).

exectails:
I'm afraid this solution won't work. If you just close the dialog, without handling the still yielding Lua thread, the NPC will continue where it left off when you pressed escape. We have to cancel the running function.

PR #20, opened 2016-04-03
HandleItem has been improved and now you can set amounts of items

vegax87:
Usage: /item ItemID amount

to add silver coins: /item 900011 amount

there's a bug with the stack of silver coins: everytime I add them, they generate a new ItemUniqueId in the database instead of updating the values

### DIFF
--- a/src/ChannelServer/Util/GmCommands.cs
+++ b/src/ChannelServer/Util/GmCommands.cs
@@ -260,24 +260,43 @@ namespace Melia.Channel.Util
 
 		private CommandResult HandleItem(ChannelConnection conn, Character character, Character target, string command, string[] args)
 		{
-			if (args.Length < 3)
+			if(args.Length < 2)
 				return CommandResult.InvalidArgument;
-
+			
 			int itemId;
 			int amount;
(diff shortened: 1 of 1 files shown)


exectails:
>there's a bug with the stack of silver coins

Meh. More a missing feature than a bug. I knew I forgot something, I meant to add auto stacking on adding before release.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -260,11 +260,12 @@ private CommandResult HandleWarp(ChannelConnection conn, Character character, Ch
 
 		private CommandResult HandleItem(ChannelConnection conn, Character character, Character target, string command, string[] args)
 		{
-			if (args.Length < 2)
+			if (args.Length < 3)
```
exectails: Why not make the amount parameter optional? That way you don't need it every time you want to spawn something, especially for things like equipment, that will always be 1.

gastonb: Amount should be optional, and default to 1.

vegax87:
I fixed it

exectails:
I've optimized your code a little, take a look.

PR #21, opened 2016-04-03
	Adding packets for Skills handling and basic skills files structure

gastonb:
Adding packets for Skills handling (ground skills) and basic skills files structure


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -836,5 +870,71 @@ public void CZ_CHECK_PING(ChannelConnection conn, Packet packet)
...
+			Log.Warning("Pos2: {0},{1},{2}", x2, y2, z2);
+			Log.Warning("Dir: {0},{1}", cos, sin);
+			Log.Warning("unk3: {0}", unk3);
+			Log.Warning("unk6: {0}", unk6);
+			Log.Warning("unk7: {0}", unk7);
+
+			var character = conn.SelectedCharacter;
+
+			// Check parameters
+			var packetPosition1 = new Shared.World.Position(x1, y1, z1);
```
exectails: You don't need to include the full namespace every time, add a using for Shared.World instead.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1510,7 +1510,239 @@ public static void ZC_PC_MOVE_STOP(Character character, Position position, Direc
...
+		{
+			var packet = new Packet(Op.ZC_NORMAL);
+			packet.PutInt(0x1c);
+			packet.PutByte(0);
+			packet.PutBinFromHex("9F D2 42 0B");
+			packet.PutInt(id); // Used to be skillId, but observed 00 00 00 00 now
+			packet.PutFloat(position.X);
+			packet.PutFloat(position.Y);
+			packet.PutFloat(position.Z);
+			packet.PutFloat(0.707f);
```
exectails: Why don't you use the direction if you pass it?

src/ChannelServer/Skills/Skill.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// Returns cast time of skill, specific for its creature.
+        /// </summary>
+        /// <returns></returns>
+        public int GetCastTime()
+        {
+            return 1;
+        }
+    }
+
+    /// <summary>
```
exectails: Spaces =<

src/ChannelServer/Skills/Skill.cs
```diff
@@ -0,0 +1,101 @@
...
+	/// Current state of a skill.
+	/// </summary>
+	public enum SkillState
```
exectails: Are you copy-pasting Aura? =<

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -488,7 +489,16 @@ private CommandResult HandleClearInventory(ChannelConnection conn, Character sen
 
 			return CommandResult.Okay;
 		}
-	}
+
+		private CommandResult HandleTestHealSkill(ChannelConnection conn, Character sender, Character target, string command, string[] args)
```
exectails: I think we shouldn't add a test command for every little thing, better use the test command that's already there, and don't commit the changes to it.

gastonb: I left it there, to test Ground skills. I could change that to TestGroundSkill() in the future.

src/ChannelServer/Skills/Skill.cs
gastonb: I based the structure of Skills on Aura Project, modifying as needed. Most of it is an empty skeleton for later use. Once we know how and where to store data.

SkillInfo, SkillData, SkillStates are all things that we will definitelly use.

src/ChannelServer/Skills/Skill.cs
gastonb: I realized VS2015 was set to "spaces" right before doing the commit. I changed to TABS and modified all my code to TABS. If anything is using spaces, is unintended.

src/ChannelServer/Network/ChannelPacketSender.cs
gastonb: Direction is in the parameters, because the packet will need it. Line 1617 and 1618. I used constants (0.707f) for testing purposes. We don't know the purpose of that packet yet, and don't know what direction to send from the function calling it.

src/ChannelServer/Network/ChannelPacketHandler.cs
gastonb: I will do a commit with that change. Thanks.

src/ChannelServer/Network/ChannelPacketSender.cs
gastonb: I will add Position use , but commented out, for future reference for other developers.

src/ChannelServer/Skills/Skill.cs
gastonb: Fixed in the next commit.

exectails: Use auto-formatting, makes your life easier.

src/ChannelServer/Skills/Skill.cs
exectails: Maybe, though adding things like Used and Completed, which we most likely won't need, is potentially confusing for others reading the code. And the data class would belong in Shared. What I'm trying to say is, I think we should keep unnecessary test or dummy code on master to a minimum, to keep the source relatively clean and to reduce the amount of clean up we have to do later on. In my opinion this would be absolutely fine in a branch, but not on master.

src/ChannelServer/Network/ChannelPacketSender.cs
exectails: >don't know what direction to send from the function calling it

How about a direction composed of those two values you're currently hard-coding =P

gastonb: Ok. I will do those changes. Thanks for your suggestion.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -860,5 +894,75 @@ public void CZ_CHECK_PING(ChannelConnection conn, Packet packet)
...
+			var x2 = packet.GetFloat();
+			var y2 = packet.GetFloat();
+			var z2 = packet.GetFloat();
+			var cos = packet.GetFloat();
+			var sin = packet.GetFloat();
+			var unk3 = packet.GetInt(); // This seems to be "target actorId"
+			var unk6 = packet.GetByte();
+			var unk7 = packet.GetByte();
+
+			Log.Warning("CZ_SKILL_GROUND");
```
exectails: Same as above. I realize I haven't specified this anywhere for Melia yet, master is supposed to be the clean and stable branch. There should be no unnecessary testing code, or random warnings that don't mean anything to the user. To work on unfinished features in peace, until they're ready, we should utilize branches.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -806,48 +807,81 @@ public void CZ_CLIENT_HIT_LIST(ChannelConnection conn, Packet packet)
...
+				case 1:
+					var count = packet.GetInt();
+					if (count != 5)
+						throw new Exception("Unknown CZ_REQ_NORMAL_TX_NUMARG format, expected 5 stats.");
+
+					for (int i = 0; i < count; ++i)
+					{
+						var stat = packet.GetInt();
+
+						Log.Warning("stat{1}: {0}", stat, i);
```
exectails: A warning? What are you warning the user about? o.o If you want to log something for debug purposes you should use Debug, which is only visible in the console if you enable it in the log.conf.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1275,11 +1275,12 @@ public static void ZC_EXP_UP_BY_MONSTER(Character character, int exp, int jobExp
 		/// </summary>
 		/// <param name="character"></param>
 		/// <param name="exp"></param>
-		public static void ZC_EXP_UP(Character character, int exp)
+		/// <param name="jobExperience"></param>
+		public static void ZC_EXP_UP(Character character, int exp, int jobExperience)
```
exectails: Just a pointer, but this would've made a nice separate commit, now it's kinda hidden in between this:

ss 2016-04-06 at 09 55 03

PR #23, opened 2016-04-04
Item auto stacking

S-anasol:
PR #20


exectails:
Instead of looking for one item stack where the new stack fits in, I think we should rather look through all available stacks and fill them up as we go. For example, if you have 2 stacks that are both at 5 and the max stack is 10, and you add a stack with 6 to the inventory, you wouldn't want a third stack, but for the first to be filled up, and the one remaining to be added to the second, resulting in one 10 and one 6 stack.

Though adding it to AddSilent is good, it being used by the database, that way we automatically combine stacks when the character is loaded.

S-anasol:
yeah i thought about this variant, but was not sure if it possible =)

Updating src now, will try to fix.

>automatically combine stacks when loaded

yeah it was first working part of code :dancer:

exectails:
>but was not sure if it possible

It does complicate it a little, as you have to update an unknown amount of items. You might need to refactor Add and AddSilent a little.

S-anasol:
possible - to do in-game i mean.

In code it ofc possible, i already have idea how it can be done(+/-), thought about it yesterday before sleep :sleepy:

S-anasol:
Done :)

src/ChannelServer/World/Inventory.cs
```diff
@@ -194,19 +204,71 @@ public Item GetItem(EquipSlot slot)
...
 		/// <param name="item"></param>
 		/// <return>Index of the item.</return>
-		public int AddSilent(Item item)
+		public List<object> AddSilent(Item item)
 		{
 			var cat = item.Data.Category;
-
 			lock (_syncLock)
 			{
+				List<object> items_updated = new List<object>();
```
exectails: Why did you make this a list of objects? If you want to return something special you should make a dedicated type for it.

src/ChannelServer/World/Inventory.cs
```diff
@@ -194,19 +204,71 @@ public Item GetItem(EquipSlot slot)
...
 
-				_items[cat].Add(item);
-				_itemsWorldIndex[item.WorldId] = item;
+				var subIndex = -1;
+				var expected_amount = item.Amount;
+
+				if (item.Data.MaxStack > 1 && this.HasItem(item.Id))
+				{
+					int item_index = subIndex;
+					while (expected_amount > 0)
```
exectails: This entire loop is much more redundant and complicated than it needs to be, maybe you want to reevaluate it, see if you can improve it a little? Note the repeating lines in both cases of the if/else for example.

src/ChannelServer/World/Inventory.cs
```diff
@@ -194,19 +204,71 @@ public Item GetItem(EquipSlot slot)
...
+					{
+						long inventory_index = this.GetStackableItemIndex(cat, item.Id, item.Amount);
```
exectails: Why don't you return an int right away, instead of a long that you have to cast to an int before using it?

src/ChannelServer/World/Inventory.cs
```diff
@@ -194,19 +204,71 @@ public Item GetItem(EquipSlot slot)
...
+								possible_add_to_current_stack = expected_amount;
+								_items[cat][subIndex].Amount += possible_add_to_current_stack;
+								_itemsWorldIndex[_items[cat][subIndex].WorldId] = _items[cat][subIndex];
+
+								expected_amount -= possible_add_to_current_stack;
+								item.Amount -= possible_add_to_current_stack;
+							}
+
+							item_index = subIndex;
+							items_updated.Add(new { index = item_index, diff = possible_add_to_current_stack });
```
exectails: Since you have two values, an index and a diff for the index, you basically have a key/value relation, which you could return as a Dictionary.

src/ChannelServer/World/Inventory.cs
```diff
@@ -217,12 +279,30 @@ public int AddSilent(Item item)
 		/// <return>Index of the item.</return>
 		public int Add(Item item, InventoryAddType addType)
 		{
-			var index = this.AddSilent(item);
+			List<object> indexes = this.AddSilent(item);
+
+			object last_item = indexes.Last();
+			int last_index = (int)last_item.GetType().GetProperty("index").GetValue(last_item);
```
exectails: Using reflection to access values is very inefficient. This will solve itself when you don't use anonymous objects anymore and can directly access the properties.

exectails:
Please refer to our [coding conventions](https://github.com/aura-project/melia#coding-conventions) in regards to naming conventions for variables. You should be using "camelCase" consistently.

S-anasol:
updated codestyle and remove some code duplication.

Hope it good now :)

src/ChannelServer/World/Inventory.cs
```diff
@@ -209,19 +220,62 @@ public Item GetItem(EquipSlot slot)
...
 		/// <param name="item"></param>
 		/// <return>Index of the item.</return>
-		public int AddSilent(Item item)
+		public Dictionary<int, int> AddSilent(Item item)
 		{
 			var cat = item.Data.Category;
-
 			lock (_syncLock)
 			{
+				Dictionary<int, int> itemsUpdated = new Dictionary<int, int>();
```
exectails: Just a tip, if you use the [var keyword](https://msdn.microsoft.com/en-us/library/bb383973.aspx), you don't need to type as much, the compiler tries to figure out the type itself then, based on what you put in. E.g. `var itemsUpdated = new Dictionary<int, int>();`

src/ChannelServer/World/Inventory.cs
```diff
@@ -209,19 +220,62 @@ public Item GetItem(EquipSlot slot)
...
+
+				if (item.Data.MaxStack > 1 && this.HasItem(item.Id))
+				{
+					int itemIndex = inventoryIndex;
+					while (expectedAmount > 0)
+					{
+						inventoryIndex = this.GetStackableItemIndex(cat, item.Id, item.Amount);
+						if (inventoryIndex != -1 && _items[cat].Count > inventoryIndex)
+						{
+							var spaceInStack = _items[cat][inventoryIndex].Data.MaxStack - _items[cat][inventoryIndex].Amount;
```
exectails: Another tip, if you create a local variable for that item, the code becomes a little easier to read, and should you ever need to change something, you have less to do.

E.g. `var spaceInStack = item.Data.MaxStack - item.Amount;`
You see what's going on in an instant, without having to read the indices for the lists, to find out what you're working with.

exectails:
There's one more redundancy that can be removed, can you find it?^^

src/ChannelServer/World/Inventory.cs
S-anasol: just stupid rules xD

Firstly i write 
Dictionary<int, int> itemsUpdated;

but then VS scream on me that it want initializtion.

S-anasol:
forgot one thing.
Will fix, dont merge :)

S-anasol:
Now it work properly :+1:

src/ChannelServer/World/Inventory.cs
```diff
@@ -232,12 +271,30 @@ public int AddSilent(Item item)
 		/// <return>Index of the item.</return>
 		public int Add(Item item, InventoryAddType addType)
 		{
-			var index = this.AddSilent(item);
+			Dictionary<int, int> indexes = this.AddSilent(item);
+
+			var lastItem = indexes.Last();
```
exectails: Another tip: you can save one line here. `var lastIndex = indexes.Values.Last();`

src/ChannelServer/World/Inventory.cs
```diff
@@ -232,12 +271,30 @@ public int AddSilent(Item item)
...
+			int lastIndex = lastItem.Key;
+
+			foreach (var itemObject in indexes)
+			{
+				int index = itemObject.Key;
+				int diff = itemObject.Value;
+				
+				if (index == lastIndex)
+				{
+					addType = InventoryAddType.PickUp;
```
exectails: Hm... you're overriding the parameter here. If you `Add` something with a certain add type, it shouldn't use PickUp for the last, but whatever you put in.

S-anasol: If i add not stackable item it should be new always - yes.
If add some pieces to stack it shouldn't be new - no.
But last stack can be new sometime and always new for nonstackable.

Not sure how it should be.

exectails: I'm not sure you understood what I meant, say I call `character.Inventory.Add(item, InventoryAddType.Something)` from somewhere. I specified that add type, because that, presumably, affects how the client displays the add, and I want to use `Something`. But you aren't using the value I pass to function, you override it.

What you should do is switch to NotNew if it actually is not new, or use what was passed to the function. Otherwise the parameter would be completely pointless and could be removed.

exectails:
I did some work on the inventory myself yesterday and had to update items. I realized they use ZC_ITEM_REMOVE for that, not ADD. While ADD with a negative value *works*, we should probably use REMOVE, if we want to *remove* something.

exectails:
@S-anasol Are you going to finish this PR, or should I take over?

exectails:
I tried to use your PR as base, but there were several issues and I ended up recoding it from scratch. A few things I noticed:

- In `GetStackableItemIndex` you're potentially creating hundreds of anonymous objects just to find the indices, which causes a lot of garbage to be collected, as adding an item to the inventory is a very frequent task.
- `Add` doesn't use the specified add type, as mentioned before, but worse, it decides on an add type based on the dictionary returned by `AddSilent`. Since the order of the elements in a dictionary isn't guaranteed, you can't be sure the result of `Last` will actually be the last item.
- `GetStackableItemIndex` was missing a lock.

I'll close this in favor of #52, which is a little cleaner.

PR #24, opened 2016-04-04
Hirasu's NPC pack + 574 warps

vegax87:
- I manually fixed the paths and put them inside npc folder

- UPDATE: Added 574 warps and fixed some wrong spawns

- Note: some of them are related to quests and world events


exectails:
Hm... the NPC names, don't they use localization? I see them in the localization files, shouldn't we use those, @Hirasu? If someone wants to use a different language for Melia, the names should ideally get localized.

vegax87:
Theorically you need to write their name everytime...

```
private int addnpc(IntPtr L)
		{
			var monsterId = Melua.luaL_checkinteger(L, 1);
			**var name = Melua.luaL_checkstring(L, 2);**
			var mapName = Melua.luaL_checkstring(L, 3);
			var x = (float)Melua.luaL_checknumber(L, 4);
			var y = (float)Melua.luaL_checknumber(L, 5);
			var z = (float)Melua.luaL_checknumber(L, 6);
			var direction = Melua.luaL_checkinteger(L, 7);
			var dialog = Melua.luaL_checkstring(L, 8);
```

exectails:
>Theorically you need to write their name everytime...

What do you mean? You do need to specify a name to send to the client of course, but I assume that name can be a localization key, just like we do for warps. That way the client handles the localization.

toscompendium:
@vegax87 Though I believe you missed @exectails whole point...

**For Ex**:
```
--[NPC]
addnpc(20115,"[Item Merchant]{nl} Mirina","c_Klaipe",510.7029,-1.292879,-349.3194,45,"DummyCall")
```

**Is the english NPC name**:
`ETC_20150317_009195	[Item Merchant]{nl}         Mirina`

**the korean NPC name is**:
`ETC_20150317_009195		[도구 상인]{nl}    미리나`

if we single out a single translation of an item, monster, NPC, quest etc etc, then we lock this project down to a single client and it becomes UNinternational. There has to be another area for translation an NPC should be called by it's CODE NAME **ETC_20150317_009195**, which then refers to a table that translates it from the servers running translation tables.

https://github.com/Treeofsavior/EnglishTranslation		_<name>_**_kor.tsv** are the master korean files.
https://github.com/Treeofsavior/ThaiTranslation

vegax87:
Ok understood, in the future we could use "LocalKey" just like warp points now...

anyway I've added almost all warps, keep in mind that some NPC and warps are directly related to quest events and world events, so don't worry if you find some blocked road or petrified NPC...

Hirasu:
@toscompendium 
@exectails 
@vegax87 
NPCPack with localkey https://github.com/Hirasu/melia/commit/0b05a3ec1ee88aa2929b681193b3d15201666ef1

PR #28, opened 2016-04-06
/g is reserved for guild chat, so I renamed /go to /tele

vegax87:
Before you were forced to write //go YourName <destination> because this GM command conflicted with the guild chat command, now it's more comfortable...

### DIFF
--- a/src/ChannelServer/Util/GmCommands.cs
+++ b/src/ChannelServer/Util/GmCommands.cs
@@ -49,7 +49,7 @@ namespace Melia.Channel.Util
 			Add("speed", "<speed>", HandleSpeed);
 			Add("iteminfo", "<name>", HandleItemInfo);
 			Add("monsterinfo", "<name>", HandleMonsterInfo);
-			Add("go", "<destination>", HandleGo);
+			Add("tele", "<destination>", HandleTele);
 			Add("clearinv", "", HandleClearInventory);
 
 			// Dev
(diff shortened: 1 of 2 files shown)


src/ChannelServer/Util/GmCommands.cs
```diff
@@ -49,7 +49,7 @@ public GmCommands()
 			Add("speed", "<speed>", HandleSpeed);
 			Add("iteminfo", "<name>", HandleItemInfo);
 			Add("monsterinfo", "<name>", HandleMonsterInfo);
-			Add("go", "<destination>", HandleGo);
+			Add("tele", "<destination>", HandleTele);
```
Cosrnos: I think it'd be preferred that this was renamed to warp, just to be consistent with other aura projects ;)

vegax87:
there's also /warp but you need to specify the exact name of the region (i.e: f_orchard_32_3)

src/ChannelServer/Util/GmCommands.cs
exectails: @Cosrnos Aura has a go command as well. Warp takes map and coordinates, while go warps to pre-defined destinations. E.g. `>go dun`/`/go c_Klaipe`

Cosrnos:
Ahh my bad. Maybe /jump then? I'm just wondering if we could remove some of the ambiguity. I feel tele is a little counterintuitive but that's definitely a personal opinion.

exectails:
@Cosrnos jump is the GM command to teleport "in-map", on Aura and Melia xD

Cosrnos:
WELP I don't know then. Tele might be the best name for it. or /to \*gestures emphatically*

exectails:
lol. I'm with you though, tele isn't the nicest name ever. I see a whole different problem here though, we can't use any command names that start with "g" (guild) or "p" (party), maybe more. Maybe we should use a different default command prefix?

Cosrnos:
Good point. I think changing the prefix would be the correct direction to go since we'll need to avoid this problem in the future. using > or ! as a prefix would solve all the problems!

vegax87:
maybe the good old @ prefix? :D

exectails:
Personally I'm not a big fan of @, because the key combination you need for it on a German keyboard is a little bothersome. On Aura we're using `>`, maybe some consistency would be nice, for people who switch back and forth between Melia and Aura?

Hirasu:
Maybe make it changeable in the command.conf file like:  CommadPrefix: "@"  or  CommadPrefix: "/"

PR #32, opened 2016-04-07
Basic AI movement + MonsterBuilder

rincew1nd:
Here is wonky AI movement + monster builder.


src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1472,32 +1472,57 @@ public static void ZC_MOVE_DIR(Character character, float x, float y, float z, f
...
+        /// <param name="monster"></param>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <param name="z"></param>
+        public static void ZC_MOVE_PATH(Monster monster, float x, float y, float z)
+        {
+            var packet = new Packet(Op.ZC_MOVE_PATH);
+
+            packet.PutInt(monster.Handle);
+            packet.PutFloat(monster.Position.X);
```
Tachiorz: There should be ints not floats

rincew1nd: ZC_MOVE_PATH contains move vector with coordinates described by int variables, not float. Or am i wrong? Note: Also it contains 2 unknown float after path (first seems to be speed, and second always zero)
Prove:
http://i.imgur.com/iGkvGYD.png
http://mabinogi.pp.ua/g/8d127679d00830f32385bceca3229616.png

src/ChannelServer/World/MonsterBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+using Melia.Shared.World;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Melia.Channel.World
+{
+    class MonsterBuilder
```
exectails: What are you trying to achieve with this class? Why not simply use the Monster class?

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -432,24 +432,24 @@ public static void ZC_ABILITY_LIST(Character character)
...
+			packet.PutFloat(entity.Speed);
 			packet.PutFloat(0);
 
 			// [i11257 (2016-03-25)]
 			{
 				packet.PutByte(0);
 			}
 
-			character.Map.Broadcast(packet, character);
+            entity.Map.Broadcast(packet, entity);
```
exectails: Please set your IDE to use tabs, as outlined in our coding conventions, which can be found in the [readme](https://github.com/aura-project/melia#coding-conventions).

src/ChannelServer/World/MonsterBuilder.cs
rincew1nd: Builders are easier to use. The way it's done now need from you to specify all major values manually to object, but builders will done it by themselfs with better visualization.

src/ChannelServer/Network/ChannelPacketSender.cs
rincew1nd: My bad. Something goes wrong while copy pasting.

src/ChannelServer/World/MonsterBuilder.cs
exectails: That is highly debatable in my opinion. Personally I dislike factories/builders and I don't see the appeal. What you're doing here can be done with the actual object just as easy, and constructors were *invented* to initialize classes with already available information. It's like you're moving the constructor to its own class. Not only that, you're basically creating a copy of `Monster`, forcing us to maintain one class for the price of two.

I think this is terribly inefficient.

src/ChannelServer/Scripting/ScriptManager.cs
```diff
@@ -413,7 +413,7 @@ private int logdebug(IntPtr L)
 		/// <remarks>
 		/// Parameters:
 		/// - int monsterId
-		/// - string name / Localkey
```
exectails: Why did you remove this? It documents that you can use either a name or a localization key.

src/ChannelServer/Scripting/ScriptManager.cs
```diff
@@ -443,11 +443,13 @@ private int addnpc(IntPtr L)
 			if (name.StartsWith("ETC_") || name.StartsWith("QUEST_"))
 				name = "@dicID_^*$" + name + "$*^";
 
-			var monster = new Monster(monsterId, NpcType.NPC);
```
exectails: There's nothing wrong with the way you initiate monster here and further down, but you might want to refrain from changing things just so they're to *your* liking. It creates unnecessary diffs, and this being an open-source project not everything will always be like you would normally do it.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -42,7 +42,7 @@ public GmCommands()
 			Add("jump", "<x> <y> <z>", HandleJump);
 			Add("warp", "<map id> <x> <y> <z>", HandleWarp);
 			Add("item", "<item id> [amount]", HandleItem);
-			Add("spawn", "<monster id>", HandleSpawn);
+			Add("spawn", "<monster id> <x> <y> <z>", HandleSpawn);
```
exectails: What do you believe we can achieve by letting users specify a position?

The command manager currently can only handle indexed parameters, which makes them a rare commodity, because each new one has an impact on the usability and complicates the parsing. I'd suppose there aren't many who have a need to spawn a monster at a specific position, usually you'd want to spawn it where *you* are, or at the position of someone else, both of which can be achieved without a position. In my opinion there are better uses for the parameters.

On a side-note, the parameters you added are optional, you should use squared brackets to document that, see the item command for an example.

src/ChannelServer/World/Entity.cs
```diff
@@ -10,5 +10,7 @@ public interface IEntity
 		int Hp { get; }
 		Map Map { get; }
 		Position Position { get; }
+		Direction Direction { get; }
+		float Speed { get; }
```
exectails: My original thought was to use Speed for the base speed of an entity, while GetSpeed, like in Character provides the actual Speed, factoring in buffs and the like. (A separate method to indicate that something is calculated on the fly.) If you change that, you should remove the GetSpeed method entirely.

src/ChannelServer/World/Monster.cs
```diff
@@ -149,5 +228,157 @@ public void Kill(Character killer)
...
+		
+		/// <summary>
+		/// Sets direction toward target and updates clients.
+		/// </summary>
+		public void RotateTowardTarget()
+		{
+			if (this.Hp <= 0) return;
+
+			if (this.Target == null) return;
+			this.SetDirection(
```
exectails: Feel free not to break non-comment lines at 80, as outlined in our coding conventions.

src/ChannelServer/World/Monster.cs
```diff
@@ -149,5 +228,157 @@ public void Kill(Character killer)
...
+				Shared.Util.Math2.AngleBetweenTwoEntity(
```
exectails: Note: Personally I believe a `using` increases the readability significantly, because you don't need the full name. Even if it's only for one two instances.

src/ChannelServer/World/Monster.cs
```diff
@@ -149,5 +228,157 @@ public void Kill(Character killer)
...
+
+		/// <summary>
+		/// Attack target entity if monster is agred
+		/// </summary>
+		public void AttackTarget()
+		{
+			if (this.Target != null && this.Hp > 0 && this.IsInRangeOfAttack)
+			{
+				//this.Target.Hp -= this.Attack;
+				RotateTowardTarget();
```
exectails: Please use `this.`, as outlined in our coding conventions.

src/Shared/Util/Math2.cs
```diff
@@ -70,5 +73,45 @@ public static bool Between(int val, int min, int max)
...
 			return (val >= min && val <= max);
 		}
+
+		/// <summary>
+		/// Calculate rotation toward target
+		/// </summary>
+		/// <param name="source"></param>
+		/// <param name="target"></param>
+		/// <returns></returns>
+		public static Direction AngleBetweenTwoEntity(Position source, Position target)
```
exectails: I'd propose to move this to Position or Direction, as it's not a general use Math function, but something very specific to the game. Same for VectorTwoEntity.

rincew1nd:
Right now there is no point in changing Speed variable and GetSpeed methods in class.
@gastonb has make some classes for movable NPC, i believe. Need to wait a little before he will merge it to original repository.

exectails:
>Right now there is no point in changing Speed variable and GetSpeed methods in class.

There's no need to change speed? I use the speed command constantly, and you don't know when someone might want to add buffs and the like^^ My point was, it was there and you removed it only partially, which might cause confusion about the difference between Speed and GetSpeed.

rincew1nd:
I removed Speed from IEntity, cause not all entity can move at all. For that reason i added speed variable and method to monster class.
For movable entity we need to have another interface with have speed, @gastonb done this in his repository.

Issue #4, opened 2016-04-01
Combat (Range)


exectails:
Getting combat to work will basically be the most important thing for TOS, as it's the defining feature of the game. Unfortunately there is one big problem with combat in Melia right now, the client attacks whatever monster is closest on the screen. It doesn't limit the range and it only ever attacks one enemy.

After the iCBT2 I've spent quite a few hours trying to figure this out, but I didn't get very far. The client obviously decides what's in range to be hit, and "hacking" videos during the betas confirm that the server didn't really check what the client sent, but there must be something that limits the client during normal operation.

I would expect there to be some value in some packet, or maybe a property on the monsters, the characters, or the items, but I just don't know what the deciding factor is yet, as nothing jumped out at me during my research.

vegax87:
Take a look at wizard packets posted by Ironheart: http://aura-project.org/forum/topic/221-mage-packets/  you can find ranged skills with different AoE Ratios
for example, Energy Bolt has AoE Ratio 1 while Earthquake (exactly like Swordman's normal attacks) has AoE Ratio 4, that value must be there...

S-anasol:
What problem with this on server?
You can easy check distance to deny/allow hit.

I did someting like: https://gist.github.com/S-anasol/d30517059fe350c4a1821f795a172d13
Ofc its proof-of-concept. But it work)))
I can hit monster only when really close to it.
And of course allowed range should be calculated dynamically with many other conditions :)

PS
not a C# coder.

exectails:
@S-anasol You misunderstand, the problem is not checking if the hit should hit, it's preventing the client from *trying* to hit in the first place, as it will display the hit, even if the server decides not to damage the monster.

S-anasol:
oh i see)
only solution is wating for official client update?

Or official server does not allow to the client to do so?

exectails:
As I've described above, the server sends *something* to the client that tells it the range it can hit monsters in. We have to figure out what that something is. It's a problem with Melia, not the client.

gastonb:
It might be related to weapon currently equipped?

exectails:
@gastonb It's quite possible, though there must be a default value, as the issue is the same if you don't have a weapon equipped.

exectails:
Hum... when you mentioned equipping I realized the information could come from other packets as well, like ZC_NORMAL. I then looked at a equipping log, where I switched weapons, and found a pretty long ZC_NORMAL with *properties*.

[...]

PR #34, opened 2016-04-09
Fixing #4 attack range.

Tachiorz:


exectails:
Looks like this doesn't fix splash yet, but finally we have a range =)

PR #38, opened 2016-04-09
Changes Command Operator from '/' to '>'

vegax87:


Cosrnos:
This might be a small tweak code wise but it's a pretty big tweak for those used to using commands ^^

exectails:
This was intentional, see https://github.com/aura-project/melia/blob/master/system/conf/commands.conf#L7. Your change would prevent the command manager from catching official commands.

PR #42, opened 2016-04-11
Sector System

gastonb:
Added sector system. 
- it handle actor's positions
- it handle actor's interactions.


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -908,14 +908,35 @@ public void CZ_SKILL_GROUND(ChannelConnection conn, Packet packet)
...
+
+			Log.Debug("SKILL pos1 {0},{1},{2}", x1,y1,z1);
+			Log.Debug("SKILL pos2 {0},{1},{2}", x2, y2,z2);
+
+
+			// Add skill to world
+			// Get map
+			var map = ChannelServer.Instance.World.GetMap(character.MapId);
+
+			SkillActor PESkill = new SkillActor();
```
exectails: Should be camelCase.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -908,14 +908,35 @@ public void CZ_SKILL_GROUND(ChannelConnection conn, Packet packet)
...
+			PESkill.ownerHandle = character.Handle;
+			PESkill.SkillId = skillId;
+			PESkill.Position = skillPosition;
+			PESkill.Init();
+
+			map.AddSkill(PESkill);
```
exectails: We've talked a little about his yesterday on the chat, in my opinion calling a skill an actor is wrong. A skill can be many things and isn't necessarily an entity on a map. What is an entity would be the skill *effect*.

src/ChannelServer/World/Character.cs
```diff
@@ -165,11 +164,118 @@ public int GetJumpType()
...
+		public void Process()
+		{
+			if (this.IsMoving)
+				ProcessMove();
+		}
+
+		/// <summary>
+		/// Process Character's movement
+		/// </summary>
+		public void ProcessMove()
```
exectails: Does this need to be public? If not, the IsMoving check is redundant.

src/ChannelServer/World/Character.cs
```diff
@@ -165,11 +164,118 @@ public int GetJumpType()
...
+			}
+				
+		}
+
+		/// <summary>
+		/// Set character's destination position
+		/// </summary>
+		public void SetDestination(Position pos)
+		{
+			this._destination = pos;
```
exectails: No `this.`

src/ChannelServer/World/Character.cs
```diff
@@ -182,14 +288,17 @@ public void Move(float x, float y, float z, float dx, float dy, float unkFloat)
...
 			this.SetDirection(dx, dy);
 			this.IsMoving = false;
+			*/
+			this.SetDestination(new Position(x, y, z));
 
 			// Sending ZC_MOVE_STOP works as well, but it doesn't have
 			// a direction, so the character stops and looks north
 			// on other's screens.
-			Send.ZC_PC_MOVE_STOP(this, this.Position, this.Direction);
+			//Send.ZC_PC_MOVE_STOP(this, this.Position, this.Direction);
```
exectails: If you don't send ZC_PC_MOVE_STOP here, but from the tick, the character won't stop on other clients until the tick. If the tick is several hundreds of milliseconds long, + the time it takes for the packet to arrive, you might get a big difference in what the clients display.

src/ChannelServer/World/Map.cs
```diff
@@ -97,6 +134,11 @@ public void AddCharacter(Character character)
 
 			lock (_characters)
 				_characters[character.Handle] = character;
+
+			if (!SectorManager.Add(character, character.Position))
```
exectails: Shouldn't it be enough to pass the entity? It has the Position property after all.

src/ChannelServer/World/Map.cs
```diff
@@ -109,6 +151,8 @@ public void RemoveCharacter(Character character)
 				_characters.Remove(character.Handle);
 
 			character.Map = null;
+
+			SectorManager.Remove(character);
```
exectails: While it's unlikely to happen, *if* a tick would occur between setting Map to null and the entity actually being removed from the sector manager, a visit event could make use of Map, causing a null exception.

src/ChannelServer/World/Map.cs
```diff
@@ -270,6 +314,27 @@ public virtual void Broadcast(Packet packet, IEntity source, bool includeSource
...
+				Log.Debug("Error adding skill placeable entity into SectorManager at: {0} {1} {2}", skill.Position.X, skill.Position.Y, skill.Position.Z);
+			}
+		}
+
+		public void RemoveSkill(SkillActor skill)
+		{
+			lock (_skills)
+				_skills.Remove(skill.Handle);
+
+			SectorManager.Remove(skill);
```
exectails: `this.`

src/ChannelServer/World/Sector.cs
```diff
@@ -0,0 +1,125 @@
...
+			if (sector == null)
+				return;
+
+			this.neighbors[(int)direction] = sector;
+		}
+
+		/// <summary>
+		/// Performs a visit for a given visitor and range, to all reachable entities in this sector.
+		/// </summary>
+		public void VisitInternal(Position pos, IVisitor visitor, float range)
```
exectails: Should a "internal" method be public?

src/ChannelServer/World/Sector.cs
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+	}
+
+	/// <summary>
+	/// Sector directions
+	/// </summary>
+	public enum SectorDirection : int
+	{
+		SECTOR_N = 0,
```
exectails: Should be PascalCase.

src/ChannelServer/World/SectorManager.cs
```diff
@@ -0,0 +1,258 @@
...
+		/// <summary>
+		/// Sectors for this sector manager
+		/// </summary>
+		private Sector[] sectors;
+
+		/// <summary>
+		/// Amount of "units" in width.
+		/// IMPORTANT: This is not the max Position in X, but the max amount of units. The length of the map in X direction.
+		/// </summary>
+		private int width;
```
exectails: Underscore.

src/ChannelServer/World/SkillActor.cs
```diff
@@ -0,0 +1,60 @@
...
+using Melia.Shared.World;
+
+namespace Melia.Channel.World
+{
+	public class SkillActor : Actor, IVisitor
+	{
+		public float range;
+		public int SkillId;
+		public bool fired;
+		public int delayRemove = 200;
```
exectails: Looks like this should be a const.

src/Shared/World/Actor.cs
```diff
@@ -0,0 +1,73 @@
...
+	{
+		/// <summary>
+		/// Actor's Handle
+		/// </summary>
+		public int Handle {	get; set; }
+
+		/// <summary>
+		/// Flag indicating if this actor must be destroyed
+		/// </summary>
+		public bool ToDestroy;
```
exectails: Always prefer properties over fields.

src/Shared/World/Actor.cs
```diff
@@ -0,0 +1,73 @@
...
+		/// Check if an actor is close enough (to a given range) for another actor. 
+		/// This function doesn't take in count Radius. 
+		/// </summary>
+		public bool IsMyArea(Position pos, float range)
+		{
+			// THIS RANGE CHECK must be FAST. 
+			// It is used to concider if an actor is close enough to other, to make further checks for collision (isNearObject)
+
+			// Check if object is within range
+			return this.Position.InRange2D(pos, (int)range); // Change this function for a function not using Math.Sqrt()
```
exectails: InRange2D doesn't use Sqrt.

exectails:
Naming and convention issues aside the system looks pretty good. I think we should rethink the class hierarchy of entities, actors, characters, etc. in general, but so far so good.

src/ChannelServer/World/SkillActor.cs
```diff
@@ -0,0 +1,60 @@
...
+		public void Init()
+		{
+			this.range = 0.0f;
+			this.Handle = ChannelServer.Instance.World.CreateHandle();
+		}
+
+		public bool OnVisit(Actor entity)
+		{
+			// check if attackable
+			if (entity.GetType() != typeof(Character))
```
exectails: `if(entity is Character)`

exectails:
I'll leave this open while @gastonb is on vacation.

exectails:
Closing for now, since it's not mergeable, and the issues were apparently never fixed.

Issue #41, opened 2016-04-11
Scripting naming conventions


exectails:
The naming of variables and functions in our scripts is currently not very clearly defined. While the built-in functions, like `resetstats` use all lower case without spaces, some functions inside the scripts use snake_case, and some use PascalCase. This is because of different influences, but we should agree on one and stick to it.

The reason I used nospace in the built-in functions was that I wanted to mimic eAthena a little, to the point where several functions might have the same name, making scripting for former Athena users easier. However, I didn't feel comfortable using that convention for longer function and variable names, so I switched to underscores in the custom NPCs. For variable names I then used camelCase, because that's what I'm used to from most C-like languages.

I don't think we should apply our C# conventions to Lua, as they're very different languages with different idioms. All Lua functions and variable names in official examples seem to be nospace, so that seems like the most logical choice to me. We could then use underscores for categorization. For example: `npc_statresetter`.

vegax87:
Another question related to scripting: how do we handle multiple scripts of the same NPC ID? I mean, every Bulletin Board in Tree of Savior has a different message, do we need to create something like: Board_1.lua , Board_2.lua or handle these variables elsewhere?

exectails:
@vegax87 In the future you should make a new issue for technical inquires, make a thread on the forum, or ask on the chat, for unrelated questions.

You definitely don't need multiple files, though right now we would need multiple functions. But when someone on the chat made a suggestion in this regard, I had an idea.

We could use the last parameter of addnpc, the function name, as either a function name or a localization key, similar to the names. That way you don't have to create a dedicated function for every simple NPC, that only has one line of dialog anyway. For example:

```lua
addnpc(20111, "ETC_20150317_009196", "c_Klaipe", 394, -1, -475, 45, "ETC_3938548505")
```
instead of
```lua
addnpc(20111, "ETC_20150317_009196", "c_Klaipe", 394, -1, -475, 45, "npc_board1")
function npc_board1()
    msg("ETC_3938548505")
end
```

This reminds me that we should add comments or English names to the addnpc calls, it's impossible to tell what's what in there =<

PR #48, opened 2016-04-16
Simplified client-side dialog support

exectails:
Allows for localization keys and dialog names to be passed to addnpc in place of a Lua function name, so no dedicated function has to be written for the NPC. As described on #41.

Example:
```lua
addnpc(20139, "QUEST_LV_0100_20150317_001418", "c_Klaipe", -60, 79, -446, -33.69, "ETC_20150317_000015")
addnpc(20137, "ETC_20151224_019889", "c_Klaipe", -205, 79, -515, 51.17, "MASTER_ICEMAGE_basic1")
```


innovagen:
How can i know that?

NPCid xxxx use (ETC_xxxx_xxxx) ?

Example:
NPCid 20139 use (ETC_20150317_000015)

exectails:
You can either log the dialog packets with PaleTree, or you can search the NPC and its dialog in the client. Key files: ies_mongen.ipf/* and ies.ipf/npcselectdialog.ies,

PR #49, opened 2016-04-16
Script functions getitem and delitem

exectails:


Cosrnos:
getitem and delitem are kind of weird names like can't it be additem and removeitem? I feel those are far less ambiguous?

exectails:
@Cosrnos That's what they were called in eAthena, I thought copying some names would be a nice touch for people coming from there.

Cosrnos:
personally I think giveItem or takeItem would work then, you wouldn't *give* the world an item, you'd add one to it

exectails:
giveitem it is I guess. Using the same name across our server emulators is not a bad thing either.

exectails:
>takeItem

I actually didn't even think of the NPC's POV... what's the most intuitive one? Player, NPC, or server? o~o

exectails:
@Cosrnos Thinking about it, the player POV does make sense. There's no object for the function, nothing passed to it, but it implicitly happens on the character. That's the case for all functions, unless otherwise specified. If an innocent player read such a script, I would imagine them to see `getitem` and think to themselves "right, I *get* the item".

Why is naming stuff always such a drag @_@

exectails:
`learnskill`, `dropitem`, `(un)equip`, `warp`, `resetstats`, I'm starting to think the player POV makes a lot of sense.

PR #66, opened 2016-05-02
Multi Level up and Job attribute on lua's getpc()

huenato:
>> ScriptManager.cs
- Added Level Up function
- Added getpc attribute on account data: job
>> Character.cs
- Added Level Up function that supports multiple levels

### DIFF
--- a/src/ChannelServer/Scripting/ScriptManager.cs
+++ b/src/ChannelServer/Scripting/ScriptManager.cs
@@ -122,7 +122,7 @@ namespace Melia.Channel.Scripting
 			Register(resetstats);
 			Register(changehair);
 			Register(spawn);
-            Register(levelup);
+			Register(levelup);
 		}
 
 		/// <summary>
@@ -934,7 +934,7 @@ namespace Melia.Channel.Scripting
 		///     integer maxSp,    -- Character's max SP
 		///     integer stamina,  -- Character's stamina
 		///     integer hair,     -- Character's hair
-        ///     integer job,      -- Character's job
+		///     integer job,      -- Character's job
 		///     table account
 		///     {
 		///         string  name, -- Account's name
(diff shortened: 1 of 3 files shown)


src/ChannelServer/World/Character.cs
```diff
@@ -321,11 +321,27 @@ public void LevelUp()
...
-		/// <summary>
-		/// Grants exp to character and handles level ups.
-		/// </summary>
-		/// <param name="exp"></param>
-		public void GiveExp(int exp, int jobExp, Monster monster)
+        /// <summary>
+        /// Increases character's level by the given amount of levels.
+        /// </summary>
+        /// <param name="amount"></param>
+        public void LevelUp(int amount)
```
exectails: As this method basically does exactly the same as the other one, you should combine them, to reduce redundancy. For example, the one without parameter could call this one with amount = 1, or you could remove the other one and give amount a default value of 1.

src/ChannelServer/Scripting/ScriptManager.cs
```diff
@@ -122,6 +122,7 @@ private void InitializeLua()
 			Register(resetstats);
 			Register(changehair);
 			Register(spawn);
+            Register(levelup);
```
exectails: Please set your IDE to use tabs for indention, as outlined in our coding conventions, which can be found in the [readme](https://github.com/aura-project/melia#coding-conventions).

exectails:
Once you fix the mentioned issues I'll merge this. If you currently don't have the time, I'll close this PR for now and you can create a new one later on.

huenato:
I'll finish this now. I used all my free time yesterday playing on Silute.

huenato:
Is there something that I can use to define the max level on the server?

src/ChannelServer/World/Character.cs
```diff
@@ -322,6 +322,15 @@ public void LevelUp()
 		}
 
 		/// <summary>
+		/// Increases character's level by one.
+		/// </summary>
+		/// <param name="amount"></param>
+		public void LevelUp()
+		{
+			LevelUp(1);
```
exectails: This should be prefixed with `this.`, as per our coding conventions, but I'll fix that.

exectails:
@huenato There's no max level setting yet, feel free to add it. I think it should go into a world conf file. You could use the exp db to figure out the max level for which exp values are available though.

PR #71, opened 2016-05-08
Added lempransa pond (orsha) as starting map and added team name to query on character creation

huenato:


exectails:
Looks like you forgot to push the addition of the new db class to the project, resulting in a compilation error.

huenato:
Oops, there you go

src/LoginServer/LoginServer.cs
```diff
@@ -54,7 +54,11 @@ public override void Run()
 			this.CheckDatabaseUpdates();
 
 			// Data
-			this.LoadData(DataToLoad.All, true);
+			this.LoadData(DataToLoad.Jobs, true);
```
exectails: You could also do this: `this.LoadData(DataToLoad.Jobs | DataToLoad.Maps | DataToLoad.Barracks | [...], true);`

system/db/startingcities.txt
```diff
@@ -0,0 +1,13 @@
+// Melia
+// Database file
+//---------------------------------------------------------------------------
+
+[
+// Klaipeda - West Siauliai Woods
+//---------------------------------------------------------------------------
+{ startingCityId: 1, Map: "f_siauliai_west", X: -628, Y: 260, Z: -1025 },
```
exectails: Our naming convention for db field names is camelCase, as in the other dbs.

src/Shared/Data/Database/StartingCities.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using Newtonsoft.Json.Linq;
+using Melia.Shared.Const;
+
+namespace Melia.Shared.Data.Database
+{
+	[Serializable]
+	public class StartingCityData
+	{
+		public int Id { get; set; }
```
exectails: Why not use the StartingCity enum for the id and the index?

huenato:
I'll update the code in the next 2 days. I'll be a bit busy for a while.

PR #75, opened 2016-08-11
NPC Pack 2

Hirasu:
Almost all npcs with dialogs and trigger, also some extra informations about each npc.


exectails:
Did you run around and test these dialogues? Seeing how your generated code looks much different from the example code I wrote, based on logs from iTOS, I'm a little worried about the results.

For example, the blacksmith in Orsha:
```lua
msg("TUTO_REPAIR_NPC")
msg("ORSHA_BLACKSMITH_basic01")
msg("ORSHA_BLACKSMITH_basic02")
```

Now I don't know what he's supposed to say, but there's probably much more to it, and a selection is probably missing as well. Should we dummy fill them like this?

Btw, it would be cool if you could add indention to the generator.

Hirasu:
@exectails 
It don't generate selections. "ORSHA_BLACKSMITH_basic01/ORSHA_BLACKSMITH_basic02"
I thought that should be a selection.

ies
It check the "NormalBeforeDialog1 - 4", "SD_QuestDialog1 - 4", "ID_QuestDialog1 - 4"  columns too.
But it also should check the dialogtext.ies...

> addnpc(40070, "QUEST_20151224_003161", "c_orsha", -1077.49, 500.61, 332.74, -17,"QUEST_20151224_003161_Call")
 function QUEST_20151224_003161_Call()
 end

board
Seems something is broken.

Hirasu:
I found a another way to get all dialogs and selections, I'll do a new pull if i'm done.

PR #83, opened 2017-06-06
Updated opcodes for i159264

Ci3l0phi:
Please review ignoring white space changes using `?w=1`.
All opcodes should be ordered by hexadecimal (same way as before).

Since there are so many changes since the last update, almost everything will need to be updated.

### DIFF
--- a/src/Shared/Network/Op.cs
+++ b/src/Shared/Network/Op.cs
@@ -6,1593 +6,1647 @@ using System.Reflection;
 
 namespace Melia.Shared.Network
 {
-	/// <summary>
-	/// Packet op code enum
-	/// </summary>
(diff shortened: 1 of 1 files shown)


exectails:
>Please review ignoring white space changes

I can review it like that all I want, but I'm not going to merge it like that^^ Please use tabs, as per the coding conventions.

exectails:
Thank you for the PR, and sorry I didn't review it in time. The ops are surely severely outdated already again, so I'm gonna close this for now, but if you make another PR I'll do better.

celophi:
Thank you for taking a look at this exec. I can indeed make another PR which is updated with the correct opcodes.

PR #84, opened 2017-08-01
Npc ingame name tag

Hirasu:
Fu*k, this pull request was an mistake... please close it.


celophi:
@Hirasu It doesn't have to be :)

PR #87, opened 2017-09-01
Updated barrack packet structures for i170175 compatability.

celophi:
These changes fix the ability to login to the barrack and create a new character.


src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -339,6 +335,10 @@ public void CB_COMMANDER_MOVE(LoginConnection conn, Packet packet)
 			var d1 = packet.GetFloat();	// ?
 			var d2 = packet.GetFloat();	// ?
 
+			// On a new character creation, this packet is sent with the index as this byte.
+			if (index == 0xFF)
+				return;
```
celophi: I'm not sure what the purpose of this `0xFF` byte is. It is hardcoded like this in the client.

src/LoginServer/Network/LoginPacketSender.cs
```diff
@@ -67,6 +68,13 @@ public static void BC_COMMANDER_LIST(LoginConnection conn)
 				for (int i = 0; i < Items.EquipSlotCount; ++i)
 					packet.PutShort(0);
 
+				// Unknown
+				packet.PutByte(1);
+				packet.PutByte(1);
+				packet.PutByte(1);
+				packet.PutByte(1);
+				packet.PutByte(0);
```
celophi: These bytes do not appear to be associated with the equipment properties structure.

src/Shared/Const/Items.cs
```diff
@@ -113,13 +114,13 @@ public static class Items
...
 		/// </remarks>
-		public const int EquipSlotCount = 21;
+		public const int EquipSlotCount = 22;
 
 		/// <summary>
 		/// Ids of the items equipped by default.
 		/// (Literally empty items, NoHat, NoWeapon, etc.)
 		/// </summary>
-		public static readonly int[] DefaultItems = new int[EquipSlotCount] { 2, 2, 12101, 8, 6, 7, 10000, 11000, 9999996, 9999996, 4, 9, 9, 4, 9, 9, 9, 9, 9, 10, 2 };
+		public static readonly int[] DefaultItems = new int[EquipSlotCount] { 2, 2, 12101, 8, 6, 7, 10000, 11000, 9999996, 9999996, 4, 9, 9, 4, 9, 9, 9, 9, 9, 10, 2, 4 };
```
celophi: This implementation is not going to be flexible moving forward. The IES files actually have columns that denote the default starting equipment for each class, and having that information extracted seems more useful to me.

src/Shared/Network/Helpers/CommanderHelper.cs
```diff
@@ -42,6 +42,11 @@ public static void AddCommander(this Packet packet, ICommander commander)
 			{
 				packet.PutInt(0);
 			}
+
+			// [i170175 (2017-09-01)] ?
+			{
+				packet.PutInt(0);
```
celophi: This appears to be some buffer on the `APPEARANCE_PC` object, but I haven't seen it used by the client yet. It may just be a buffer.

system/db/startingcities.txt
```diff
@@ -5,9 +5,5 @@
 [
 // Klaipeda - West Siauliai Woods
 //---------------------------------------------------------------------------
-{ startingCityId: 1, map: "f_siauliai_west", x: -628, y: 260, z: -1025 },
-
-// Orsha - Lempransa Pond
-//---------------------------------------------------------------------------
-{ startingCityId: 2, map: "f_siauliai_16", x: -2694, y: 116, z: -1663 },
+{ startingCityId: 1, map: "f_siauliai_west", x: -628, y: 260, z: -1025 }
```
celophi: Do we need to keep this file?

src/LoginServer/Network/LoginPacketHandler.cs
exectails: Hm, maybe a reset of some sort.

system/db/startingcities.txt
exectails: Is it not possible to start in different cities anymore?

src/Shared/Network/Helpers/CommanderHelper.cs
exectails: This assumption of what it does should probably go into the code =)

exectails:
Thanks for the PR, looks good. Just one thing we probably need to think/talk about is whether we want to keep the packet history in the source code.

I started putting those `// [i1203983948]` comments everywhere on the off-chance that someone might want to go back to a previous version in the future, or provide multi-version support, but here you have not documented some of the changes, and I don't know if that's even necessary anymore.

Do we want to keep this information, just in case? Then we should add it. If not, we should probably remove most of it, to prevent confusion, with someone thinking something has always been there, when it really hasn't.

celophi:
> Do we want to keep this information, just in case?

Okay. I vote for removing it. I added it just because they were already there.

src/Shared/Network/Helpers/CommanderHelper.cs
celophi: You mean a comment? Yes. It should.

system/db/startingcities.txt
celophi: Yeah it's not sadly :( (Why IMC??). Also it occurred to me that the ID for that city is actually `256` and Orsha is `512`. The only way I know is because they changed this before they removed the city option altogether.

exectails:
Yea, I don't know how useful it would really be for going back anyway. There are so many packets that we haven't mapped out yet, where we don't even have any information about what came when, even ignoring the >1 year we didn't really keep track of the changes.

Alright, I'll merge this then, unless you want to make any changes to it?

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -266,7 +258,7 @@ public void CB_COMMANDER_CREATE(LoginConnection conn, Packet packet)
 			character.Hair = hair;
 
 			character.MapId = mapData.Id;
-			character.Position = new Position(startingCityData.X, startingCityData.Y, startingCityData.Z);
+			character.Position = new Position(-628, 260, -1025);     // These are the starting coordinates for 'f_siauliai_west'.
```
exectails: We should definitely make this configurable at some point though, like in Athena^^

PR #88, opened 2017-09-01
Additional barrack packets

celophi:
Summary:
- Fixes the barrack layer selection being stuck after selecting '2'.
- Updates the client UI when a team name change is purchased (changes not persisted in the database yet.)
- Updates the character's position.
- Fixes deletion of characters.

### DIFF
--- a/doc/bt/BC_COMMANDER_DESTROY.bt
+++ b/doc/bt/BC_COMMANDER_DESTROY.bt
@@ -2,12 +2,12 @@
 //--- 010 Editor v6.0.2 Binary Template
 //
 // File: BC_COMMANDER_DESTROY.bt
-// Author: exec
-// Revision:
+// Author: exec, celophi
+// Revision: i170175
 // Purpose:
 //--------------------------------------
 
 short op;
 int sequence;
 
-byte index;
+__int64 characterId;
--- a/doc/bt/BC_NORMAL_SetPosition.bt
+++ b/doc/bt/BC_NORMAL_SetPosition.bt
@@ -1,0 +1,23 @@
+//------------------------------------------------
+//--- 010 Editor v8.0 Binary Template
+//
+//      File: BC_NORMAL_SetPosition.bt
+//   Authors: celophi
+//   Version: i170175
+//   Purpose: Notifies the client that the server successfully set the barrack position for an actor.
+//  Category: 
+// File Mask: 
+//  ID Bytes: 
+//   History: 
+//------------------------------------------------
+
+#include "common.bt"
+
+short op;
+int reserved;
+short packetLength;
+int subOp;
+
+__int64 accountId;
+byte barrackIndex;
+position pos;
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -132,6 +132,7 @@ namespace Melia.Login.Network
 			Send.BC_SERVER_ENTRY(conn, "127.0.0.1", 9001, "127.0.0.1", 9002);
 			Send.BC_COMMANDER_LIST(conn);
 			Send.BC_NORMAL_ZoneTraffic(conn);
+			Send.BC_NORMAL_TeamUI(conn);
 		}
 
 		/// <summary>
@@ -297,6 +298,7 @@ namespace Melia.Login.Network
 			}
 
 			Send.BC_COMMANDER_DESTROY(conn, index);
+			Send.BC_NORMAL_TeamUI(conn);
 		}
 
 		/// <summary>
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -186,6 +186,8 @@ namespace Melia.Login.Network
 			LoginServer.Instance.Database.UpdateTeamName(conn.Account.Id, name);
 
 			Send.BC_BARRACKNAME_CHANGE(conn, TeamNameChangeResult.Okay);
+			Send.BC_ACCOUNT_PROP(conn, conn.Account);
+			Send.BC_NORMAL_Run(conn, "THEMA_BUY_SUCCESS");
 		}
 
 		/// <summary>
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -330,7 +330,10 @@ namespace Melia.Login.Network
 			}
 
 			// Move
-			character.BarrackPosition = new Position(x, y, z);
+			// This needs to be validated such that a commander may not move to an invalid coordinate.
+			var pos = new Position(x, y, z);
+			character.BarrackPosition = pos;
+			Send.BC_NORMAL_SetPosition(conn, index, pos);
 		}
 
 		/// <summary>
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -278,13 +278,13 @@ namespace Melia.Login.Network
 		[PacketHandler(Op.CB_COMMANDER_DESTROY)]
 		public void CB_COMMANDER_DESTROY(LoginConnection conn, Packet packet)
 		{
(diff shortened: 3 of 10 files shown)


src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -132,6 +132,7 @@ public void CB_START_BARRACK(LoginConnection conn, Packet packet)
 			Send.BC_SERVER_ENTRY(conn, "127.0.0.1", 9001, "127.0.0.1", 9002);
```
celophi: Unrelated to this PR, these should be migrated to settings. Additionally, it is completely fine to send just one set of IP address and port.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -185,6 +186,8 @@ public void CB_BARRACKNAME_CHANGE(LoginConnection conn, Packet packet)
 			LoginServer.Instance.Database.UpdateTeamName(conn.Account.Id, name);
 
 			Send.BC_BARRACKNAME_CHANGE(conn, TeamNameChangeResult.Okay);
+			Send.BC_ACCOUNT_PROP(conn, conn.Account);
+			Send.BC_NORMAL_Run(conn, "THEMA_BUY_SUCCESS");
```
celophi: I considered moving strings like "THEMA_BUY_SUCCESS" into a separate location. I think it would be better to statically define them somewhere so that they can be updated only once instead of multiple places.

doc/bt/BC_COMMANDER_DESTROY.bt
```diff
@@ -2,12 +2,12 @@
...
+// Author: exec, celophi
+// Revision: i170175
 // Purpose:
 //--------------------------------------
 
 short op;
 int sequence;
 
-byte index;
+__int64 characterId;
```
exectails: Why do you prefix the int64s with underscores? Not that that doesn't work, but it's not necessary.

doc/bt/BC_NORMAL_SetPosition.bt
```diff
@@ -0,0 +1,23 @@
+//------------------------------------------------
```
exectails: Shouldn't this be part of BC_NORMAL? I assume you wanted to split it off for cleanliness, but it will only be useful as a template if you can apply it to any BC_NORMAL and get the correct result. With it split off you would have to know what the specific BC_NORMAL does to look up the template.

If we want it cleaner (I don't know how big BC_NORMAL is), maybe we could split off only the subop specific variables, and `#include` the files in the parent template.

src/LoginServer/Network/LoginPacketHandler.cs
exectails: Yes, definitely. Though the main advantage I see is auto-completion and protection from typos.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -277,13 +280,13 @@ public void CB_COMMANDER_CREATE(LoginConnection conn, Packet packet)
 		[PacketHandler(Op.CB_COMMANDER_DESTROY)]
 		public void CB_COMMANDER_DESTROY(LoginConnection conn, Packet packet)
 		{
-			var index = packet.GetByte();
+			var id = packet.GetLong();
```
exectails: They finally decided to use the ids instead of the indexes? Hallelujah.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -429,5 +436,23 @@ public void CB_REQ_CHANNEL_TRAFFIC(LoginConnection conn, Packet packet)
...
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CB_SELECT_BARRACK_LAYER)]
+		public void CB_SELECT_BARRACK_LAYER(LoginConnection conn, Packet packet)
+		{
+			var layer = packet.GetInt();
+
+			// Only these two layers are valid currently.
+			if (layer < 1 || layer > 2)
+				return;
```
exectails: I love how you can simply return in ToS. In Mabi almost every packet has a response packet, and if you don't send it, the client locks up until it receives it... Though there should be a log message, so that if something about this ever changes, or there is some reason this might have another value somehow, there will be a warning in the console and the log files, which we can look up to figure out why a user is reporting that barrack layer selection isn't working^^ Log all the things, better safe than sorry.

src/LoginServer/Network/LoginPacketSender.cs
```diff
@@ -158,6 +165,50 @@ public static void BC_COMMANDER_DESTROY(LoginConnection conn, byte index)
...
+		/// <summary>
+		/// Moves a character in the barrack.
+		/// </summary>
+		/// <param name="conn"></param>
+		/// <param name="index"></param>
+		/// <param name="position"></param>
+		public static void BC_NORMAL_SetPosition(LoginConnection conn, byte index, Position position)
+		{
+			var packet = new Packet(Op.BC_NORMAL);
+			packet.PutInt(0x02); // subop
```
exectails: We should probably make an enum for those some time.

src/LoginServer/Network/LoginPacketHandler.cs
exectails: That was... the social server? Mhm, you're right, I guess we just hardcoded it because the social server doesn't exist in Melia yet.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -330,7 +334,10 @@ public void CB_COMMANDER_MOVE(LoginConnection conn, Packet packet)
 			}
 
 			// Move
-			character.BarrackPosition = new Position(x, y, z);
+			// This needs to be validated such that a commander may not move to an invalid coordinate.
```
exectails: For TODO type comments it's useful to add a `TODO:` prefix, so IDEs can pick them up and make them searchable.

doc/bt/BC_COMMANDER_DESTROY.bt
celophi: Oh. I was under the impression that it was. I can fix those.

exectails: You can if you want, otherwise I will just run a replace all over the files some time.

exectails:
We both made changes to the BC_COMMANDER_DESTROY template, so you'll have to fix that before I can merge it.

doc/bt/BC_NORMAL_SetPosition.bt
celophi: In the case of `BC_NORMAL` it is around 20. For the `ZC_NORMAL` packets, there might be a few thousand.

I think splitting off the parent template and children would work fine.

exectails: Yea, though on the other hand, you're saying thousands, do we want to litter the folder with thousands of sub-template files?^^"

src/LoginServer/Network/LoginPacketHandler.cs
celophi: I hate those indexes. I guess if they seriously needed a smaller packet size they can do it, but really?

src/LoginServer/Network/LoginPacketHandler.cs
celophi: Yes, there should be a log message.

src/LoginServer/Network/LoginPacketHandler.cs
exectails: Yea, no idea what the idea behind those indexes was.

doc/bt/BC_NORMAL_SetPosition.bt
celophi: Not really, I was assuming a two child folders - one for `BC_NORMAL` and the other for `ZC_NORMAL`.

exectails: Yea, still a *lot* of files to track and clone. But oh well. I think we should see how it goes. One giant ZC_NORMAL template might not be that terrible either, that way you can see everything at once.

PR #89, opened 2017-09-02
Updates the structures for commanders and appearances.

celophi:
Summary:
- Additional structure created for appearance.
- Fixed `ZC_CONNECT_OK` structure.

You can test these structure definitions by running the developer console or looking at the lua scripts from the IPFs. They use the same names. I'm not needlessly attempting to change names; we can name them whatever we want to, but I would prefer to use the same or similar names as found in the client to avoid confusion.

### DIFF
--- a/doc/bt/ZC_CONNECT_OK.bt
+++ b/doc/bt/ZC_CONNECT_OK.bt
@@ -11,9 +11,8 @@
 //   History: 
 //   - i109XX:  s1 was added
 //   - i11025:  s1 was removed
-//   - i11257:  b1 was added
-//   - i170175: b2 was added
-//   - i170175: Properties were combined into the COMMANDER structure.
+//   - i11257:  isIntegrateServer was added
+//   - i170175: isIntegrateIndunServer was added
 //------------------------------------------------
 
 #include "common.bt"
--- a/doc/bt/ZC_CONNECT_OK.bt
+++ b/doc/bt/ZC_CONNECT_OK.bt
@@ -2,7 +2,7 @@
 //--- 010 Editor v8.0 Binary Template
 //
 //      File: ZC_CONNECT_OK.bt
-//   Authors: exec
+//   Authors: exec, celophi
 //   Version: i170175
 //   Purpose: 
 //  Category: 
@@ -13,6 +13,7 @@
 //   - i11025:  s1 was removed
 //   - i11257:  b1 was added
 //   - i170175: b2 was added
+//   - i170175: Properties were combined into the COMMANDER structure.
 //------------------------------------------------
 
 #include "common.bt"
@@ -25,33 +26,8 @@ byte gameMode;
 int i1;
 byte privileges;
 byte bin1[10];
-byte b1;
-byte b2;
+byte isIntegrateServer;
+byte isIntegrateIndunServer;
 
 lpString sessionKey;
-
-//short s1;
-
-int handle1;
-int i2;
-
 COMMANDER commander;
-
-float x;
-float y;
-float z;
-int exp;
-int maxExp;
-int i3;
-
-int64 characterId;
-int64 socialInfoId;
-
-int hp;
-int maxHp;
-short sp;
-short maxSp;
-int stamina;
-int maxStamina;
-short shield;
-short maxShield;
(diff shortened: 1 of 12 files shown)


src/ChannelServer/Database/ChannelDb.cs
```diff
@@ -110,8 +110,8 @@ public Character GetCharacter(long accountId, long characterId)
 					character.MaxExp = reader.GetInt32("maxExp");
 					character.Hp = reader.GetInt32("hp");
 					character.MaxHp = reader.GetInt32("maxHp");
-					character.Sp = reader.GetInt32("sp");
-					character.MaxSp = reader.GetInt32("maxSp");
+					character.Sp = reader.GetInt16("sp");
+					character.MaxSp = reader.GetInt16("maxSp");
```
celophi: Should we change these columns to be `smallint`?

doc/bt/ZC_UPDATED_PCAPPEARANCE.bt
```diff
@@ -17,4 +18,4 @@ short op;
 int sequence;
 
 int handle;
-COMMANDER commander;
+APPEARANCE_PC appearancePc;
```
celophi: I think this makes a lot of sense now.

doc/bt/common.bt
```diff
@@ -112,11 +113,30 @@ typedef struct
     int i7;
     short hair;
     short pose;
-
-    int i8;
+    int teamId;
     int i9;
 }
-COMMANDER;
+APPEARANCE_PC;
```
celophi: The `APPEARANCE_PC` structure is also a component of the `APPEARANCE_BARRACK_PC` and `BARRACK_PC` structures.

doc/bt/ZC_CONNECT_OK.bt
```diff
@@ -25,33 +26,8 @@ byte gameMode;
 int i1;
 byte privileges;
 byte bin1[10];
-byte b1;
-byte b2;
+byte isIntegrateServer;
```
exectails: The names should be changed in the history as well, and I guess we should talk about what we want the history to represent. So far I've only used it to document the changes to the actual packet structure, not changes in how we organize the files. The history of how the packet itself changed is important when you have older logs, for which you need to modify the template to make it compatible, but off the bat I can't think of a reason to document file changes.

doc/bt/ZC_UPDATED_PCAPPEARANCE.bt
exectails: Yea, it does. I'm all for this change if that's also how the client handles it, even if restructuring like this is naturally sub-optimal, since the same name suddenly means something different, and a whole lot of code changes without any *actual* change, which can easily lead to mistakes and bugs.

src/ChannelServer/Database/ChannelDb.cs
exectails: Actually, in C# it's customary to use int for pretty much anything, unless you need specific types, like for inter-operation. One reason for this is that .NET is optimized for working with ints. In this case an int also ensures that we won't have to change anything but the put call, should the type ever change. That's why they are all int, even if the packet uses a short.

Initially I did the same thing in Aura, using the exact types for everything, thinking that it would better explain what values a property can take, but for me personally it turned out that I never cared one bit about what the types actually are, because the properties need to check and clamp the incoming values anyway. Enforcing the actual types made our lives harder though, because you constantly have to figure out the type, cast, check values, etc.

Though I just noticed that our PutShort has an int overload... I don't remember why I did that, that seems unsafe and should probably be changed =P"

exectails:
>I'm not needlessly attempting to change names; we can name them whatever we want to, but I would prefer to use the same or similar names as found in the client to avoid confusion.

I think so too, that's why we already have MONSTER and COMMANDER after all. I was thinking about renaming them to PascalCase, but upper-case makes it clear that they're client structs.

doc/bt/ZC_CONNECT_OK.bt
celophi: > document the changes to the actual packet structure.
This makes the most sense to me.

exectails:
Thanks, good job.

src/ChannelServer/Database/ChannelDb.cs
celophi: Yes, that was a good explanation, and if .NET is optimized for dealing with `int` then we should take advantage of that. 

Now thinking about this, it occurred to me that if for some reason a value larger than 32767 existed in the database, placing that value as short would mean rollover. It may not be an issue if only the application stores the value, but if someone were to change values on their own in the database, they would need to be aware of the size limitations enforced by the application.

PR #91, opened 2017-09-02
Added structures for APPEARANCE_BARRACK_PC and BARRACK_PC

celophi:
These changes correct assumptions of structures that were previously `Character`.

### DIFF
--- a/src/LoginServer/Network/Helpers/AppearanceBarrackPcHelper.cs
+++ b/src/LoginServer/Network/Helpers/AppearanceBarrackPcHelper.cs
@@ -6,6 +6,7 @@ using Melia.Login.World;
 using Melia.Shared.Const;
 using Melia.Shared.Network;
 using Melia.Shared.Network.Helpers;
+using Melia.Shared.World;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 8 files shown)


src/LoginServer/Network/Helpers/AppearanceBarrackPcHelper.cs
```diff
@@ -14,43 +15,52 @@
 
 namespace Melia.Login.Network.Helpers
 {
-	public static class CharacterHelper
+	public static class AppearanceBarrackPcHelper
 	{
-		public static void AddCharacter(this Packet packet, Character character)
+		public static void AddAppearanceBarrackPc(this Packet packet, IAppearanceBarrackPc pc)
```
exectails: One has to say though, "AppearanceBarrackPc" is a mouthful =P"

src/LoginServer/World/Character.cs
```diff
@@ -51,5 +58,15 @@ public int[] GetEquipIds()
...
 			return this.Equipment;
 		}
+
+		/// <summary>
+		/// Returns the equipment properties as an array.
+		/// </summary>
+		/// <returns></returns>
+		public int[] GetEquipmentProperties()
+		{
+			// TODO: This needs to return the actual properties of equipment which have variable lengths.
```
exectails: Hm, this is probably fine for now, because the result isn't actually used, but if someone was to try to use it, it might have unexpected consequences. Maybe an empty array would be better?

src/LoginServer/Network/Helpers/AppearanceBarrackPcHelper.cs
celophi: Yes. I don't like that at all. Do you want to use an abbreviation or different name?

src/LoginServer/World/Character.cs
celophi: Yes an empty array would be fine. When I was fixing this for the current version i170175, I found that if these properties aren't exactly perfect, (values, not sizes), then it breaks the commander list or crashes the client.

PR #92, opened 2017-09-03
This change adds chat macros.

celophi:
The official server implementation makes these account specific; however, we can change that at any time or make it a setting in the future.

### DIFF
--- a/src/ChannelServer/Database/Account.cs
+++ b/src/ChannelServer/Database/Account.cs
@@ -2,6 +2,7 @@
 // For more information, see license file in the main folder
 
 using Melia.Channel.Scripting;
+using Melia.Channel.World;
 using Melia.Shared.Const;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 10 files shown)


src/ChannelServer/Database/Account.cs
```diff
@@ -54,6 +60,33 @@ public Account()
 		{
 			this.Settings = new AccountSettings();
 			this.Variables = new Variables();
+			this._chatMacros = new List<ChatMacro>();
```
exectails: `this` should only be used for publics, not privates that are already prefixed with an underscore. Though I recently thought that it might make sense to actually use `this` for them as well, *instead* of prefixing them, for consistency, and because I see this mistake made all the time with my chosen code conventions^^

celophi: I'm fine with using it only with `publics`; it is not a problem :)

PR #94, opened 2017-09-03
Added map fog revealing and exploration packets and implementation.

celophi:


src/ChannelServer/Database/Account.cs
```diff
@@ -90,6 +96,32 @@ public ChatMacro[] GetChatMacros()
...
 
 		/// <summary>
+		/// Adds a revealed map to the account.
+		/// </summary>
+		/// <param name="map"></param>
+		public void AddRevealedMap(RevealedMap revealedMap)
+		{
+			lock (_revealedMaps)
+			{
+				var oldMap = _revealedMaps.FirstOrDefault(x => x.MapId == revealedMap.MapId);
```
exectails: I don't know how often this will be called, but a Dictionary and TryGetValue might be more efficient than repeatedly calling FirstOrDefault on a potentially large list.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -1150,6 +1151,60 @@ public void CZ_ITEM_SELL(ChannelConnection conn, Packet packet)
...
+		/// </summary>
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CZ_MAP_REVEAL_INFO)]
+		public void CZ_MAP_REVEAL_INFO(ChannelConnection conn, Packet packet)
+		{
+			var mapId = packet.GetInt();
+			var visible = packet.GetBin(128);
+
+			var mapData = ChannelServer.Instance.Data.MapDb
```
exectails: All indexed databases, like MapDb, have a Find method by default that searches for the given index in the most efficient way. Same as above, the TryGetValue the db uses is more efficient than a FirstOrDefault.

src/ChannelServer/Database/Account.cs
celophi: It's called once per login to the zone server. However, it's still a good idea to use an efficient implementation if one exists.

exectails: Isn't it also called to update the data though while you run around? In that case it might be called much more often.

celophi: Oh yes you are correct. I was looking at the `Get` method. The sending of maps is done only once.

PR #95, opened 2017-09-04
Packet definitions

celophi:
This change adds various definitions from my fork. Most of them are trivial.


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -1150,6 +1150,32 @@ public void CZ_ITEM_SELL(ChannelConnection conn, Packet packet)
...
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CZ_EXCHANGE_REQUEST)]
+		public void CZ_EXCHANGE_REQUEST(ChannelConnection conn, Packet packet)
+		{
+			var targetHandle = packet.GetInt();
+		}
+
+		/// <summary>
+		/// Specifies various commands found in 'customcommand.ies'.
```
celophi: I'd like it if we can add this server data file in the near future. It is necessary for me to continue the work on quests.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -879,7 +879,7 @@ public static void ZC_OBJECT_PROPERTY(ChannelConnection conn, Character characte
 			packet.PutLong(character.Id);
 			foreach (var property in properties)
 			{
-				packet.PutShort(property);
+				packet.PutInt(property);
 				switch (property)
 				{
 					case ObjectProperty.PC.HP: packet.PutFloat(character.Hp); break;
```
celophi: As an aside, is there a reason you did not choose to make `ObjectProperty` a dictionary? I found instances where I needed to iterate through a list of these and was forced to write a switch statement like the one above. When we move toward sending a list of all stats for a character (I think 50-60 stats), then these statements will get lengthy.

That's just an observation I wanted to make. If you have a good reason for them being constant variables, then just let me know.

exectails: You mean like `Dictionary<ObjectProperty, float>`? The problem with that is that not all properties are floats as far as I know, there are at least also ints and strings. As a result you would need do make it `<ObjectProperty, object>`, and that would result in a lot of boxing, unboxing, and type checking.

Tachiorz was working on an alternative way of handling properties back when we started, though that solution has problems in addition to the boxing. https://github.com/aura-project/melia/issues/53

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1676,6 +1730,30 @@ public static void ZC_UPDATE_SP(Character character, short spValue)
...
 
 		/// <summary>
+		/// Updates a characters HP for damage and healing.
+		/// </summary>
+		/// <param name="character"></param>
+		/// <param name="amount"></param>
+		/// <param name="isDamage"></param>
+		public static void ZC_ADD_HP(Character character, int amount, bool isDamage)
+		{
+			character.HPChangeCounter += 1;
```
exectails: Is it a good idea to do logic inside a packet sender? It seems like this might have unforeseen consequences if someone doesn't expect a sender to actually *apply* the damage, instead of just updating the client. I believe this would be better placed in a method on Character, which then calls this function.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1782,4 +1873,11 @@ public MsgParameter(string key, string value)
 			this.Value = value;
 		}
 	}
+
+	public enum FactionType : int
+	{
+		PLAYER = 1,
```
exectails: Code convention: PascalCase

src/ChannelServer/Network/ChannelPacketSender.cs
celophi: I should be more clear. The class named `ObjectProperty.cs` looks likes this.
```
public static class ObjectProperty
	{
		public static class Account
		{
			public const short MaxPVPGrade = 6299;
			public const short Medal = 6300;
                        ...
                }
        }
```

I'm suggesting something such as this.
```
public static class ObjectProperty 
{
    public static readonly Dictionary<string, int> Account = new Dictionary<string, int>
    {
        { "MaxPVPGrade", 6299 },
        ...
    }
}
````

Comment (exectails) [reply src/ChannelServer/Network/ChannelPacketSender.cs]: Oh, well, that would be highly inefficient. Every access to one property id would entail a look up in a dictionary.  You would also lose intellisense and value safety for the names, unless you add another class just for the names.

In case you also want to know why it's consts and not an enum, that's because we don't actually need enumeration for them, and this way we don't have to cast all the time.

To still iterate over such a class you can actually use reflection though.
```csharp
foreach (var constant in typeof(ObjectProperty.Account).GetFields())
{
	Log.Debug(constant.Name + ": " + constant.GetValue(null));
}
```

src/ChannelServer/Network/ChannelPacketHandler.cs
exectails: Done.

doc/bt/CZ_EXCHANGE_REQUEST.bt
```diff
@@ -0,0 +1,17 @@
+//------------------------------------------------
+//--- 010 Editor v8.0 Binary Template
+//
+//      File: CZ_EXCHANGE_REQUEST.bt
+//   Authors: celophi
+//   Version: i170175
+//   Purpose: Indicates to the client a request to trade with another character.
```
exectails: *To* the client? CZ packets come *from* the client, don't they?

PR #96, opened 2017-09-04
Added help topics and server data.

celophi:


src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -1757,6 +1757,32 @@ public static void ZC_SKILL_MELEE_GROUND(Character character, int id, Position p
...
+					.FirstOrDefault(x => x.ClassName == topic);
+
+				if (data == null)
+				{
+					Log.Error("ZC_HELP_LIST: Help data '{0}' not found.", topic);
+					return;
+				}
+
+				packet.PutInt(data.HelpId);
+				packet.PutByte(0); // Unknown.
```
exectails: My guess (see template) was that this is a boolean, for whether the topic was shown to the player already.

src/Shared/Server.cs
```diff
@@ -223,6 +228,7 @@ public enum DataToLoad
 		Exp = 0x80,
 		Dialogues = 0x100,
 		Shops = 0x200,
+		Help = 0x400,
```
exectails: Another plus for changing how we load stuff, no conflicts for two people adding new dbs at the same time anymore^^

system/db/help.txt
```diff
@@ -0,0 +1,90 @@
+// Melia
+// Database file
+//---------------------------------------------------------------------------
+
+[
+{ HelpId: 1, ClassName: "TUTO_MOVE_KB", DbSave: "YES", BasicHelp: "YES",  },
```
exectails: Boolean values should probably use actual booleans, instead of YES and NO.

PR #98, opened 2017-09-04
Fixes the "Quit", "Logout", and "Return to barrack", buttons.

celophi:


doc/bt/CZ_RUN_GAMEEXIT_TIMER.bt
```diff
@@ -0,0 +1,18 @@
...
+// File Mask: 
+//  ID Bytes: 
+//   History: 
+//------------------------------------------------
+
+short op;
+int sequence;
+int checksum;
+
+string destination; // Maximum length is 32 characters.
```
exectails: If this is a fixed size string, shouldn't it be char[32]?

celophi: It's not fixed. If you write it as `char[32]`, the array will contain junk bytes. (The value of the column displayed will be correct though).

exectails: It's not? It's always 32 byte, isn't that the definition of fixed?^^ You're right that the array will contain junk the client didn't clear of course, but what will you do if, for example, they decide to add an int after destination now?

celophi: Oh I see. Perhaps I didn't realize what "fixed" means.

exectails: At least that's what I meant^^

PR #99, opened 2017-09-04
Adds TP properties and fixes barrack map purchasing.

celophi:
Note that purchased barrack maps are not persisted. This just fixes the charging.

### DIFF
--- a/src/LoginServer/World/Character.cs
+++ b/src/LoginServer/World/Character.cs
@@ -47,6 +47,7 @@ namespace Melia.Login.World
 		{
 			this.Level = 1;
 			this.Channel = 1;
+			this.BarrackLayer = 1;
 			this.Equipment = Items.DefaultItems.ToArray();
 		}
 
--- a/src/Shared/Const/Message.cs
+++ b/src/Shared/Const/Message.cs
@@ -71,4 +71,12 @@ namespace Melia.Shared.Const
 		/// </summary>
 		public const string KEYBOARD_TUTORIAL = "KEYBOARD_TUTORIAL";
 	}
+
+	/// <summary>
+	/// Constant strings sent with `BC_NORMAL_Run`.
+	/// </summary>
+	public static class BarrackMessage
+	{
+		public const string THEMA_BUY_SUCCESS = "THEMA_BUY_SUCCESS";
+	}
 }
--- a/system/db/barracks.txt
+++ b/system/db/barracks.txt
@@ -3,7 +3,7 @@
 //---------------------------------------------------------------------------
 
 [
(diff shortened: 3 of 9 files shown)


system/db/barracks.txt
```diff
@@ -3,7 +3,7 @@
 //---------------------------------------------------------------------------
 
 [
-{ mapId: 11, name: "Cozy Lodge", price: 0, characters: 5 },
-{ mapId: 12, name: "Pyromancer Room", price: 5, characters: 7 },
-{ mapId: 13, name: "Banquet Hall", price: 5, characters: 10 },
+{ mapId: 11, name: "Cozy Lodge", price: 0, characters: 4 },
+{ mapId: 12, name: "Pyromancer Room", price: 69, characters: 4 },
+{ mapId: 13, name: "Banquet Hall", price: 69, characters: 4 },
```
celophi: I seem to recall that each barrack had a larger available number of character slots, but the data file is not showing that.

src/Shared/Const/Message.cs
```diff
@@ -71,4 +71,12 @@ public static class AddonMessage
 		/// </summary>
 		public const string KEYBOARD_TUTORIAL = "KEYBOARD_TUTORIAL";
 	}
+
+	/// <summary>
+	/// Constant strings sent with `BC_NORMAL_Run`.
+	/// </summary>
+	public static class BarrackMessage
+	{
+		public const string THEMA_BUY_SUCCESS = "THEMA_BUY_SUCCESS";
```
celophi: @exectails: Once you have decided whether we should change the styling of these messages, we can change all of them at once.

src/LoginServer/World/Character.cs
```diff
@@ -47,6 +47,7 @@ public Character()
 		{
 			this.Level = 1;
 			this.Channel = 1;
+			this.BarrackLayer = 1;
```
celophi: This is not directly related to the PR, but should be fixed anyway. It prevents a layer of '0' being saved to the database, and consequently, the character will not show up when `BC_COMMANDER_LIST` is sent.

exectails: If that's the case BarrackLayer should probably not even accept a 0 if that's not valid. The property could be clamped to 1~max, like I did for HP and co. in my last commit.

src/Shared/Const/Message.cs
exectails: Yea, I'm not sure. What do you think? Seeing upper case underscore naming makes me cringe, but it makes sense from the perspective of someone who digs around the client^^

celophi: I think it's fine to rename them. Not everything needs to be the same as the client.

PR #102, opened 2017-09-05
Added stance change packet.

celophi:


doc/bt/ZC_STANCE_CHANGE.bt
```diff
@@ -13,6 +13,6 @@
 
 short op;
 int sequence;
-int checksum;
```
exectails: Whoops, thanks for the fix^^

PR #106, opened 2017-09-06
Default equipment for character creation.

celophi:
This change adds the default equipment necessary when we create characters.

Note:
- There is one inventory item which seems to be incorrect in our default items array. We'll need to change this to the correct id.

### DIFF
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -9,6 +9,7 @@ using Melia.Shared.Network;
 using Melia.Shared.Util;
 using Melia.Shared.Util.Security;
 using Melia.Shared.World;
+using System.Collections.Generic;
 
 namespace Melia.Login.Network
 {
@@ -258,11 +259,22 @@ namespace Melia.Login.Network
 			character.Spr = jobData.Spr;
 			character.Dex = jobData.Dex;
 
-			if (!character.SetDefaultEquipment())
+			// Initialize with the default equipment set.
+			var equipment = new Dictionary<EquipSlot, int>();
+			foreach (var equip in jobData.DefaultEquip)
 			{
-				Log.Error("CB_COMMANDER_CREATE: An error occurred while setting the default equipment for a new character.");
-				return;
+				var itemData = LoginServer.Instance.Data.ItemDb.FindByClass(equip.Value);
+				if (itemData == null)
+				{
+					Log.Error("CB_COMMANDER_CREATE : Unable to find item data with class name '{0}'.", equip.Value);
+					Send.BC_MESSAGE(conn, MsgType.CannotCreateCharacter);
+					return;
+				}
+
+				equipment[equip.Key] = itemData.Id;
 			}
+
+			character.SetEquipment(equipment);
 
 			conn.Account.CreateCharacter(character);
 
--- a/src/LoginServer/Network/LoginPacketHandler.cs
+++ b/src/LoginServer/Network/LoginPacketHandler.cs
@@ -258,6 +258,12 @@ namespace Melia.Login.Network
 			character.Spr = jobData.Spr;
 			character.Dex = jobData.Dex;
 
+			if (!character.SetDefaultEquipment())
+			{
+				Log.Error("CB_COMMANDER_CREATE: An error occurred while setting the default equipment for a new character.");
+				return;
+			}
+
 			conn.Account.CreateCharacter(character);
 
 			Send.BC_COMMANDER_CREATE_SLOTID(conn, character);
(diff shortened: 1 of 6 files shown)


src/Shared/Const/Items.cs
```diff
@@ -52,28 +52,28 @@ public static int GetIndex(this InventoryCategory cat, int index)
...
+		Ring2, // OUTERADD2
+		Outer2, // BODY
+		Pants, // PANTS
+		Ring3, // PANTSADD1
+		Ring4, // PANTSADD2
+		Bracelet1, // RING1
+		Bracelet2, // RING2
+		Necklace, // NECK
+		Hat, // HAT_T
+		Lens, // LENS
```
celophi: Hopefully the case on these comments don't bother you @exectails . I do want to keep them around because they change often enough, and having the correct names even as a reference will make it easy.

src/Shared/Data/Database/Items.cs
```diff
@@ -33,6 +33,12 @@ public ItemData Find(string name)
 			return this.Entries.FirstOrDefault(a => a.Value.Name.ToLower() == name).Value;
 		}
 
+		public ItemData FindByClass(string name)
+		{
+			name = name.ToLower();
+			return this.Entries.FirstOrDefault(a => a.Value.ClassName.ToLower() == name).Value;
+		}
```
celophi: I added this just for utility. We probably will have other instances where we need to search by class name.

src/LoginServer/World/Character.cs
```diff
@@ -64,5 +65,33 @@ public int[] GetEquipmentProperties()
...
+		/// <summary>
+		/// Sets the default equipment for a character.
+		/// </summary>
+		/// <returns></returns>
+		public bool SetDefaultEquipment()
+		{
+			var jobData = LoginServer.Instance.Data.JobDb.Find(this.Job);
+			if (jobData == null)
+			{
+				Log.Error("SetDefaultEquipment : Unable to set default equipment for job '{0}' because it doesn't exist.", this.Job);
```
celophi: I don't really like logging in the domain like this. Perhaps there is a nicer way.

src/LoginServer/World/Character.cs
```diff
@@ -64,5 +65,33 @@ public int[] GetEquipmentProperties()
 			// TODO: This needs to return the actual properties of equipment which have variable lengths.
 			return this.GetEquipIds();
 		}
+
+		/// <summary>
+		/// Sets the default equipment for a character.
+		/// </summary>
+		/// <returns></returns>
+		public bool SetDefaultEquipment()
```
celophi: The logic here should really only be invoked during character creation, and I thought I might place it in the constructor. However, we also use the constructor when restoring from the data persistence layer (I think ORMs and data mappers use reflection to avoid this problem). Let me know your thoughts.

src/Shared/Const/Items.cs
```diff
@@ -52,28 +52,29 @@ public static int GetIndex(this InventoryCategory cat, int index)
...
+		Wing, // WING
```
celophi: Hopefully the case on these comments don't bother you @exectails . I do want to keep them around because they change often enough, and having the correct names even as a reference will make it easy.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -258,6 +258,12 @@ public void CB_COMMANDER_CREATE(LoginConnection conn, Packet packet)
 			character.Spr = jobData.Spr;
 			character.Dex = jobData.Dex;
 
+			if (!character.SetDefaultEquipment())
```
exectails: What happens to the client in this case? Is it okay with not receiving a response? And if not, is there a way to recover from the mistake of the server, or should the connection maybe be killed to get them out of there?

src/LoginServer/World/Character.cs
exectails: I think there's little need to worry, because where is this gonna be called anyway, except for on creation? If someone went in there and just called it somewhere else for some reason, they might even have a reason for it.

If you want to be 100% sure I would either create a bool that is only not set for new characters, and set at the end of this and for existing characters, to check if it was called already, or check if Equip is empty.

src/LoginServer/World/Character.cs
exectails: You could always make the method take a list of items that you prepare in the packet handler.

src/Shared/Data/Database/Items.cs
exectails: Absolutely, I wanted to add the same for looking up the default equip.

src/LoginServer/Network/LoginPacketHandler.cs
celophi: The result is that the client would not progress away from the character creation screen. I see that this should at the very least reply with `BC_MESSAGE` to inform the user about why the action did not work.

PR #107, opened 2017-09-06
Added packet for setting the hat visibility state.

celophi:


exectails:
Unbelievable. You know, I suspected something like this, but then I thought "... no, that can't be it. Who in their right mind would set the "show/hide" button to "show" on default, when it's really hidden by default?

Good job^^

Issue #103, opened 2017-09-05
NPC dialog source


exectails:
Our NPC's dialog code currently looks something like this:
```lua
if select("TUTO_GIRL_basic_02", "!@#$TUTO_GIRL_select01#@!", "!@#$Auto_JongLyo#@!") ~= 1 then
	return
end
if select("TUTO_GIRL_basic_03", "!@#$TUTO_GIRL_select02#@!", "!@#$Auto_JongLyo#@!") ~= 1 then
	return
end
if select("TUTO_GIRL_basic_04", "!@#$TUTO_GIRL_select03#@!", "!@#$Auto_JongLyo#@!") ~= 1 then
	return
end

msg("TUTO_GIRL_basic_05")
```
We take advantage of the fact that the client has the dialogues for all NPCs, and just send localization keys. This has the advantage that it's pretty easy to create NPCs from logs, and we don't have to worry about localization, as long as there is an official translation for your language.

However, we also put ourselves at the mercy of the client. Any string can change at any time, just like their keys, and the script we wrote might not make any sense anymore after a few updates. That wouldn't be a problem if we sent the actual text, without relying on the client. It would also allow users to actually figure out what's going on in a script, instead of just guessing, based on structure and keys.

```lua
if select("It would not be strange if the world ended tomorrow.{nl}However, I just want to know. Why the goddesses disappeared..", "Ask about the disappearance of the goddesses", "Cancel") ~= 1 then
	return
end
if select("I'm not sure I understand it completely, but what I'm sure of is that the goddesses didn't disappear all at once.", "Ask what happened", "Cancel") ~= 1 then
	return
end
if select("Goddess Vakarine left us shortly before Medzio Diena.{nl}Both Zemyna and Ausrine disappeared soon after.{nl}It's already been four years since.", "I think you forgot to mention Goddess Laima", "Cancel") ~= 1 then
	return
end

msg("She has never appeared before.{nl}Although I've read somewhere that she does appear every now and then...")
```
Typing all this out is certainly more work though, even if a Pale plugin could help with that, and I don't know if anyone will actually care about the dialog. It should be said though that writing the script would also be easier with actual text.

I thought I'd create a discussion for it, because I'm not sure which is better, and we should definitely stick to one.

celophi:
I can see why this feature would be desired; however, I'm not knowledgeable enough about how localization would work without the keys here. @exectails Would you mind explaining this in depth?

[...]

PR #109, opened 2017-09-06
Adds the default set of chat macros.

celophi:


system/db/chatmacros.txt
```diff
@@ -0,0 +1,16 @@
+﻿// Melia
+// Database file
+//---------------------------------------------------------------------------
+
+[
+{ "id": "1", "name": "Set1", "pose": "8", "text": "@dicID_^*$ETC_20151001_014770$*^" },
```
celophi: This suffers from the same issue as #103.

src/ChannelServer/Database/Account.cs
```diff
@@ -87,6 +88,26 @@ public void AddChatMacro(ChatMacro macro)
 		/// <returns></returns>
 		public ChatMacro[] GetChatMacros()
 		{
+			// Initialize with the default set of macros if none exist.
+			if (_chatMacros.Count() == 0)
```
celophi: Worth it to reduce nesting?

system/db/chatmacros.txt
exectails: Indeed =/ Having values like these seems so pointless, because you have *no* idea what the actual text is until you look it up.

src/ChannelServer/Database/Account.cs
exectails: Would this make more sense in the ctor, so it doesn't has to be checked on every call?

src/ChannelServer/Database/Account.cs
```diff
@@ -87,6 +88,26 @@ public void AddChatMacro(ChatMacro macro)
...
+			{
+				var macroData = ChannelServer.Instance.Data.ChatMacroDb
+					.Where(x => x.Id <= 10) // maximum number of chat macros allowed.
+					.OrderBy(x => x.Id);
+
+				if (macroData == null)
```
exectails: This will never be null, because you're guaranteed to get an enumerator from the above call.

src/ChannelServer/Database/Account.cs
```diff
@@ -87,6 +88,26 @@ public void AddChatMacro(ChatMacro macro)
 		/// <returns></returns>
 		public ChatMacro[] GetChatMacros()
 		{
+			// Initialize with the default set of macros if none exist.
+			if (_chatMacros.Count() == 0)
+			{
+				var macroData = ChannelServer.Instance.Data.ChatMacroDb
+					.Where(x => x.Id <= 10) // maximum number of chat macros allowed.
```
exectails: Should all 10 be added/sent? I get only the 7 defaults that actually have a pose id sent to me on iToS.

system/db/chatmacros.txt
celophi: It's true. The alternative in this case would be something like `{img emoticon_0001 30 30}`. Do you think this is better in this scenario?

celophi: Update:
I realized that this is not data that we need to keep in Melia (at least for now).

exectails: Well, if it were the official value (`{img emoticon_0022 30 30} I'm sorry`), at the very least a user could see that, and change it if they wanted to do. The way it is now it's just some cryptic dict id, where a user might not even think they *can* change it, and if so, to what.

celophi: All good points. The client will send the `dicID` if I recall. What do you think about converting this value to regular text before it is stored in the database?

exectails: The client sends the dicID? Isn't the text customizable? Or do you mean it sets up and sends the default?

src/ChannelServer/Database/Account.cs
```diff
@@ -86,6 +87,17 @@ public Account()
 			this.Variables = new Variables();
 			_chatMacros = new List<ChatMacro>();
 			_revealedMaps = new Dictionary<int, RevealedMap>();
+
+			// Initialize with the default set of macros.
+			var macroData = ChannelServer.Instance.Data.ChatMacroDb
+				.Where(x => x.Id <= 10) // maximum number of chat macros allowed.
```
exectails: While this will probably never fail, I have a bad feeling when I look at it, because *theoretically* the database could suddenly include a 0 id, or even negative ones, be it through a bug, or a user error, or whatever. It would be safer to get all, order them, and add them until you've reached a max of 10.

And what about the 0 poses, should they actually be added? Like I said, I only get 7 macros when I log into iToS.

PR #110, opened 2017-09-06
Added normal packet for account updates.

celophi:
Adds the packet necessary for providing account related information to the zone server.
We should put additional thought into the structure of this considering the information is shared.

### DIFF
--- a/doc/bt/common.bt
+++ b/doc/bt/common.bt
@@ -277,3 +277,13 @@ byte checkCompression()
 {
     return (ReadUShort() == 0xFA8D);
 }
+
+typedef struct
+{
+    short size;
+    short s1;
+
+    if(size > 0)
+        byte properties[size-2]; // todo, short:val list
+}
+AccountProperties;
--- a/src/Shared/Network/Helpers/AccountHelper.cs
+++ b/src/Shared/Network/Helpers/AccountHelper.cs
@@ -1,11 +1,14 @@
 ﻿// Copyright (c) Aura development team - Licensed under GNU GPL
 // For more information, see license file in the main folder
 
-using Melia.Login.Database;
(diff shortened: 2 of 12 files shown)


doc/bt/common.bt
```diff
@@ -277,3 +277,13 @@ byte checkCompression()
...
     return (ReadUShort() == 0xFA8D);
 }
+
+typedef struct
+{
+    short size;
+    short s1;
+
+    if(size > 0)
+        byte properties[size-2]; // todo, short:val list
```
celophi: This structure can use something similar to the `ZC_OBJECT_PROPERTY.bt` template.

src/Shared/Network/Helpers/AccountHelper.cs
```diff
@@ -32,4 +30,12 @@ public static void AddAccountProperties(this Packet packet, Account account)
...
 		}
 	}
+
+	public interface IAccount
+	{
+		int Medals { get; }
+		int GiftMedals { get; }
+		int PremiumMedals { get; }
+		int SelectedBarrack { get; }
+	}
```
celophi: Perhaps this should really be a base class such as `BaseCharacter.cs`.

src/Shared/Network/Helpers/AccountHelper.cs
```diff
@@ -32,4 +30,12 @@ public static void AddAccountProperties(this Packet packet, Account account)
 			packet.PutFloat(account.SelectedBarrack);
 		}
```
celophi: Additionally, we could opt for using the `mapvisibility` data for populating all of the `HadVisited` object properties. This is why I decided on storing the map ID.

PR #115, opened 2017-09-07
Updated the CZ_ITEM_USE packet.

celophi:


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -650,6 +650,35 @@ public void CZ_HEAD_ROTATE(ChannelConnection conn, Packet packet)
...
+			}
+
+			// Do not allow use of locked items.
+			if (item.IsLocked)
+			{
+				Log.Warning("CZ_ITEM_USE: User '{0}' tried to use a locked item.", conn.Account.Name);
+				return;
+			}
+
+			// TODO: Implement use of item.
```
celophi: I left out the rest of the implementation, but included a stub because in order to record client-side logic that we should also verify server side too.

exectails: A server message would be good, to let the user know that the feature isn't actually working yet, and they don't wonder why *nothing* happened, not even an error or an unimplemented message in the channel.

PR #116, opened 2017-09-07
Added specification for ZC_EQUIP_ITEM_REMOVE.

celophi:


exectails:
Huh, when is this packet used?

PR #117, opened 2017-09-08
Added packet to change equipment durability.

celophi:


src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -490,6 +490,21 @@ public static void ZC_EQUIP_ITEM_REMOVE(Character character, Item item, int mess
 		}
 
 		/// <summary>
+		/// Updates the durability of an item in an equipment slot.
+		/// </summary>
+		/// <param name="character"></param>
+		/// <param name="slot"></param>
+		/// <param name="durability">Value in thousandths that the item has remaining.</param>
```
celophi: At first I had kept this value as a `float`, and done math to convert it; however, I think you mentioned before to leave logic out of the send methods if possible.

exectails: Minor math to convert something we use into what the client expects would be totally fine. However, there's a reason they used an int instead of a float: float imprecision. Since we will do math on the durability, adding and subtracting from it, we can't be sure that the value will match up or transfer 100% perfectly. With ints you have that guarantee.

PR #126, opened 2017-09-11
various BC_NORMAL definitions

celophi:


doc/bt/BC_NORMAL.bt
```diff
@@ -81,14 +81,33 @@ switch(type)
     case 0x0C:
         byte bin1[size-12]; // zipped zone information
         break;
+
+    // Enable packets from client to barrack.
+    // (CB packets are disabled normally after connecting to the zone server.)
```
celophi: This might be necessary when visiting barracks of other players, but I'm not certain.

doc/bt/BC_NORMAL.bt
```diff
@@ -81,14 +81,33 @@ switch(type)
...
+    // The client will respond with CB_CLIENT_INTEGRITY_FAIL
+    case 0x18:
+        break;
+
+    // Update the number of purchased character slots in the barrack.
+    case 0x19:
+        int count;
+        break;
+
+    // Nexon Game Security Callback.
```
celophi: I don't know if we will ever use this, but I want to define everything possible.

doc/bt/ZC_NORMAL.bt
```diff
@@ -112,6 +117,10 @@ switch (type)
 	case 0x14E:
 		lpString sessionKey;
 		break;
+
+    // Nexon Game Security Callback.
+    case 0x16D;
+        break;
```
celophi: Our formatting is off D:

src/Shared/Network/SubOp.cs
```diff
@@ -25,12 +30,14 @@ public static class Zone
 			public const int LevelUp = 0x11;
 			public const int ClassLevelUp = 0x14;
 			public const int Unkown_1c = 0x1C;
+			public const int BarrackSlotCount = 0x3C;
```
celophi: Apparently you can do this from the zone server.

doc/bt/BC_NORMAL.bt
exectails: That comment should've made it into the template^^

PR #128, opened 2017-09-12
Additional definitions

celophi:


doc/bt/ZC_SET_LAYER.bt
```diff
@@ -0,0 +1,17 @@
+//------------------------------------------------
+//--- 010 Editor v8.0 Binary Template
+//
+//      File: ZC_SET_LAYER.bt
+//   Authors: celophi
+//   Version: i171140
+//   Purpose: Sets a UI overlay. This is used for cutscenes and possibly other cases.
```
exectails: This sounds interesting. What exactly does it mean? Is this used for the phasing? Or is it just something related to the UI?

doc/bt/ZC_UPDATE_SHIELD.bt
```diff
@@ -0,0 +1,18 @@
+//------------------------------------------------
+//--- 010 Editor v8.0 Binary Template
+//
+//      File: ZC_UPDATE_SHIELD.bt
+//   Authors: celophi
+//   Version: i171140
+//   Purpose: Updates an actor's shield value.
```
exectails: I should play some ToS some time, I still don't know what exactly this "shield" property is that I keep seeing in our code =P

doc/bt/ZC_SET_LAYER.bt
celophi: I don't know what "phasing" means, but when you enter a cutscene, all of the UI elements disappear. The character cannot interact except with dialog controls. This is known as "director mode", I think, and this packet enables that.

doc/bt/ZC_UPDATE_SHIELD.bt
celophi: Me either! But, you should be able to find all of these in the lua files. That might give us a hint at what it does.

doc/bt/ZC_SET_LAYER.bt
exectails: Phasing is dynamic instancing, you run into a part of the map and suddenly you see nothing but yourself and your party. In ToS that's used for "story" based cutscenes and battles. I was thinking the layer parameter might be like an instance/phase id. Though I guess it could be a padded byte toggle as well.

exectails: Hm, doesn't look like a toggle
```text
12 0D FF FF FF FF FA 02 00 00
```

PR #129, opened 2017-09-12
Adds additional client definitions.

celophi:


src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -1438,6 +1438,24 @@ public void CZ_CLEAR_INDUN_REG(ChannelConnection conn, Packet packet)
...
+		public void CZ_SYNC_POS(ChannelConnection conn, Packet packet)
+		{
+			var handle = packet.GetInt();
+			var x = packet.GetFloat();
+			var y = packet.GetFloat();
+			var z = packet.GetFloat();
+
+			// TODO: Sanity checks. 
+			// Make certain that the displacement from the current position is realistic.
+			conn.SelectedCharacter.SetPosition(x, y, z);
```
exectails: I'm not sure we should just add teleportation packets that we might easily forget about^^ It also has a handle parameter, so it's probably not only for the selected character.

PR #132, opened 2017-09-14
Added base stat structure.

celophi:


src/Shared/Network/Helpers/CommanderHelper.cs
```diff
@@ -32,14 +32,23 @@ public static void AddCommander(this Packet packet, ICommander commander)
...
 			packet.PutLong(commander.Id + 1);   // SocialInfoId
 
-			packet.PutInt(commander.Hp);
-			packet.PutInt(commander.MaxHp);
-			packet.PutShort(commander.Sp);
-			packet.PutShort(commander.MaxSp);
-			packet.PutInt(commander.Stamina);
-			packet.PutInt(commander.MaxStamina);
-			packet.PutShort(0); // Shield
-			packet.PutShort(0); // MaxShield
```
celophi: I'm not sure where "MaxShield" comes from, but no such thing exists on this structure.

src/ChannelServer/Database/ChannelDb.cs
```diff
@@ -112,12 +112,12 @@ public Character GetCharacter(long accountId, long characterId)
 					character.MapId = reader.GetInt32("zone");
 					character.Exp = reader.GetInt32("exp");
 					character.MaxExp = reader.GetInt32("maxExp");
-					character.MaxHp = reader.GetInt32("maxHp");
-					character.Hp = reader.GetInt32("hp");
-					character.MaxSp = reader.GetInt32("maxSp");
-					character.Sp = reader.GetInt32("sp");
-					character.MaxStamina = reader.GetInt32("maxStamina");
-					character.Stamina = reader.GetInt32("stamina");
+					character.BaseStat.MaxHp = reader.GetInt32("maxHp");
```
exectails: We should talk about this change. Grouping related properties like this has some advantages, like slightly cleaner parent classes, but there's also problems, like how it makes it harder to get into the code.

If I'm a new contributor, wanting to modify a character's stamina for example, and I can't find a stamina property on it, I'd open the Character class, looking for it, confused. Is it called something weird? Did Intellisense fail me? What exactly are "base stats"? And this process will repeat for every property I'm looking for, from "base stats" over the actual stats to god knows what, and not only on Character, but potentially on Item, Monster, etc. too.

Since you made this a struct in the templates, I'd assume it's a struct officially and you wanted to mimic that, maybe saying that we don't have to do that for everything, but imo that would make it even more confusing, if one property is "hidden" somewhere, and another isn't.

Creating helpers for all these group classes also complicates the packet writing and maintenance, because it's not a simple matter of reading the code anymore, you have to hop back and forth between methods, trying following what's going on. That's less of a problem if you have ironclad structures based on the client's data, but I don't know how ironclad they are, and anybody who doesn't get its information that way will have a harder time.

And all this for seemingly little reason, aside from a more concise parent/wrapper class. I mainly see disadvantages in this change.

src/Shared/Network/Helpers/CommanderHelper.cs
exectails: We probably assumed it was another "current/max" pair, based on the others.

src/ChannelServer/Database/ChannelDb.cs
celophi: I think you make good points. My main thought process behind doing this was, as you mentioned, to mimic the models that I see in the client. Furthermore, I thought that it is very likely that these stats are properties that exist on all classes inheriting from the base actor; whereas, the others are specific only to player characters.

That being said, it probably isn't necessary to have a separate class to represent this. What does make sense to me is to have some representation where it is clear that they are properties of the entity/actor. Whether that is an interface/comment/structure is probably not as significant, and we should choose to not detract away from the ease of working and recognizing the character class and what belongs to it.

exectails: An interface is probably a good idea, just what I was thinking as I read your comment. It's too bad C# doesn't support mixins.

exectails:
Any news on this pull request?

exectails:
Closing for now, as there are conflicts and because the pull request hasn't been updated in 3 weeks.

PR #133, opened 2017-09-14
Clarified some experience packets.

celophi:


doc/bt/ZC_EXP_UP.bt
```diff
@@ -15,4 +15,4 @@ short op;
 int sequence;
 
 int exp;
-int jobExp;
+int totalExp;
```
exectails: Oh~ yea, that makes a lot more sense xD

Issue #112, opened 2017-09-06
Feature database


exectails:
This is something that doesn't really affect us just yet, but that I think we should talk about before it does affect us.

**Problem**

It's pretty typical for MMOs to change with updates, and often times these updates include changes that players don't like. In turn these changes are often times disabled on pservers, but I feel like most server emulators do a pretty lackluster job of providing ways to quickly and comfortably change these settings.

Let's take Ragnarok as an example, and more specifically, the destruction of Morroc. When this update hit, eAthena changed all warps and NPCs to match the new official ones. A lot of warps were even removed from the desert. Now if you wanted to restore Morroc, you not only had to modify your client, but you had to get an older version of eAthena to copy the old warps and NPCs from, while disabling the new ones. So far so doable with some tinkering, but then Morroc and the desert went through another change in RE. Now you had three Morrocs to choose from for the client, changes in more warps and NPCs, and even new entrances and exits for Ant Hell in the desert.

I recently went through puzzling together a pre-RE server with a restored Morroc, and while doable, it's simply not fun, and the entire time I was wondering why nobody had thought of a better way to handle these changes, to make them toggleable more easily. I mean... they didn't even leave the old code in, they just removed it, as if it wasn't important anymore.

Same thing goes for how they handled RE, with separate sub-folders, duplicate scripts and databases, where it gets really annoying to even find what you're looking for.

**Solution**

For a potential solution, let me introduce you to the "feature db" we have in our Mabinogi emulator, Aura. It's basically a list of toggles for every feature that changed or was added at some point in Mabinogi's life. Organized in a tree of "Generation" and "Season" updates (major and minor updates in Mabinogi), a user can enable and disable entire updates by changing a single bool. Don't want a certain skill to be attainable? Change one bool. Don't want pets? Change one bool. Want updates 1\~5 and 7\~8, but not 6, because that changed several systems, which you didn't like? Enable all bools but that one.

These are then checked in the code and the databases to adjust the game to whatever the user wants. Monsters will drop certain items that have a feature setting only if that feature is enabled, dungeons can only be entered if they're enabled, and if an item changed completely, in addition to the old one you add the new version with a feature check, so it's loaded and overrides the older version if the feature is enabled.

```js
// Generation 2
//---------------------------------------------------------------------------
{ name: "G2", enabled: false, children: [
	// Season 1
	{ name: "G2S1", enabled: true, children: [
		// Enables G2 mainstream quest line
[...]

PR #136, opened 2017-09-19
Updated warps

Hirasu:
Add new warps and comments.


exectails:
It would be nice if all our scripts had a proper header, ideally with a history/changelog, akin to eAthena, like our custom scripts already have. That will be even more important should #112 come into play.

PR #137, opened 2017-09-20
GM Command RecallAll added

lRaikeNl:
Recalls all players that are currently online.


celophi:
As I understand, a similar command exists in emulators of other games; however, it is important to note that the client currently can only handle so many character instances in a single location. The limits of this will vary by system, but there is a debug command that can be executed to test concentration of characters in a single location.

I'm not available to provide details of that command right now, but I will update this comment later with details.

lRaikeNl:
I will let the command check if there are 150 + players on the server, if yes, then the command should abort.

All I found was var zoneMaxPcCount = 150; on LoginPacketSender.cs, I would like to make this a constant. But don't know in which file I should put it.

exectails:
@lRaikeNl The numbers are currently purely cosmetic on Melia, though it is good to know that the client won't stop players from logging in.

@celophi That might not a bad idea, although that problem basically exists on any MMO. Be it 5, 50 or 500 players, any client will complain if you put too many characters on the screen at once, and like you said, it's also largely dependent on the system.

Should we handicap the user by preventing them to use a command because *we* think they went overboard? Or should we consider it a user error? What other commands would this apply to? I mean, I had a lot of fun playing with `spawn` on eAthena years ago, finding out how many monsters server and client were able to handle before everything crashed^^

src/Shared/Const/Channel.cs
```diff
@@ -0,0 +1,13 @@
...
+// For more information, see license file in the main folder
+
+namespace Melia.Shared.Const
+{
+	/// <summary>
+	/// Channel constants
+	/// </summary>
+	public static class Channel
+	{
+		public const int ZoneMaxPcCount = 150;
```
exectails: The thought of moving this magic number to a central place is certainly not wrong, but what it should probably actually be is an option in the conf files or the server db, not a const in the source.

src/ChannelServer/World/WorldManager.cs
```diff
@@ -177,5 +177,23 @@ public Character GetCharacterByTeamName(string teamName)
 
 			return null;
 		}
+
+		/// <summary>
+		/// Returns all Characters that are currently online.
+		/// </summary>
+		public Character[] GetCharacters()
```
exectails: Nothing wrong with this code, but there's a shorter and potentially slightly faster way to do it if you're interested:
```csharp
lock (_mapsLock)
	return _mapsId.Values.SelectMany(a => a.GetCharacters()).ToArray();
```

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -542,6 +543,44 @@ private CommandResult HandleRecall(ChannelConnection conn, Character sender, Cha
...
+			foreach (var character in characters)
+			{
+				if (character.Id != target.Id)
+				{
+					var location = target.GetLocation();
+					character.Warp(location);
+
+					character.ServerMessage("You've been warped to {0}.", location);
+
+					return CommandResult.Okay;
```
exectails: By returning here, from inside the loop, you're basically canceling the command after the first successful warp, recalling only the first character.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -49,6 +49,7 @@ public GmCommands()
 			Add("go", "<destination>", this.HandleGo);
 			Add("goto", "<team name>", this.HandleGoTo);
 			Add("recall", "<team name>", this.HandleRecall);
+			Add("recallall", string.Empty, this.HandleRecallAll);
```
exectails: When all other commands use an empty string (`""`), a `string.Empty` sticks out a little bit, going against the convention. Btw, there's not much to actually gain from using `string.Empty`, it's just 10 more characters for no reason^^

lRaikeNl:
I will implement the suggestions you made, once I'm at home. (@work)

src/ChannelServer/Util/GmCommands.cs
lRaikeNl: Thanks for the information, I could only test with 2 instances of ToS.

src/Shared/Const/Channel.cs
lRaikeNl: Should I create a *.conf file called channel.conf for that? No includes like the world.conf, or with a channel sub-folder?

exectails: >so it restricts accessing clients when the limit is reached.

Yes, that's not a thing yet because server's don't communicate with each other yet, and the entire system of managing channels is non-existent. We don't have to go out of our way to shoehorn that in now, when it will change again soon anyway.

Where exactly that option should go isn't clear as a result either. Either a channel.conf, or a field on the servers data, which the login server has access to as well. Though we'll probably need a channel.conf either way in that case, because right now our channel"s" don't even know which channel they are.

Based on all this, I would say it would fit best into login.conf for the moment, as that's the server that needs the information, controls who goes where, whether you can actually login, etc. As long as we only have that one channel, that's the best location.

exectails:
One more thing regarding limiting the amount of characters. I believe the entire functionality of this command is questionable. Of course it doesn't hurt to have it, but its use is very limited from the get-go. When would you even use it?

The most likely response is "events", but that only works on very small servers. Generally you don't want to disrupt your player's game against their will, even during an event. Rather, you would probably tell people where to go, and if necessary, do an area warp.

As a result, this is a command that's designed for small servers, where the admin quickly wants to gather every one of maybe a few dozen players tops in one spot for some reason. You're doing something very wrong if you attempt to recall hundreds or thousands of players, and that will, or at least should, be obvious to anybody. If they do go against that, they probably have a reason for it, and I don't know if we should actually attempt to think for the user, setting an arbitrary limit.

lRaikeNl:
The command would be for administrative purposes only. Expecially if there are only GM's, Administrators or Beta Users on the Server. So everyone could check bugs on a map for example.
I'd also set the command to Authentification Level 99.

It's easier to use this command than calling like 10 people with 10 single commands.

Also I'd implement a Map Recall, like recall everyone that is currently on the map. But that will be an other case/commit.

exectails:
>The command would be for administrative purposes only

Interesting, never seen it used that way, only for events. Case in point, the suggestion of adding this command on an issue on @celophi's fork:


I guess your use-case supports my basic idea though, it's not something you would generally use unless appropriate. Whether it should be auth 50 or 99 by default really depends on how it's primarily used I think, as that will dictate who will use it the most.

Under the premise that it will primarily be used by GMs for events, on small servers, which there will probably be more of than large ones, 50 might seem more appropriate.

lRaikeNl:
> Interesting, never seen it used that way, only for events.

We will some time get to the point where we need to check the server-side with more people (playing), so that command could be useful.

exectails:
Certainly, like I said, it can't hurt to have it, even if *I* can't think of a situation where I would actually use it^^

PR #139, opened 2017-09-22
GM Command RecallMap added

lRaikeNl:
Warps all players on the map to target's location.

### DIFF
--- a/src/ChannelServer/Util/GmCommands.Handlers.cs
+++ b/src/ChannelServer/Util/GmCommands.Handlers.cs
@@ -676,32 +676,28 @@ namespace Melia.Channel.Util
 			if (args.Length > 2)
 				return CommandResult.InvalidArgument;
 
-			Character[] characters = null;
-			
-			// Recall same map.
-			if (args.Length == 1)
-				characters = target.Map.GetCharacters(a => a != target);
+			var map = target.Map;
 
-			// Recall other map
-			if (args.Length == 2)
+			// TODO: Once we have support for channels and map servers,
+			//   add warp from other servers and restrict recall to
+			//   channel's max player count.
+			if (args.Length > 1)
 			{
-				// TODO: Once we have support for channels and map servers,
-				//   add warp from other servers and restrict recall to
-				//   channel's max player count.
+				// Search for map by name and id
+				if (int.TryParse(args[1], out var mapId))
+					map = ChannelServer.Instance.World.GetMap(mapId);
+				else
+					map = ChannelServer.Instance.World.GetMap(args[1]);
 
-				// Get map id
-				if (!int.TryParse(args[1], out int mapId))
-					return CommandResult.InvalidArgument;
-
-				var map = ChannelServer.Instance.World.GetMap(mapId);
+				// Check map
 				if (map == null)
 				{
-					sender.ServerMessage("Invalid map id.");
+					sender.ServerMessage("Unknown map.");
 					return CommandResult.Okay;
 				}
-
-				characters = map.GetCharacters(a => a != target);
 			}
+
+			var characters = map.GetCharacters(a => a != target);
 
 			// Check for characters
 			if (!characters.Any())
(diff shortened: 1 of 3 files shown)


src/ChannelServer/Util/GmCommands.Handlers.cs
```diff
@@ -687,23 +742,7 @@ private CommandResult HandleRecallAll(ChannelConnection conn, Character sender,
...
-			if (sender == target)
-			{
-				sender.ServerMessage("You have called {0} characters to your location.", characters.Length);
-			}
-			else
-			{
-				sender.ServerMessage("You have called {0} characters to target's location.", characters.Length);
-				target.ServerMessage("{1} called {0} characters to your location.", characters.Length, sender.TeamName);
-			}
+			RecallCharacters(sender, target, characters);
```
exectails: This should either have a `this.`, or the method should be static, as per conventions. And wouldn't the function be better placed in the "functions part" of the commands class, instead of in between the handlers?

src/ChannelServer/Util/GmCommands.Handlers.cs
```diff
@@ -48,6 +48,7 @@ public GmCommands()
 			Add("go", "<destination>", this.HandleGo);
 			Add("goto", "<team name>", this.HandleGoTo);
 			Add("recall", "<team name>", this.HandleRecall);
+			Add("recallmap", "", this.HandleRecallMap);
```
exectails: It might be useful to have take an optional parameter, so you can recall everybody on a specific map to you. For example, to get everyone out of an event map.

lRaikeNl: Good idea, I'll implement that.

src/ChannelServer/Util/GmCommands.Handlers.cs
lRaikeNl: I was thinking about the "functions part" of the commands class, but as you said, it would be better to have handlers and the constructor on the same file, so the function should be there as well.

most likely direclty above or below the functions that use it so it can be found quickly.

I'll add a "this." to it.

exectails: My line of thought was that the handlers part has the handlers, incl the list of handlers, while the other part has the managing and potential helper functions. Maybe the partial was a mistake, that only complicates placement^^

lRaikeNl: The partial on this place was definitely no mistake. Like I said, I was also thinking about placing the function on the "functions part", and so be it.

exectails: I mean I'm not sure either^^ Technically you could even put such helper functions into a third partial. I liked that idea, but not it feels complicated =P

lRaikeNl: I wouldn't recommend more than 2 partials, if it is avoidable. Exceptions would be formulars, but we don't use them on melia for now. Let's just let the function where it is right now. The usability is good there where it is.

src/ChannelServer/Util/GmCommands.Handlers.cs
```diff
@@ -662,6 +663,60 @@ private CommandResult HandleRecall(ChannelConnection conn, Character sender, Cha
...
+		/// <returns></returns>
+		private CommandResult HandleRecallMap(ChannelConnection conn, Character sender, Character target, string command, string[] args)
+		{
+			if (args.Length > 2)
+				return CommandResult.InvalidArgument;
+
+			Character[] characters = null;
+			
+			// Recall same map.
+			if (args.Length == 1)
```
exectails: Looks good, but I was wondering what you think about how I would've done it.
```csharp
var map = sender.Map;
if (args.Length > 1)
{
	// Search for map by name and id
	if (int.TryParse(args[1], out var mapId))
		map = ChannelServer.Instance.World.GetMap(mapId);
	else
		map = ChannelServer.Instance.World.GetMap(args[1]);

	// Check map
	if(map == null)
	{
		sender.ServerMessage("Unknown map.");
		return CommandResult.Okay;
	}
}

var characters = map.GetCharacters(a => a != target);
```

lRaikeNl: Pushed, looks good for me.

Just don't forget the comment to check for max. player count. ^^

PR #145, opened 2017-10-18
Add Player properties with calculation

Hirasu:
Added Critical Atk/Rate, HP/SP Recovery amount, Sub Atk, Def, Mdef, Accuracy, Evasion, Block, Block Penetration, Critical Resistance, Aoe Defence Ratio, Movement Speed.

Lv1
Lv15


src/ChannelServer/World/Character.cs
```diff
@@ -407,6 +407,8 @@ public int Stamina
 		/// Used together with skill's splash rates.
 		/// TODO: Figure out where it comes from (official name: SR).
 		/// TODO: Check if it changes with jobs, items, stances, etc.
+		/// Item Bonus: Aoe Attack Ratio (SR)
+		/// Changed to 'AoeRatio'
```
exectails: Why not remove it then?

src/ChannelServer/World/Character.cs
```diff
@@ -528,7 +612,7 @@ public int MaxMAtk
 		{
 			get
 			{
-				var baseValue = 20;
+				var baseValue = 22;
```
exectails: Where did you get these 22s from? The client files (calc_property_pc) uses 20 o.o

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+				value += byBuffs + byRateBuffs;
+
+				return (int)value;
+			}
+		}
+
+		/// <summary>
+		/// Returns Block Penetration.
+		/// </summary>
+		public int BlockPen
```
exectails: Feel free to use more descriptive names, like "Defense" and "BlockPenetration". No reason to cut corners, IntelliSense will complete it anyway^^

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+		public int AoeRatio
+		{
+			get
+			{
+				var baseValue = 4;
+				var byItem = 0f; // TODO: Cached Aoe Attack Ratio for inventory/equip  ("SR")
+
+				var value = baseValue + byItem;
+
+				//if(hasBuff("Finestra") && hasAttribute("Finestra: Splash"))
```
exectails: This is most-likely not how the buffs are gonna work. For example, the buff effects for "Finestra" would be added to the character's list of buffs on activation, which are then added to the value via byBuffs and other variables. I only included those hasBuff comments in the previous properties for future reference, because it was hardcoded in the client files as well.

It would be pretty annoying if we had to hardcode every single buff like this^^

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+				//	value += 3
+
+				// Buffs: "SR_BM"
+				var byBuffs = 0;
+
+				// Rate buffs: TOS don't have a Aoe attack Ratio rate bonus (correct me if im wrong)
```
exectails: What are you basing these properties on? I would've guessed the aforementioned Lua file, but this doesn't look like `SCR_Get_SR`, which I would guess would be the equivalent, as it's missing things.

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+				value += byBuffs;
+
+				return (int)value;
+			}
+		}
+
+		/// <summary>
+		/// Returns HP Recovery amount.
+		/// </summary>
+		public int RHp
```
exectails: Definitely a candidate for a better name.

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+		{
+			get
+			{
+				var baseValue = 1;
+				var level = this.Level;
+				var stat = this.Con;
+
+				var ByLevel = level / 2f;
```
exectails: This variable should start with a lower case b, as per our coding conventions.

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+		}
+
+		/// <summary>
+		/// Returns SP Recovery amount.
+		/// </summary>
+		public int RSp
+		{
+			get 
+			{
+				var baseValue = 3;
```
exectails: I don't think this property has a base value.

src/ChannelServer/World/Character.cs
```diff
@@ -555,6 +639,392 @@ public int MaxMAtk
...
+				var level = this.Level;
+				var stat = this.Spr;
+
+				var ByLevel = level / 2f;
```
exectails: I think this is incorrect, the client files have `local byLevel = lv * jobObj.JobRate_SP;` for `byLevel`.

src/ChannelServer/World/Character.cs
```diff
@@ -622,12 +1094,28 @@ private void AddReferenceProperties()
...
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.DR, () => this.Evasion));
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.BLK, () => this.Block));
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.BLK_BREAK, () => this.BlockPen));
+
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.DEF, () => this.Def));
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.MDEF, () => this.Mdef));
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.CRTDR, () => this.CritRes));
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.SDR, () => this.AoeRatioDef));
+
+			this.Properties.Add(new RefFloatProperty(PropertyId.PC.MSPD, () => this.GetSpeed()));
```
exectails: Are you sure MSPD is the speed? "SPD" surely stands for "speed", but what's the M?

Hirasu: Movement speed, after that its appear in the F1 window.

src/ChannelServer/World/Character.cs
Hirasu: Looks like its out-dated, the http://www.tosbase.com/tools/stat-calculator/ use the same as the calc_property_pc and its always 2 lower then on the offi server.

src/ChannelServer/World/Character.cs
Hirasu: I wasn't sure about that, I'll change it

src/ChannelServer/World/Character.cs
Hirasu: I didn't looked to the client files, I should do that too...

src/ChannelServer/World/Character.cs
Hirasu: Nevermind, the problem was I didn't thought about the Collection bonus, using a new account now.

src/ChannelServer/Util/GmCommands.Handlers.cs
```diff
@@ -519,6 +519,7 @@ private CommandResult HandleSpeed(ChannelConnection conn, Character sender, Char
 
 			target.Speed = speed;
 			Send.ZC_MOVE_SPEED(target);
+			Send.ZC_OBJECT_PROPERTY(target, PropertyId.PC.MSPD);
```
exectails: Does this make the line above redundant?

src/ChannelServer/Database/ChannelDb.cs
```diff
@@ -90,6 +90,9 @@ public Account GetAccount(string name)
 		public Character GetCharacter(long accountId, long characterId)
 		{
 			var character = new Character();
+			int LastHp;
```
exectails: These should be camelCase, as per our coding conventions.

src/ChannelServer/World/Character.cs
```diff
@@ -306,251 +306,1830 @@ public int ClassLevel
...
+				var byBonus = 0;
+
+				var value = byLevel + byStat + byItem + byBuffs + byBonus;
+
+				return (int)value;
+			}
+		}
+
+		/// <summary>
+		/// Gets or set Job SpRate
```
exectails: This comment could be better, what exactly is the "SP Rate"? Think about what you would want to know about that property when writing the comment.

src/ChannelServer/World/Character.cs
```diff
@@ -306,251 +306,1830 @@ public int ClassLevel
...
+		/// Beware, some of these are reference properties, that can't be
+		/// set directly. Use this object's actual properties whenever
+		/// possible.
+		/// </remarks>
+		public Properties Properties { get; } = new Properties();
+
+		/// <summary>
+		/// Returns minimum physical ATK.
+		/// </summary>
+		public int MinPAtk
```
exectails: Are these all based on the client files now?

src/LoginServer/Database/LoginDb.cs
```diff
@@ -155,16 +155,8 @@ public void CreateCharacter(long accountId, Character character)
 					cmd.Set("bz", character.BarrackPosition.Z);
 
 					cmd.Set("hp", character.Hp);
-					cmd.Set("maxHp", character.MaxHp);
 					cmd.Set("sp", character.Sp);
-					cmd.Set("maxSp", character.MaxSp);
 					cmd.Set("stamina", character.Stamina);
-					cmd.Set("maxStamina", character.MaxStamina);
-					cmd.Set("str", character.Str);
```
exectails: I think this change is problematic. I get the idea, and it might be what officials do, but sending a job stat that comes from the static data instead of a "free flowing" value is *highly* inflexible. Getting to 1 on each stat with these changes would basically require hacks, but that will be something that users will probably want, to allow completely free stat distribution.

exectails:
So... _many_... _properties_... @_@

src/ChannelServer/World/Character.cs
Hirasu: Yes

src/ChannelServer/Util/GmCommands.Handlers.cs
Hirasu: This is for the F1 window so that you can see the changed Speed. "Movement Speed: XX"

exectails: So it doesn't actually change the movement speed? Alright.

sql/main.sql
```diff
@@ -32,11 +32,11 @@ CREATE TABLE IF NOT EXISTS `characters` (
...
-  `str` float NOT NULL DEFAULT '0',
-  `con` float NOT NULL DEFAULT '0',
-  `int` float NOT NULL DEFAULT '0',
-  `spr` float NOT NULL DEFAULT '0',
-  `dex` float NOT NULL DEFAULT '0',
+  `str` float NOT NULL DEFAULT '1',
+  `con` float NOT NULL DEFAULT '1',
+  `int` float NOT NULL DEFAULT '1',
+  `spr` float NOT NULL DEFAULT '1',
+  `dex` float NOT NULL DEFAULT '1',
```
Hirasu: I changed it back to the original one.

exectails:
Sorry for losing sight of this PR, I think I was waiting for some more fixes, though I don't remember what they might've been. Right now my only concern is how many properties there are right in the Character class, that seems really unwieldy. Do you think we should maybe move the stats/properties to their own sub-class?

exectails:
Closing this, as it has been merged, in modified form, via merging another fork in 538ed68206091903238d9dc9104627f0ec2ae18d.

PR #153, opened 2018-05-14
Include monsters spawn

Danil0v3s:
I've created the primary structure so the spawnzones can work based on [9e8e0069e1](https://github.com/gastonb/melia/tree/9e8e0069e1a7f2b95d981a4d0cf17c6c5bbe20db)
Tested and working fine.
I've noticed that the spawn is adding monsters in the same place, but I think it is caused by the configuration. I need to check about the sectors so the server can know where to place monsters.

### DIFF
--- a/src/ChannelServer/Scripting/ScriptManager.cs
+++ b/src/ChannelServer/Scripting/ScriptManager.cs
@@ -98,6 +98,9 @@ namespace Melia.Channel.Scripting
 			Register(addnpc);
 			Register(addwarp);
 
+			//Spawn
+			Register(addspawn);
+
 			// General
 			Register(var);
 			Register(logdebug);
@@ -136,6 +139,8 @@ namespace Melia.Channel.Scripting
 			Register(changehair);
 			Register(spawn);
 			Register(levelup);
+
+
 		}
 
 		/// <summary>
@@ -607,6 +612,39 @@ namespace Melia.Channel.Scripting
 			Melua.lua_pop(L, 1);
 
 			Log.Debug(msg);
+
+			return 0;
+		}
+
+		private int addspawn(IntPtr L)
+		{
+			SpawnData spawnData = new SpawnData();
+
+			spawnData.spawnName = Melua.luaL_checkstring(L, 1);
+			spawnData.mapName = Melua.luaL_checkstring(L, 2);
+			spawnData.monsterName = Melua.luaL_checkstring(L, 3);
+			spawnData.xMin = Melua.luaL_checkinteger(L, 4);
+			spawnData.xMax = Melua.luaL_checkinteger(L, 5);
+			spawnData.zMin = Melua.luaL_checkinteger(L, 6);
+			spawnData.zMax = Melua.luaL_checkinteger(L, 7);
+			spawnData.count = Melua.luaL_checkinteger(L, 8);
+			spawnData.countVariation = Melua.luaL_checkinteger(L, 9);
+			spawnData.respawnTime = Melua.luaL_checkinteger(L, 10);
+
+			spawnData.isFixedLocation = Melua.luaL_checkinteger(L, 11) > 0;
+			var x = (float)Melua.luaL_checknumber(L, 12);
+			var y = (float)Melua.luaL_checknumber(L, 13);
+			var z = (float)Melua.luaL_checknumber(L, 14);
+			spawnData.Position = new Position(x, y, z);
+
+			var direction = Melua.luaL_checkinteger(L, 15);
+			spawnData.Direction = new Direction(direction);
+
+			var map = ChannelServer.Instance.World.GetMap(spawnData.mapName);
+			if (map == null)
+				return Melua.melua_error(L, "AddSpawn: {1} Map '{0}' not found.", spawnData.mapName, spawnData.spawnName);
+
+			map.AddSpawnZone(spawnData);
 
 			return 0;
 		}
--- a/src/ChannelServer/World/Map.cs
+++ b/src/ChannelServer/World/Map.cs
@@ -32,6 +32,12 @@ namespace Melia.Channel.World
 		/// </summary>
 		private Dictionary<int, Monster> _monsters;
 
+
+		/// <summary>
+		/// Collection of Spawn Zones
+		/// </summary>
+		private List<SpawnZone> _spawns;
+
 		/// <summary>
 		/// Map name.
 		/// </summary>
@@ -54,6 +60,7 @@ namespace Melia.Channel.World
 		{
 			_characters = new Dictionary<int, Character>();
 			_monsters = new Dictionary<int, Monster>();
+			_spawns = new List<SpawnZone>();
 
 			this.Id = id;
 			this.Name = name;
@@ -93,6 +100,12 @@ namespace Melia.Channel.World
 				foreach (var character in _characters.Values)
 					character.LookAround();
 			}
+		}
+
+		private void ProcessMonsters()
+		{
+			foreach (var monster in _monsters.Values)
+				monster.Process();
 		}
 
 		/// <summary>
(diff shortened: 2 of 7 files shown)


exectails:
Thank your for your PR. I've noticed several instances of your code not following our coding conventions, so I'd like to ask you to take a look at it: https://github.com/aura-project/melia/blob/master/CONTRIBUTING.md#coding-conventions
Some examples are:
- Favor `var` over explit typing.
- Use PascalCase for public properties and fields.
- Use properties in favor of fields.
- Use `this.` to access public properties and fields.

src/ChannelServer/Scripting/ScriptManager.cs
```diff
@@ -611,6 +616,39 @@ private int logdebug(IntPtr L)
...
+			spawnData.respawnTime = Melua.luaL_checkinteger(L, 10);
+
+			spawnData.isFixedLocation = Melua.luaL_checkinteger(L, 11) > 0;
+			var x = (float)Melua.luaL_checknumber(L, 12);
+			var y = (float)Melua.luaL_checknumber(L, 13);
+			var z = (float)Melua.luaL_checknumber(L, 14);
+			spawnData.Position = new Position(x, y, z);
+
+			var direction = Melua.luaL_checkinteger(L, 15);
+			spawnData.Direction = new Direction(direction);
```
exectails: Does a spawn need a direction? From what I can tell this value isn't used at all.

src/ChannelServer/World/Map.cs
```diff
@@ -95,6 +102,12 @@ private void UpdateVisibility()
 			}
 		}
 
+		private void ProcessMonsters()
```
exectails: What is this method's purpose? Looks like it's never used.

src/ChannelServer/World/Monster.cs
```diff
@@ -189,7 +191,18 @@ public void Kill(Character killer)
 			this.DisappearTime = DateTime.Now.AddSeconds(2);
 			killer.GiveExp(exp, classExp, this);
 
+			if (this.spawnZone != null)
+			{
+				this.spawnZone.NotifyEntityDeath(this.Handle);
```
exectails: This will certainly work, but if I may make a suggestion, what do you think about using events? If Monster had a `Died` event that is raised here, the spawner could subscribe to that when the monster is spawned. That's much more flexible, can be used by any number of other methods and classes, and the Monster doesn't need a reference to its spawn zone, reducing dependencies. The zone simply gets notified about the death and spawns another monster.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+{
+
+
+	public struct SpawnData
+	{
+		public string spawnName;
+		public string mapName;
+		public string monsterName;
+		public int count;
+		public int countVariation;
```
exectails: You should always document code which's behavior isn't immediately obvious. Fields like this or methods like `ProcessMonsters` could be anything, and without documentation you need to read the code to learn what they're even doing.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+		public Direction Direction { get; set; }
+		public int respawnTime;
+		public int xMin, xMax, zMin, zMax;
+	}
+
+	public class SpawnZone
+	{
+
+		public int Id;
+		public SpawnData Data { get; set; }
```
exectails: Those two properties should probably have private setters or no setter at all, so they can't be accidentally changed from outside.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+		private bool _isInitialized;
+		private bool _respawnEnabled;
+		private Random rndGenerator;
+
+		public SpawnZone(SpawnData spawnData)
+		{
+			this.Data = spawnData;
+			this.MonsterData = ChannelServer.Instance.Data.MonsterDb.Find(this.Data.monsterName);
+
+			if (this.MonsterData == null)
```
exectails: It would be better to (also) verify this when the spawn zone is created from the script, to let the scripter know immediately that the monster doesn't exist. Otherwise they potentially have to go search for which script even created this spawner.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		public void Init()
+		{
+			if (_map == null || this.MonsterData == null)
+				return;
+
+			// Calculate count of spawn entities
+			Random rnd = new Random();
+			_calculatedTotalEntities = this.Data.count + rnd.Next(this.Data.countVariation);
```
exectails: You basically always generate a random number to spawn for the entire session? I know monsters spawn counts seem random, but from my experience they vary based on the amount of players, or rather kills in the area. The more monsters are killed in quick succession the more monsters spawn, to account for the increase in player numbers.

An implementation of that would probably use a base count and increase the respawn number by a certain amount, maybe double, based on the kill frequency.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+
+		public void DoSpawn(Object obj)
+		{
+			Log.Debug("DoSpawn");
+			if (!_isInitialized && !_respawnEnabled)
+				return;
+
+
+			Monster monster = new Monster(this.MonsterData.Id, Shared.Const.NpcType.Monster);
+			if (monster == null)
```
exectails: It's impossible for monster to be null here because you just created it. If something went wrong during creation you would get thrown out by an exception. There are more unnecessary null checks like this in this PR.

src/ChannelServer/World/SpawnZone.cs
```diff
@@ -0,0 +1,219 @@
...
+
+				if (IsInPolygon(xPos, zPos))
+				{
+					rndPosition = new Position(xPos, 260, zPos);
+					return true;
+				}
+
+				if (i == 50)
+				{
+					Log.Warning("Polygon seems complicated in spawn {0} {1}", this.Id, this.Data.spawnName);
```
exectails: It looks you're using a simple rectangle for the spawn area, there shouldn't be a reason to use loops or have warnings for complicated polygons, because you will definitely find a random position in a rectangle. There's not even a need for the "IsInside" check.

src/ChannelServer/World/TasksPoolManager.cs
```diff
@@ -0,0 +1,87 @@
...
+namespace Melia.Channel.World
+{
+
+	public class CbFunctionData
+	{
+		public TasksPoolManager.TimerCallbackFunction cbFunction;
+		public Object obj;
+	}
+
+	public class TasksPoolManager
```
exectails: What's the specific purpose for creating your own task manager, is something wrong about `System.Threading.Tasks.Task`?

src/ChannelServer/World/Monster.cs
daniloleemes: Interface maybe?

exectails: What are you proposing exactly?

exectails:
Are you still working on this?

daniloleemes:
I want to contribute more and fix this, but now I dont have much time... Preparing for a hackathon.
I'll see what I can do tonight

src/ChannelServer/World/TasksPoolManager.cs
daniloleemes: I dunno how to use that

exectails: Instead of that entire custom class you would for example do this:
```csharp
Task.Delay(this.Data.respawnTime * 1000).ContinueWith(this.DoSpawn);
```
https://msdn.microsoft.com/en-us/library/system.threading.tasks.task%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396

exectails:
Closing this, as it has been merged, in modified form, via merging another fork in 538ed68206091903238d9dc9104627f0ec2ae18d.

PR #155, opened 2018-07-29
[WIP] .NET Core conversion

redbaty:
This should be working, although I have not tested it through many scenarios.
Most changes were made to the code compilation process on swebs, and for that reason, the project is now included in the solution as well.


redbaty:
Hmm, apparently the Travis configuration has to be rebuilt to fit the new csproj format, I'll try and convert it as well.

PS: What do you guys think of Ef Core, is there a reason why you hadn't used the entity framework system? It would fair pretty well as for performance using an ef core + Dapper combo.

PR #162, opened 2018-12-23
Adds unique session keys and client integrity checks.

celophi:


system/conf/login.conf
```diff
@@ -2,6 +2,9 @@
 // Configuration file
 // --------------------------------------------------------------------------
 
+// IPF MD5 checksum
+ipf_checksum: ce9d9325d0fd35ceae14ce5261d75c98
```
celophi: This needs to be generated any time specific IPF archives change. In my own extractor, I generate this one time. I'm trying to think what would be appropriate for Melia...

Probably the easiest thing to do would be to add the generation code to IPFBrowser.

exectails: I usually don't include checks like these if it means that you have to update it regularly, especially if it's not easy to obtain for the common user. Do you think that's worth the trouble? I think it should probably at least be optional, like don't check if the option is empty or something. What do you think?

celophi: If it is optional, I think that is fine. I think it is more important that the functionality is documented and exists, so that someone isn't left wondering what this part of the packet represents or is used for.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -415,6 +415,16 @@ public void CB_CHECK_CLIENT_INTEGRITY(LoginConnection conn, Packet packet)
...
+
+				// Even though the integrity check fails, send these packets as well.
+				// This is done because the client is in a hanging state waiting for these packets.
+				// Without these, the invalid IPF message will not be visible.
+				Send.BC_COMMANDER_LIST(conn);
+				Send.BC_NORMAL_ZoneTraffic(conn);
+				Send.BC_NORMAL_TeamUI(conn);
+
+				// Terminate any further requests from the client at this point.
+				conn.IgnorePackets = true;
```
celophi: I added this after finding out that the IPF mismatch message can be easily bypassed. Unfortunately, this check cannot be done until *after* the client actually enters the barrack, so there has to be something done about preventing the client from making further requests.

src/LoginServer/Network/LoginPacketHandler.cs
```diff
@@ -388,15 +391,41 @@ public void CB_BUY_THEMA(LoginConnection conn, Packet packet)
...
+			var bytes = Encoding.UTF8.GetBytes(LoginServer.Instance.Conf.Login.IpfChecksum + conn.IntegritySeed);
+
+			var md5 = MD5.Create();
+			md5.TransformFinalBlock(bytes, 0, bytes.Length);
 
-			// Ignore for now.
-			// TODO: Add option for accepted checksums.
+			var result = BitConverter.ToString(md5.Hash).Replace("-", "").ToLower();
+
+			if (checksum != result)
```
exectails: Might want to lower the checksum string as well or use `string.Compare` just in case the client ever starts to upper case it for some reason.

src/LoginServer/Util/Configuration/Files/Login.cs
```diff
@@ -18,13 +18,30 @@ public class LoginConfFile : ConfFile
...
+
 		public void Load()
 		{
 			this.Require("system/conf/login.conf");
 
+			this.IpfChecksum = this.GetString("ipf_checksum");
 			this.StartLocation = this.LoadStartLocation();
 		}
 
+		private void LoadIpfSettings()
```
exectails: You're not calling this method, so `VerifyIpf` wouldn't be set. You're also getting `IpfChecksum` twice. You can put both directly into `Load`, like in the other conf files, `LoadStartLocation` is only a thing because it's more involved than a simple get call. You might also want to use default values in case some user deletes the options or conf files, and for `VerifyIpf` you can use `GetBool` instead of `GetString`. Saves you the mannual check and it supports "yes" instead of "true" (so fancy |D).

src/Shared/Network/Connection.cs
```diff
@@ -59,6 +71,14 @@ public Connection()
 
 			this.State = ConnectionState.Open;
 			this.Address = "?:?";
+			this.IntegritySeed = _random.Next();
```
exectails: You could use `RandomProvider.Get().Next()` here, saves you from creating an instance yourself, especially since it's never used again.

src/Shared/Network/Connection.cs
```diff
@@ -126,6 +146,12 @@ private void OnReceive(IAsyncResult result)
 					return;
 				}
 
+				// When set, ignore packets until the client eventually terminates the connection.
+				if (this.IgnorePackets)
+				{
+					this.BeginReceive();
```
exectails: You have to `return` if you don't want the packet to be handled, and I actually don't know what happens if you call `BeginReceive` twice =o What do you think about moving this further down where `HandlePacket` is called? You could wrap that call in an if checking for `IgnorePackets`, ignoring only the actual handling of the packets, and still get the parsing to make sure packets are valid.

PR #185, opened 2021-09-06
Basic Skill Implementation and Warehouse

SalmanTKhan:
Figured I'd create a pull request as my current code has some work on parties, which makes it even more confusing.


exectails:
Thank you for your PR. I see a a few issues with the code, such as violations to the coding conventions laid out in our contribution guidelines, but before we get to that, would it be possible for you to reorganize this PR a little and split it up? Like into party, skill, and warehouse for example? The amount of binary template updates also seems a little excessive, so I assume there's a lot of unrelated updates in here, is that right? Or is this reorganization what proved difficult, and made you say it's a "lost cause" earlier on the chat?

SalmanTKhan:
Sure I'll close this PR request and I'll try to make smaller PR requests with related stuff.

PR #186, opened 2021-09-06
Skill Handler

SalmanTKhan:
Round 2? Less unrelated code in this commit, now I know there are a couple of issues, but hence why I opened the PR, so you guys can point them out.


src/ChannelServer/World/Skill.cs
```diff
@@ -55,6 +55,11 @@ public int SpendSp
 			}
 		}
 
+		/// <summary>
+		/// Over Heat Counter
```
exectails, line 59: Some of your comments could be a little more descriptive. From the name of the property, you can already infer that this is the "OverHeatCount(er)". The comment should ideally tell you more about what's going to happen if you use use the property. We weren't great at documenting the code either back then, but we should make an effort to improve on this =) Even a simple "Gets or sets the overheat count." tells you more about what's going to happen. You get or set that value, that's it, no hidden side-effects. Though you could even go into more detail about what that actually means.

src/ChannelServer/World/Character.cs
```diff
@@ -2798,22 +2798,39 @@ public void LookAround()
...
+		/// <returns></returns>
+		private bool CanCast(Skill skill)
+		{
+			if (skill != null)
 			{
-				Send.ZC_OVERHEAT_CHANGED(this, skill);
-				Send.ZC_COOLDOWN_CHANGED(this, skill);
-				Send.ZC_SKILL_FORCE_TARGET(this, null, skill, 0);
+				// To Do move pre-cast logic into Character.CanCast?
+				if (skill.Data.BasicSp != 0)
```
exectails, line 2823: There's a `SpendSp` property on Skill, that does almost the same calculation you're doing here, and returns the SP necesary to use the skill. Wouldn't it be more intuitive to use that property, to have the skill tell you how much SP it requires?

src/ChannelServer/World/Character.cs
```diff
@@ -2798,22 +2798,39 @@ public void LookAround()
...
+				{
+					var requiredSp = (int)skill.Data.BasicSp + (skill.Data.LvUpSpendSp * skill.Level - 1);
+					if (this.Sp < requiredSp)
+						return false;
+					this.Sp -= (int)skill.Data.BasicSp;
```
exectails, line 2828: Since we'll surely need to modify the SP from other places as well in the future, wouldn't it be better to have like a `ModifySp` method on character, that we can use to both modify the SP and update them? That way you also can't accidentally forget to send the packet.

src/ChannelServer/World/Character.cs
```diff
@@ -2798,22 +2798,39 @@ public void LookAround()
...
+		}
+
+		/// <summary>
+		/// 
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <returns></returns>
+		private bool CanCast(Skill skill)
+		{
+			if (skill != null)
```
exectails, line 2820: Little style tip, if you check for null and then return false, you get rid of an indention level, making the code easier to read. The same can be done for the SP check and the require check, until you're left with only 1 or 2 indention levels.

src/ChannelServer/World/Skill.cs
```diff
@@ -55,6 +55,11 @@ public int SpendSp
...
+		/// </summary>
+		public int OverHeatCount { get; set; }
```
exectails, line 61: I opted to go with the name "Overheat" instead of "OverHeat" in the cool down database, since "Over" and "Heat" as separate words don't make a lot of sense to me. We should stick to one or the other, either correcting the Koreans' typo, or not.

src/ChannelServer/World/Skill.cs
```diff
@@ -80,7 +87,10 @@ public Skill(Character character, int skillId, int level)
 			this.Character = character;
 			this.Id = skillId;
 			this.Level = level;
+			this.OverHeatCount = 0; // So they don't reuse this value just cooldown value?
```
exectails, line 90: Class fields are initialized with their default value automatically, you don't need to do this manually.

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Melia.Channel.Network;
+using Melia.Shared.Data.Database;
+using Melia.Shared.World;
+
+namespace Melia.Channel.World
+{
+	interface ITargetedSkillHandler
```
exectails, line 12: You should always use appropriate access modifiers, which in this case would be `public` in my opinion. If we are going to add plugins in the future, we might want to be able to modify skill behavior from them, in which case we'll need access to these classes and interfaces.

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+				Send.ZC_COOLDOWN_CHANGED(caster, skill);
+				skill.OverHeatCount = 0;
+			}
+
+			var damage = caster.GetRandomPAtk() + 100;
+			Send.ZC_SKILL_FORCE_TARGET(caster, target, skill, damage);
+			// Unsafe cast for now we're only dealing with monsters
+			if (target != null)
+			{
+				var monster = (Monster)target;
```
exectails, line 45: I guess we need another interface, maybe `IAttackableEntity`, or `ISkillTarget`, which would have methods like `TakeDamage`, which we can then use for the target list. I assume monsters will use skills and skill handlers as well, and maybe some that players also have, in which case it would be inconvenient to manually handle the target type in the handlers.

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+			if (skill.Data.OverHeat != 0)
+				skill.OverHeatCount++;
+			Send.ZC_OVERHEAT_CHANGED(caster, skill);
+			if (skill.Data.OverHeat == 0 || skill.Data.OverHeat == skill.OverHeatCount)
+			{
+				Send.ZC_COOLDOWN_CHANGED(caster, skill);
+				skill.OverHeatCount = 0;
+			}
+
+			var damage = caster.GetRandomPAtk() + 100;
```
exectails, line 40: Is the attack damage still a problem?^^ Maybe we should take a look at the calculations and see what's wrong there. `MinPAtk` and `MaxPAtk` might be outdated or buggy.

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	class TargetedGroundSkillHandler : ITargetGroundSkillHandler
+	{
+		public void Handle(Skill skill, Character caster, Position position, List<IEntity> targets)
+		{
+			if (skill.Data.OverHeat != 0)
+				skill.OverHeatCount++;
+			Send.ZC_OVERHEAT_CHANGED(caster, skill);
+			if (skill.Data.OverHeat == 0 || skill.Data.OverHeat == skill.OverHeatCount)
```
exectails, line 59: Instead of manually checking variables, it would be clearer to add properties and methods to do these things. For example, this would be much more descriptive and easier to use:
```cs
if (skill.IsOverheated)
```
Otherwise, you have to guess a little what the code is supposed to do, which I did here, I'm not actually sure `IsOverheated` would be correct^^

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+			{
+				Send.ZC_COOLDOWN_CHANGED(caster, skill);
+				skill.OverHeatCount = 0;
+			}
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, position);
+			switch (skill.Id)
+			{
+				case 20007:
+					caster.Map.GetVisibleMonsters(caster);
+					var targets = caster.Map.GetAttackableMonstersInRange(position.X, position.Y, position.Z, (int)skill.Data.MaxRange).ToList<IEntity>();
```
exectails, line 93: It seems like a waste to allocate an entire new list, just to cast the list of monsters to a list of entities, especially since you don't seem to do anything that would require a list. Wouldn't it be more efficient to leave it as a monster list and make the type of `targets` on `ZC_SKILL_MELEE_GROUND` and `IEnumerable`? That way you don't need this allocation. Though if you were to add the interface I proposed for targets, you could return that from `GetAttackable` and use it without casts.

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+					Send.ZC_SYNC_EXEC(caster, 1234);
+					// Skill Cast Duration? Cancel's Visible Animation Otherwise you stay in casting animation
+					Send.ZC_SKILL_DISABLE(caster);
+					break;
+				default:
+					//Send.ZC_SYNC_START(caster, 1234, 1);
+					//Send.ZC_SYNC_END(caster, 1234, 0);
+					//Send.ZC_SYNC_EXEC_BY_SKILL_TIME(caster, 1234, skill.Data.HitDelay);
+
+					targets = caster.Map.GetAttackableMonstersInRange(position.X, position.Y, position.Z, (int)skill.Data.MaxRange).ToList<IEntity>();
```
exectails, line 125: If you use curly braces for switch cases you create new scopes and variable names aren't shared between them anymore. This can make it clearer where a variables come from, or rather, that they're created for this case, and aren't shared.

src/Shared/Data/Database/Skills.cs
```diff
@@ -124,9 +124,9 @@ protected override void ReadEntry(JObject entry)
 			info.UseType = entry.ReadEnum<SkillUseType>("useType");
 			info.Attribute = entry.ReadEnum<SkillAttribute>("attribute");
 
-			info.OverHeat = entry.ReadInt("overheat", 0);
-			info.OverHeatDelay = entry.ReadInt("overheatDelay", 0);
-			info.OverHeatGroup = entry.ReadString("overheatGroup", null);
+			info.OverHeat = entry.ReadInt("overheat");
```
exectails, line 127: The overheat fields are values not every skill has. This is communicated by using explicit default values for when the fields are missing. Removing them takes that away from the code.

exectails (commented):
Thank you for your effort. Here's a few potential problems I've noticed, with suggestions how they could be improved.

src/ChannelServer/Network/ChannelPacketHandler.cs
```diff
@@ -927,12 +930,12 @@ public void CZ_SKILL_TARGET(ChannelConnection conn, Packet packet)
 			var targetHandle = packet.GetInt();
 
 			var character = conn.SelectedCharacter;
+			var skill = character.Skills.Get(skillId);
 
-			if (character.CanCast(skillId))
+			if (skill != null)
```
exectails, line 935: Was this change intended? Now you aren't checking and reducing the SP anymore, are you?

src/ChannelServer/World/Map.cs
```diff
@@ -199,13 +199,13 @@ public void RemoveMonster(Monster monster)
...
-		public List<Monster> GetAttackableMonstersInRange(float x, float y, float z, int range)
+		public IEnumerable<Monster> GetAttackableMonstersInRange(float x, float y, float z, int range)
 		{
 			var position = new Position(x, y, z);
 
-			List<Monster> result;
+			IEnumerable<Monster> result;
 			lock (_monsters)
-				result = _monsters.Values.Where(a => a.Position.InRange2D(position, range) && a.NpcType == NpcType.Monster).ToList();
+				result = _monsters.Values.Where(a => a.Position.InRange2D(position, range));
```
exectails, line 208: Quick tip: You could simply return from inside the lock if there's nothing after it.

src/ChannelServer/World/Skill.cs
```diff
@@ -87,10 +115,9 @@ public Skill(Character character, int skillId, int level)
 			this.Character = character;
 			this.Id = skillId;
 			this.Level = level;
-			this.OverHeatCount = 0; // So they don't reuse this value just cooldown value?
 			this.Data = ChannelServer.Instance.Data.SkillDb.Find(skillId) ?? throw new ArgumentException($"Unknown skill '{skillId}'.");
-			this.CoolDownData = ChannelServer.Instance.Data.CoolDownDb.Find(this.Data.CoolDownGroup) ?? throw new ArgumentException($"Unknown skill '{skillId}' cooldown group '{this.Data.CoolDownGroup}'.");
-			this.OverHeatData = ChannelServer.Instance.Data.CoolDownDb.Find(this.Data.OverHeatGroup) ?? throw new ArgumentException($"Unknown skill '{skillId}' overheat group '{this.Data.OverHeatGroup}'.");
+			this.CooldownData = ChannelServer.Instance.Data.CoolDownDb.Find(this.Data.CooldownGroup) ?? throw new ArgumentException($"Unknown skill '{skillId}' cooldown group '{this.Data.CooldownGroup}'.");
```
exectails, line 119: Didn't we use Cool**D**own everywhere? Honestly, I never know whether I should write Cooldown or CoolDown. "The cooldown", as a thing, feels kind of correct. What do you say? 🤔

src/ChannelServer/World/SkillHandler.cs
```diff
@@ -5,129 +5,182 @@
...
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, targetPosition);
 			switch (skill.Id)
 			{
 				case 20007:
-					caster.Map.GetVisibleMonsters(caster);
-					var targets = caster.Map.GetAttackableMonstersInRange(position.X, position.Y, position.Z, (int)skill.Data.MaxRange).ToList<IEntity>();
+					var targets = caster.Map.GetAttackableMonstersInRange(targetPosition.X, targetPosition.Y, targetPosition.Z, (int)skill.Data.SplashRange);
 					var damage = caster.GetRandomPAtk() + 100;
-					Send.ZC_SKILL_MELEE_GROUND(caster, skill, position.X, position.Y, position.Z, targets, damage);
+					Send.ZC_SKILL_MELEE_GROUND(caster, skill, targetPosition.X, targetPosition.Y, targetPosition.Z, targets, damage);
```
exectails, line 136: Why not simply pass the position to the sender, instead of breaking it up?

src/ChannelServer/World/Skill.cs
SalmanTKhan: I hit the good ole google to double-check. I went with Cooldown since we're fixing the OverHeat to Overheat.
https://www.merriam-webster.com/dictionary/cooldown

src/ChannelServer/World/SkillHandler.cs
SalmanTKhan: Regarding the use position vs the broken up float values, should I just default to the position?

src/ChannelServer/World/Skill.cs
exectails: Yea, I assumed that was the reason. If you search both "cooldown" and "cool down", the latter has more results, but then there's dictionaries... Let's just go with your change^^

exectails: Oh! Wikipedia actually differentiates between the two, saying cooldown is a gaming term and cool down an excercise.

src/ChannelServer/World/SkillHandler.cs
exectails: Unless there's a reason to use separate values, passing the position is usually to be preferred. Less to type, easier to read.

src/ChannelServer/Network/ChannelPacketHandler.cs
SalmanTKhan: It wasn't an intended change when I removed CanCast, readded it within the SkillHandlers.

src/Shared/Data/Database/Skills.cs
SalmanTKhan: But the defaults handle that already, I guess you'd like to leave it more verbose. If we were setting different default values, it'd make more sense.
`internal static int ReadInt(this JObject obj, string key, int def = 0) { return (int)(obj[key] ?? def); }`
`internal static string ReadString(this JObject obj, string key, string def = "") { return (string)(obj[key] ?? def); }`

exectails: You're right, functionally it makes absolutey no difference. The only difference is that, when reading the code, you can see that default values are used, which communicates that these fields are optional in the data, and aren't necessarily set to something. It's not all that important, I just wanted to let you know why those defaults were set.

SalmanTKhan: Sounds good, will leave it in.

PR #188, opened 2021-09-10
Initial Buff Implementation

SalmanTKhan:
* Buff packets implemented
* Buff behavior implemented, over buffing and updating current buff timer on buff refresh.
* Buffs don't have any effects now


src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,70 @@
...
+		/// <param name="character"></param>
+		/// <param name="buffId"></param>
+		/// <param name="level"></param>
+		public Buff(Character character, BuffId buffId)
+		{
+			this.Character = character;
+			this.Id = buffId;
+			this.Handle = ChannelServer.Instance.World.CreateBuffHandle();
+			this.Data = ChannelServer.Instance.Data.BuffDb.Find(buffId) ?? throw new ArgumentException($"Unknown buff '{buffId}'.");
+			// TODO: Some durations have 0, so give them atleast a lifetime of 5 seconds for testing
```
exectails, line 57: Comments shouldn't go past 80 characters, as per our coding conventions.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+	/// Character buffs.
+	/// </summary>
+	public class Buffs
+	{
+		private readonly Dictionary<BuffId, Buff> _buffs = new Dictionary<BuffId, Buff>();
+
+		/// <summary>
+		/// The owner of this object.
+		/// </summary>
+		public Character Character { get; }
```
exectails, line 24: Should this be IEntity?

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+		}
+
+		/// <summary>
+		/// Adds given buff and updates the client. Replaces existing
+		/// buffs.
+		/// </summary>
+		/// <param name="buff"></param>
+		public void Add(Buff buff)
+		{
+			var buffExists = _buffs.ContainsKey(buff.Id);
```
exectails, line 58: You need to lock collections before accessing them, since the collection could change at _any_ point in a multi-threaded environment.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+		/// </summary>
+		/// <param name="buffId"></param>
+		/// <returns></returns>
+		public bool Remove(BuffId buffId)
+		{
+
+			lock (_buffs) {
+				var buff = _buffs[buffId];
+				if (_buffs.Remove(buffId))
+					// Update client...
```
exectails, line 90: Well... yea, update client, isn't that what you're doing with this call?^^"

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+		/// Removes buff with given id, returns false if it
+		/// didn't exist. Updates the client on success.
+		/// </summary>
+		/// <param name="buffId"></param>
+		/// <returns></returns>
+		public bool Remove(BuffId buffId)
+		{
+
+			lock (_buffs) {
+				var buff = _buffs[buffId];
```
exectails, line 88: You're not checking whether the buff exists, couldn't this cause an exception?

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+
+		/// <summary>
+		/// Removes buff with given id, returns false if it
+		/// didn't exist. Updates the client on success.
+		/// </summary>
+		/// <param name="buff"></param>
+		/// <returns></returns>
+		public bool Remove(Buff buff)
+		{
+			Log.Debug("Removing buff {0} which expired at {1}.", buff.Data.ClassName, buff.RemovalTime);
```
exectails, line 105: It's a test log, so it's not that important, but generally speaking, shouldn't such a log rather go where the buff is removed because it expired, instead of in the general purpose Remove method, that could be called from anywhere?

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+		}
+
+		/// <summary>
+		/// Add a Buff or Update a Buff if it exists
+		/// with a given id.
+		/// </summary>
+		/// <param name="buffId"></param>
+		public void AddOrUpdate(BuffId buffId)
+		{
+			if (!_buffs.ContainsKey(buffId)) {
```
exectails, line 124: Please format your code as per our coding conventions, with the curly brace on its own line. Have you considered getting an extension, like Format on Save? That way it happens automatically.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+				var buff = new Buff(this.Character, buffId);
+				Add(buff);
```
exectails, line 126: `this.`

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+			} else
+			{
+				Update(buffId);
+			}
+		}
+
+		/// <summary>
+		///
```
exectails, line 134: Empty documentation comments are sad to see. Worse than no comments in a way, because someone took the time to create the structure, but then didn't write anything^^" And a comment might be very helpful for someone, because even if you look at the code, it's not clear at a glance what exactly this method is doing.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+
+		/// <summary>
+		/// 
+		/// </summary>
+		/// <param name="buffId"></param>
+		public void Update(BuffId buffId)
+		{
+			lock (_buffs)
+			{
+				var buff = _buffs[buffId];
```
exectails, line 141: No existence check again.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,239 @@
...
+		/// <param name="buffId"></param>
+		/// <returns></returns>
+		public bool Has(BuffId buffId)
+		{
+			lock (_buffs)
+				return _buffs.ContainsKey(buffId);
+		}
+
+		/// <summary>
+		/// Returns current level of given buff, returns 0 if buff
```
exectails, line 210: You say "current level", but it always returns the level from the data, can the level actually change?

src/ChannelServer/World/Map.cs
```diff
@@ -90,8 +90,12 @@ private void UpdateVisibility()
 		{
 			lock (_characters)
 			{
+				// TODO move buff update from here to somewhere else?
```
exectails, line 93: Yea, to the component system^^ To be added.

src/ChannelServer/World/WorldManager.cs
```diff
@@ -9,6 +9,7 @@ namespace Melia.Channel.World
 	public class WorldManager
 	{
 		private int _handles = 0;
+		private int _buffhandles = 0;
```
exectails, line 12: Do buffs use their own handle id range, starting at 0? Or do they go into the same pool as the others?

src/ChannelServer/Network/Helpers/BuffHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+			packet.PutInt(1);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			packet.PutInt(buff.OverbuffCounter);
+			packet.PutInt(buff.Data.Duration);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			if (entity is Character)
```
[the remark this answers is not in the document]
SalmanTKhan: Currently, I wasn't too sure how the server adds buffs to monsters, needs more info, so I handled the case for characters. Rather than add un-needed fields to IEntity. An example is how the server handles monster's deaths vs player deaths
ZC_DEAD is sent with no killer info on player death while on monster death it is set.
```C#
		/// <summary>
		/// Broadcasts ZC_DEAD in range of entity, which kills it visually.
		/// </summary>
		/// <param name="entity">Entity to kill visually.</param>
		public static void ZC_DEAD(IEntity entity, IEntity killer)
		{
			var packet = new Packet(Op.ZC_DEAD);
			packet.PutInt(entity.Handle);
			packet.PutByte(0);
			if (killer != null)
				packet.PutByte(1);
			else
				packet.PutByte(0);
			packet.PutByte(0);
			packet.PutByte(0);
			packet.PutPosition(entity.Position);
			packet.PutInt(0);
			if (killer != null)
			{
				packet.PutInt(killer.Handle);
				packet.PutInt(0);
				packet.PutLong(0x40);
				packet.PutLong(0x35);
				packet.PutLong(0x40);
			}

			entity.Map.Broadcast(packet, entity);
		}
```

src/ChannelServer/Skills/Cleric/Heal.cs
```diff
@@ -0,0 +1,56 @@
...
+using Melia.Shared.Util;
+using Melia.Shared.World;
+
+namespace Melia.Channel.Skills.Cleric
+{
+	/// <summary>
+	/// Handler for the Archer skill Multishot.
+	/// </summary>
+	[SkillHandler(SkillId.Cleric_Heal)]
+	class Heal : IGroundSkillHandler
```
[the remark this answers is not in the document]
SalmanTKhan: Sounds good, damn these default class creations though.

src/ChannelServer/Skills/Cleric/Heal.cs
```diff
@@ -0,0 +1,56 @@
...
+			}
+
+			if (skill.SpendSp > 0)
+				caster.ModifySp(-skill.SpendSp);
+
+			skill.IncreaseOverheat();
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, targetPosition);
+
+			// The hitbox seems pretty small, there's presumably more going
```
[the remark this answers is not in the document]
SalmanTKhan: So splash range isn't the correct value here altogether, from what I figure, that value is the skill after cast "splashing" to the next monster. While we're treating it as the target range of the initial cast. For skills that are held to cast and draw a green range on the floor use the values from skill_selectground_drawrange.ies. For example, 30001 Archer Multishot has a draw range of 30. Not too sure for other skills.

src/ChannelServer/Network/Helpers/BuffHelper.cs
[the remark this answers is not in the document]
exectails: I see. Though the difference in `ZC_DEAD´ could also be in the killer, instead of the dead. I guess we'd need a PvP log to say for sure.

src/ChannelServer/Skills/Cleric/Heal.cs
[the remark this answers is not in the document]
exectails: You mean it would be easier to do a switch/case?^^

src/ChannelServer/Skills/Cleric/Heal.cs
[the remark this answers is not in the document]
exectails: I thought splash range might be a catch-all. It might mean different things for the different target types (though as I commented, there's surely some kind of calculation involved). For a `Circle` type it might mean the radius, while for a `Square` it might be the width/height, and for Thrust, which uses `Square`, I figured it might be the width of the imagined rectangle that is draw between source and target position, to designate the hit box. I think skill_selectground_drawrange has decidedly too few entries to apply to a lot of skills. And doesn't it sound like it's primarily for _drawing_? So the client know how to display the target area? Your interpretation of splash being solely for splash makes sense, but how is the target area defined then?

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: https://marketplace.visualstudio.com/items?itemName=mynkow.FormatdocumentonSave
Something like that?

exectails: Yea, that's the one I use. It runs the format document command every time you save a file. I find it pretty convenient, that way the code is always formatted in the way defined by the editorconfig.

src/ChannelServer/World/WorldManager.cs
SalmanTKhan: So similar to those skill "handles" or skill actor ids, they seem to be independent of entity handles.
The player/character handle is 379706 and the Buff "handle" was 331185, next buff "handle" I got was about a minute later was 331846, a minute later it was 332232. So it looks like it overlaps with regular entity handles.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Instead of level changing, I think over buff counter is the right way to go about it. Because buffs have different effects on stacks, maybe rename Over Buff Counter to Buff Stacks? So maybe level is useless in buffs.

src/ChannelServer/World/WorldManager.cs
exectails: Sound like it. Alright 👍

src/ChannelServer/World/Entities/Components/Buffs.cs
exectails: Do I have that right, that buffs can stack, and that that changes their effect?

SalmanTKhan: I'm basing it off buffs I've seen, thus far. 
https://treeofsavior.com/page/class/view.php?c=Scout
Oblique Fire - 5 stacks maximum: increases your movement speed when 3 stacks or lower and decreases the enemy's movement speed when 4 stacks or more.

exectails: Hm. Yea, that does sound like that might be how it works.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -55,12 +55,15 @@ public void AddSilent(Buff buff)
...
 		public void Add(Buff buff)
 		{
-			var buffExists = _buffs.ContainsKey(buff.Id);
-			this.AddSilent(buff);
-			if (buffExists)
-				Send.ZC_BUFF_UPDATE(this.Character, buff);
-			else
-				Send.ZC_BUFF_ADD(this.Character, buff);
+			var buffExists = this.Has(buff.Id);
+			lock (_buffs)
```
exectails, line 59: This lock is unnecessary, because you switched to using Has and AddSilent, which both handle the locking. You should also only lock for as long as necessary, meaning as long as you need to access the resource that is to be locked, because while you're sending packets, no other thread can access the locked area.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -121,27 +120,30 @@ public bool Remove(Buff buff)
...
 		{
-			if (!_buffs.ContainsKey(buffId)) {
-				var buff = new Buff(this.Character, buffId);
-				Add(buff);
-			} else
+			if (!this.Has(buffId))
 			{
-				Update(buffId);
+				var buff = new Buff(this.Owner, buffId);
+				Add(buff);
```
exectails, line 126: `this.` ^v

exectails:
It would be good if you could put more detail into your commit names or descriptions. Once this gets merged, there will be no reference to the PR this was part of anymore, it would just say "Pull Request Changes". Reviewing commits is also sometimes a little easier if you split them up. One commit with a list of changes gives you no information which changes related to which problem. Splitting them up also makes reverting commits much easier, since you can simply revert a specific set of changes.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: What's the preferred way to do it? Use the lock then access the resource and send the packet within the lock or lock it to get the reference and send packet outside of the lock?

exectails: Like I said, only lock for as long as absolutely necessary, to free up the resource for other threads. In this case, and if you weren't using Has and AddSilent, which handle it, you might do something like this:
```cs
public void Add(Buff buff)
{
	var updated = false;
	lock (_buffs)
	{
		updated = _buffs.ContainsKey(buff.Id);
		_buffs[buff.Id] = buff;
	}

	if(updated)
		Send.ZC_BUFF_UPDATE(this.Owner, buff);
	else
		Send.ZC_BUFF_ADD(this.Owner, buff);
}
```

src/ChannelServer/Skills/Cleric/Heal.cs
```diff
@@ -0,0 +1,51 @@
...
+			if (skill.SpendSp > 0)
+				caster.ModifySp(-skill.SpendSp);
+
+			skill.IncreaseOverheat();
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, targetPosition);
+			var healAmount = (int)(caster.GetRandomMagicAmplification() * skill.Data.SkillFactor / 100f);
+
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, targetPosition, null, 0);
+			// Heal uses ZC_HEAL_INFO instead of ZC_ADD_HP
```
exectails, line 46: Does this cause any issues? Or does `ZC_HEAL_INFO` maybe support any additional features? Should we have a separate method for it?

SalmanTKhan: Heal Info seems to be used where visual text is needed for healing, heal spell, recovery (both hp and sp) items. Passive hp/sp recovery is done via ADD_HP/UPDATE_SP.

exectails:
Are you planning to do more work on this, or should I merge it for now?

SalmanTKhan:
Updated Buff Implementation.

src/ChannelServer/Skills/General/SelfSkillHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Melia.Channel.Skills.General
+{
+	/// <summary>
+	/// Targeted ground skill implementation when the client provides targets
+	/// </summary>
+	public class SelfSkillHandler : ISelfSkillHandler
+	{
+		public void Handle(Skill skill, Character caster, Position castPosition, Direction direction)
+		{
+			if (skill.SpendSp > 0)
```
exectails, line 20: Couldn't you simplify this by just checking whether caster.Sp is lower than SpendSp? Because that will only be the case if SpendSp is greater than 0 and the caster doesn't have enough Sp.

src/ChannelServer/Skills/Skill.cs
```diff
@@ -171,7 +171,7 @@ public void IncreaseOverheat()
 			Send.ZC_OVERHEAT_CHANGED(this.Character, this);
 
 			// TODO: Apply overheat cooldown
-			if (this.IsOverheated)
+			if (this.IsOverheated || !this.CanOverheat)
```
exectails, line 174: Did you add this because the packets actually need to be sent, or because officials send them, even if the skill can't overheat?

src/ChannelServer/Skills/Swordsman/Bash.cs
```diff
@@ -0,0 +1,62 @@
...
+using Melia.Channel.World;
+using Melia.Channel.World.Entities;
+using Melia.Shared.Const;
+using Melia.Shared.Util;
+using Melia.Shared.World;
+
+namespace Melia.Channel.Skills.Swordsman
+{
+	/// <summary>
+	/// Handler for the Swordsman skill Thrust.
```
exectails, line 13: Is there much of a reason to create a handler for Bash when you don't even change the comments?^^ Remember, the skill handlers will most definitely not stay the way they are, and every single one we add will have to be updated later.

src/ChannelServer/Skills/Wizard/EarthQuake.cs
```diff
@@ -0,0 +1,45 @@
...
+			Send.ZC_NORMAL_Skill_1D(caster, caster);
+
+			var targets = caster.Map.GetAttackableEntitiesInRange(caster, targetPosition, skill.Data.SplashRange + skill.Data.SplashHeight);
+			var damage = (int)(caster.GetRandomPAtk() * skill.Data.SkillFactor / 100f);
+
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, targetPosition, targets, damage);
+
+			foreach (var target in targets)
+			{
+				target.TakeDamage(damage, caster, DamageVisibilityModifier.Invisible, 0);
```
exectails, line 41: Does this code compile? I removed half of those arguments from TakeDamage, it doesn't seem like you update these handlers yet, right?

src/ChannelServer/Skills/Wizard/EnergyBolt.cs
```diff
@@ -0,0 +1,56 @@
...
+				{
+					Log.Warning("Energy Bolt: Player {0} cast skill farther than max range ({1} > {2}).", caster.Name, castRange, skill.Data.MaxRange);
+					return;
+				}
+
+				skill.IncreaseOverheat();
+
+				Send.ZC_SKILL_READY(caster, skill, caster.Position, target.Position);
+
+				// The hitbox seems pretty small, there's presumably more going
```
exectails, line 39: I think you might not want to copy my comments into every skill handler, when we don't know whether they apply to those skills^^ Unless they do... do they?

src/ChannelServer/Skills/Wizard/EnergyBolt.cs
```diff
@@ -0,0 +1,56 @@
...
+				var radius = (int)skill.Data.SplashRange * 2;
+
+				var targets = caster.Map.GetAttackableEntitiesInRangeSortedByDistance(caster, target.Position, radius);
+				var damage = (int)(caster.GetRandomPAtk() * skill.Data.SkillFactor / 100f);
+
+				Send.ZC_NORMAL_Skill_1D(caster, target);
+				Send.ZC_SKILL_RANGE_CIRCLE(caster, target, skill);
+				Send.ZC_SKILL_FORCE_TARGET(caster, target, skill, damage);
+				target.TakeDamage(damage, caster, DamageVisibilityModifier.Invisible, 0);
+			} else
```
exectails, line 50: This else isn't formatted correctly, as per our coding conventions. Didn't you want to get an auto-formatter?^^

src/ChannelServer/Skills/Wizard/Lethargy.cs
```diff
@@ -0,0 +1,62 @@
...
+			var radius = (int)skill.Data.SplashRange * 2;
+
+			var targets = caster.Map.GetAttackableEntitiesInRectangle(caster, castPosition, targetPosition, radius);
+			var damage = (int)(caster.GetRandomPAtk() * skill.Data.SkillFactor / 100f);
+
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, targetPosition, null, damage);
+
+			foreach (var target in targets)
+			{
+				// TODO add buff when buff code is merged
```
exectails, line 56: Heh, I guess these actually were intended for a second PR? =P

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using Melia.Channel.World.Entities;
+using Melia.Shared.Const;
+using Melia.Shared.Data.Database;
+
+namespace Melia.Channel.World
+{
+	/// <summary>
+	/// Modeled after Skill class
```
exectails, line 9: This isn't a very useful description of what the class is and does, don't you think? If you hover over a `new Buff`, to see what that is, the text "this is like that other class" doesn't tell you much.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,123 @@
...
+		/// <summary>
+		/// Returns the buff's id.
+		/// </summary>
+		public BuffId Id { get; }
+
+		/// <summary>
+		/// Returns the buff's associated skill id.
+		/// Default value is 1.
+		/// </summary>
+		public int SkillId { get; }
```
exectails, line 32: Shouldn't this use `SkillId`?

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,123 @@
...
+		/// <summary>
+		/// The buff's removal time
+		/// </summary>
+		public DateTime RemovalTime { get; set; }
+
+		/// <summary>
+		/// If the buff has an update time
+		/// </summary>
+		/// <returns></returns>
+		public bool HasUpdateTime() => this.Data.UpdateTime != 0;
```
exectails, line 66: This is supposed to be a property, right? In that case you can remove the parantheses.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		/// <summary>
+		/// The buff's removal time
```
exectails, line 69: Copied summary. And that's unfortunate, because now I have to read the code to find out what NextUpdateTime does^^

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,123 @@
...
+			if (this.HasUpdateTime())
+				this.NextUpdateTime = DateTime.Now.AddMilliseconds(this.Data.UpdateTime);
+		}
+
+		/// <summary>
+		/// Get Next Update Time
+		/// </summary>
+		public void Update()
+		{
+			if (this.HasUpdateTime())
```
exectails, line 111: Do you think it would be cleaner to give control over the buff's updates to the buff itself? That way you wouldn't need this update time property, you wouldn't need to expose internal information, the buff could save a reference to its handler, for repeated calls, you could just forward the update to the buffs, etc.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,247 @@
...
+
+		/// <summary>
+		/// Adds given buff and updates the client.
+		/// </summary>
+		/// <param name="buff"></param>
+		public void Add(Buff buff)
+		{
+			buff.IncreaseOverbuff();
+			this.AddSilent(buff);
+			var handler = ChannelServer.Instance.BuffHandlers.GetBuff(buff.Id);
```
exectails, line 59: Shouldn't this happen in AddSilent as well? Otherwise the buff wouldn't get started there.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,247 @@
...
+		/// didn't exist. Updates the client on success.
+		/// </summary>
+		/// <param name="buffId"></param>
+		/// <returns></returns>
+		public bool Remove(BuffId buffId)
+		{
+			var buff = this.Get(buffId);
+			if (buff != null)
+			{
+				this.Remove(buff);
```
exectails, line 87: You should return true here, as that's the expected return value for when the buff _was_ removed.

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,247 @@
...
+				else
+					this.Update(buff.Id);
+			}
+		}
+
+		/// <summary>
+		/// Update a buff if it exists using the buff id.
+		/// </summary>
+		/// <param name="buffId"></param>
+		public void Update(BuffId buffId)
```
exectails, line 134: Wouldn't private be more appropriate for this method? Or is other code supposed to call this?

src/ChannelServer/World/Entities/Components/Buffs.cs
```diff
@@ -0,0 +1,247 @@
...
+		{
+			var buff = this.Get(buffId);
+			if (buff != null)
+			{
+				lock (_buffs)
+				{
+					buff.IncreaseOverbuff();
```
exectails, line 141: There's no reason to lock here.

src/ChannelServer/World/Entities/Entity.cs
```diff
@@ -52,6 +56,11 @@ public interface IEntity
 	/// </summary>
 	public interface ICombatEntity : IEntity
 	{
+		/// <summary>
```
exectails, line 59: We have the Components property to access components, so we don't need to put them into the interfaces like this.

src/ChannelServer/World/Map.cs
```diff
@@ -251,6 +251,112 @@ public List<ICombatEntity> GetAttackableEntitiesInRange(ICombatEntity attacker,
...
 
+		/// <summary>
+		/// Returns attackable monsters sorted by their distance in the given 
+		/// radius around position.
+		/// </summary>
+		/// <param name="attacker"></param>
+		/// <param name="position"></param>
+		/// <param name="radius"></param>
+		/// <returns></returns>
+		public List<ICombatEntity> GetAttackableEntitiesInRangeSortedByDistance(Character attacker, Position position, float radius)
```
exectails, line 262: The result wouldn't be sorted correctly if an entity is able to attack both monsters _and_ characters, would it? Because you're first adding the monsters sorted by range and then the characters.

src/ChannelServer/World/Map.cs
```diff
@@ -251,6 +251,112 @@ public List<ICombatEntity> GetAttackableEntitiesInRange(ICombatEntity attacker,
...
+
+		/// <summary>
+		/// Returns attackable monsters sorted by their distance in the given 
+		/// radius around position.
+		/// </summary>
+		/// <param name="attacker"></param>
+		/// <param name="position"></param>
+		/// <param name="radius"></param>
+		/// <returns></returns>
+		public ICombatEntity GetNearestAttackableEntity(Character attacker, Position position, float radius)
```
exectails, line 317: And here you wouldn't get the actual nearest entity at all if you have both monsters and characters as options.

src/ChannelServer/World/Map.cs
```diff
@@ -251,6 +251,112 @@ public List<ICombatEntity> GetAttackableEntitiesInRange(ICombatEntity attacker,
...
+			return result;
+		}
+
+		/// <summary>
+		/// Returns nearest attackable entity in the given radius around position.
+		/// </summary>
+		/// <param name="position"></param>
+		/// <param name="radius"></param>
+		/// <returns></returns>
+		public List<ICombatEntity> GetAttackableEntitiesInRangeSortedByDistance(ICombatEntity attacker, Position position, float radius)
```
exectails, line 342: Isn't this the same as the other `GetAttackableEntitiesInRangeSortedByDistance`?

src/Shared/Network/SubOp.cs
```diff
@@ -1,27 +1,27 @@
...
+    public static class SubOp
+    {
+        /// <summary>
+        /// Sub opcodes to be used with the BC_NORMAL packet.
+        /// </summary>
+        public static class Barrack
+        {
+            public const int SetBarrackCharacter = 0x00;
+            public const int SetPosition = 0x02;
+            public const int SetBarrack = 0x05;
```
exectails, line 12: Please use tabs, as per our coding conventions.

src/Shared/World/Position.cs
```diff
@@ -75,17 +76,30 @@ public double Get3DDistance(Position otherPos)
...
 		{
 			return (Math.Pow(X - otherPos.X, 2) + Math.Pow(Z - otherPos.Z, 2) <= Math.Pow(range, 2));
 		}
 
+		/// <summary>
+		/// Returns true if other position is within given range in 2D space.
+		/// </summary>
+		/// <param name="otherPos"></param>
+		/// <returns></returns>
+		public bool InRange2DDebug(Position otherPos, float range)
```
exectails, line 89: What is the purpose of this method?

exectails (commented):
The changes to the skill handlers would've made a good second PR^^

src/ChannelServer/World/Map.cs
SalmanTKhan: These methods aren't really supposed to be a part of the buff PR, they are a part of the skills branch, but the concern is correct regarding behavior, there isn't a superset of entities, so I presume I either add them to the list and then sort it.

src/ChannelServer/World/Map.cs
SalmanTKhan: Similar to the comment above, these are skill-related methods I made, to emulate the behavior I saw when skills are cast on the server, but like you said I add the monsters first and sort them and then characters and sort them. Making the code prioritize monsters first. So it was a rough implementation since the state of Melia is PVE right now.

src/ChannelServer/World/Map.cs
SalmanTKhan: That sounds about right, no need for the other one since this one accepts ICombatEntity as the attacker.

src/Shared/World/Position.cs
SalmanTKhan: That was supposed to be removed lol.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -337,14 +338,99 @@ public static void ZC_SKILL_FORCE_TARGET(Character character, IEntity target, Sk
...
 		/// <remarks>
 		/// i339415, looks hit info is being used instead of this for showing damage on a target
 		/// </remarks>
 		/// <param name="character"></param>
-		/// <param name="skillId"></param>
+		/// <param name="skill"></param>
+		/// <param name="targetPosition"></param>
+		/// <param name="targets"></param>
+		/// <param name="damage"></param>
+		public static void ZC_SKILL_MELEE_GROUND(Character character, Skill skill, Position targetPosition, IEntity target, int hitCount, int damage = 0)
```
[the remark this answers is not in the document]
SalmanTKhan: Removed skill-related changes, they aren't supposed to be a part of this PR.

src/ChannelServer/World/Map.cs
SalmanTKhan: Removed skill-related changes, they aren't supposed to be a part of this PR.

src/ChannelServer/World/Map.cs
SalmanTKhan: Removed skill-related changes, they aren't supposed to be a part of this PR.

exectails: >it was a rough implementation since the state of Melia is PVE right now

Yea, that's fine. Though I would suggest to generally keep the future in mind, and to implement things in a way that (hopefully) won't require much changes in the future. That way you save yourself the refactoring, or the potential confusion, when you or someone else calls this method, expecting certain results, and getting something else. In this particular case, it wouldn't even have required much more code after all.

src/ChannelServer/Buffs/BuffHandlers.cs
```diff
@@ -0,0 +1,70 @@
...
+using Melia.Shared.Const;
+
+namespace Melia.Channel.Buffs
+{
+	/// <summary>
+	/// Buff handler collection and manager.
+	/// </summary>
+	public class BuffHandlers
+	{
+		private IBuffHandler _defaultBuffHandler = new BuffHandler();
```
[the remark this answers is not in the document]
SalmanTKhan: Removed it and now GetBuff returns null if no buff handler exists.

src/ChannelServer/World/Entities/Entity.cs
SalmanTKhan: Removed and now accessing it through components.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Removed excess lock.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: I've moved the buff handlers to AddSilent/RemoveSilent since those are called within the counterpart functions which send the client update packets. Cuts down on the redundancy in my opinion.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan:
```
		 public bool Remove(BuffId buffId)
		{
			var buff = this.Get(buffId);
			if (buff != null)
				return this.Remove(buff);
			return false;
		}
```

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Both Add and Update define behavior for buffs, you are correct we can just set them to private and allow access to the AddOrUpdate function instead.

src/ChannelServer/World/Entities/Components/Buff.cs
SalmanTKhan: I've made the changes to move the handler and update calls to the buff itself with a reference to the handler, let me know what you think.

src/ChannelServer/Network/Helpers/BuffHelper.cs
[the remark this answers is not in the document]
SalmanTKhan: IEntity now has a Name property.

src/ChannelServer/World/Entities/Components/Buff.cs
SalmanTKhan: Dead... will do boss man, added a better summary xD.

src/ChannelServer/World/Entities/Components/Buff.cs
SalmanTKhan: Refactored handlers to be called from Buff class.

src/ChannelServer/Skills/Wizard/Lethargy.cs
SalmanTKhan: These aren't supposed to be in this PR, it's a part of the Skill branch I have.

src/ChannelServer/Skills/Wizard/EnergyBolt.cs
SalmanTKhan: Auto formatter failing me :(.

src/ChannelServer/Skills/Skill.cs
SalmanTKhan: So if you don't add that, the non-overheat skills don't go on cooldown. You didn't notice that? The official server still sends both packets even if a server can't overheat.

src/ChannelServer/Skills/Swordsman/Bash.cs
SalmanTKhan: Pretty sure that was a work in progress, I've tried to do a better job of updating comments. I had two classes left on the Skills branch, for the basic classes, which were Scout and Swordsman.

src/ChannelServer/Skills/Wizard/EarthQuake.cs
SalmanTKhan: I removed the skill handlers that weren't a part of the PR.

src/ChannelServer/World/Entities/Components/Buffs.cs
exectails: Yea, that will work.

src/ChannelServer/Network/Helpers/BuffHelper.cs
```diff
@@ -25,8 +25,8 @@ public static void AddBuff(this Packet packet, Buff buff)
 			packet.PutInt(buff.Duration);
 			packet.PutInt(0);
 			packet.PutInt(0);
-			packet.PutShort((buff.Caster as Character)?.Name.Length + 5 ?? 5);
-			packet.PutString((buff.Caster as Character)?.Name ?? "");
+			packet.PutShort(buff.Caster.Name.Length + 5);
```
exectails, line 28: For future reference, if there's unusual structures like this, where we ourselves aren't even entirely sure why they are the way they are, you should describe what's going on in a short comment. Someone else will look at this code and be really confused what this `+ 5` is about.

src/ChannelServer/Skills/Skill.cs
exectails: Oh, makes sense. I haven't looked too much into skills yet. A comment would help here^^

src/ChannelServer/Network/Helpers/BuffHelper.cs
SalmanTKhan: Sounds good 👍

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -4105,14 +4105,105 @@ public static void ZC_UPDATED_MONSTERAPPEARANCE(Monster monster)
...
+			entity.Map.Broadcast(packet);
+		}
+
+		/// <summary>
+		/// Buff list, sent when a entity spawns?
+		/// </summary>
+		/// <param name="entity"></param>
+		public static void ZC_BUFF_LIST(IEntity entity)
+		{
+			var packet = new Packet(Op.ZC_BUFF_REMOVE);
```
exectails, line 4157: Is `ZC_BUFF_REMOVE` right? The method is named `ZC_BUFF_LIST`. Same for `ZC_BUFF_CLEAR`.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,142 @@
...
+
+		/// <summary>
+		/// Creates a new instance.
+		/// </summary>
+		/// <param name="caster"></param>
+		/// <param name="target"></param>
+		/// <param name="buffId"></param>
+		/// <param name="duration"></param>
+		/// <param name="skillId"></param>
+		public Buff(ICombatEntity caster, ICombatEntity target, BuffId buffId, int duration = 0, SkillId skillId = SkillId.Normal_Attack)
```
exectails, line 87: What do you think about using `TimeSpan` instead of `int` for the duration? It's a little more to type, but it makes it much clearer what you're working with. A simple integer, especially if you didn't document anything about it, could be anything. Ticks, seconds, minutes, you'd have to guess or look up other uses of the method.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,142 @@
...
+			this.Caster = caster;
+			this.Target = target;
+			this.Id = buffId;
+			this.SkillId = skillId;
+			this.Handle = ChannelServer.Instance.World.CreateBuffHandle();
+			this.Data = ChannelServer.Instance.Data.BuffDb.Find(buffId) ?? throw new ArgumentException($"Unknown buff '{buffId}'.");
+			this.Handler = ChannelServer.Instance.BuffHandlers.GetBuff(buffId);
+			if (duration == 0)
+			{
+				if (this.Data.Duration != 0)
```
exectails, line 98: If someone were to pass duration 0 for whatever reason, it would always default to the data duration, effectively overwriting the 0. Is this the intended behavior? Either way, I believe this should be documented in the doc comments, because without looking at this code, you wouldn't know what duration 0 will do. Personally I would expect it to be infinite, someone else might expect it to run for one tick, neither might expect it to fall back on the data, etc.

SalmanTKhan: So my initial assumption was duration in the database was accurate, but then some values didn't match up, hence I added a local duration to the buff when we could override the database defaults. My proposed changed is use -1 for database default while 0 is infinite and any positive integer is a duration in milliseconds.

src/ChannelServer/World/Entities/Components/Buff.cs
SalmanTKhan: I don't mind it, but I haven't used TimeSpan before, even after looking up the documents the initializer can use ticks or hours, minutes, seconds. So I'm not too sure how about to sure how to implement this as I'm using milliseconds right now. I've renamed the parameter to be more verbose durationInMilliseconds.

src/ChannelServer/Network/ChannelPacketSender.cs
SalmanTKhan: I've fixed this implementation. ZC_BUFF_LIST is usually sent on entity entering the map.
ZC_BUFF_CLEAR is sent on entity death usually, maybe when it leaves the map too?

src/ChannelServer/Skills/Wizard/EnergyBolt.cs
SalmanTKhan: Removed not a part of this PR.

src/ChannelServer/Skills/Skill.cs
SalmanTKhan: Removed, not a part of this PR.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+	/// <summary>
+	/// Modeled after Skill class
+	/// </summary>
+	public class Buff
+	{
+		/// <summary>
+		/// The buff's owner.
+		/// </summary>
+		public Character Character { get; }
```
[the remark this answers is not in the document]
SalmanTKhan: You're correct, I should really mark these as resolved after I change them.

src/ChannelServer/Skills/Cleric/Heal.cs
```diff
@@ -0,0 +1,56 @@
...
+
+			skill.IncreaseOverheat();
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, targetPosition);
+
+			// The hitbox seems pretty small, there's presumably more going
+			// into this. Double the splash range for the width for now.
+			var radius = (int)skill.Data.SplashRange * 2;
+
+			var targets = caster.Map.GetAttackableEntitiesInRectangle(caster, castPosition, targetPosition, radius);
```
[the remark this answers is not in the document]
SalmanTKhan: Removed not really the scope of this PR.

src/ChannelServer/Skills/Cleric/Heal.cs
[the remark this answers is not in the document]
SalmanTKhan: Not a part of this PR, marking as resolved for now.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: It should.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Will keep it in mind for accessing collections.

src/ChannelServer/World/Entities/Components/Buff.cs
SalmanTKhan: Sounds good.

src/ChannelServer/World/Entities/Components/Buff.cs
```diff
@@ -0,0 +1,70 @@
...
+			if (this.Data.Duration != 0)
+				this.RemovalTime = DateTime.Now.AddMilliseconds(Math.Max(0, this.Data.Duration));
+		}
+
+		/// <summary>
+		/// Increase overbuff counter
+		/// </summary>
+		public void IncreaseOverbuff()
+		{
+			OverbuffCounter++;
```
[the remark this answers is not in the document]
SalmanTKhan: Will keep this in mind and fixed.

src/ChannelServer/World/Entities/Components/Buff.cs
exectails: For the parameter you would just use the type `TimeSpan` and when creating a buff you would, for example, use `TimeSpan.FromMilliseconds(1000)` to create it. For a time I disliked this, because of how verbose it is, but the clarity is unbeatable, which comes in handy in bigger projects that various people work on, because there's always cases where seconds or something might make more sense than milliseconds, and suddenly you have different types of durations in different parts of the code. With TimeSpan und you can just use different types when it suits you, like `TimeSpan.FromMinutes(1)`.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -4106,13 +4106,113 @@ public static void ZC_UPDATED_MONSTERAPPEARANCE(Monster monster)
...
+		/// </summary>
+		/// <param name="entity"></param>
+		public static void ZC_BUFF_LIST(IEntity entity)
+		{
+			var buffs = entity.Components.Get<BuffCollection>();
+			var buffCount = buffs?.Count ?? 0;
+			var packet = new Packet(Op.ZC_BUFF_LIST);
+			packet.PutInt(entity.Handle);
+			packet.PutByte((byte)buffCount);
+			if (buffCount > 0)
```
exectails, line 4162: Please use curly braces whenever there's more than one line in the if's scope.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -4106,13 +4106,113 @@ public static void ZC_UPDATED_MONSTERAPPEARANCE(Monster monster)
...
+			entity.Map.Broadcast(packet);
+		}
+
+		/// <summary>
+		/// Clear buff buff handle association? Sent on entity death and removal
+		/// </summary>
+		/// <param name="entity"></param>
+		public static void ZC_BUFF_CLEAR(IEntity entity)
+		{
+			var packet = new Packet(Op.ZC_BUFF_REMOVE);
```
exectails, line 4175: CLEAR vs REMOVE?

src/ChannelServer/Network/Helpers/BuffHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+		/// Adds buff data to packet.
+		/// </summary>
+		/// <remark>
+		/// This version is used in ZC_BUFF_LIST
+		/// </remarks>
+		/// <param name="packet"></param>
+		/// <param name="buff"></param>
+		public static void AddBuffData(this Packet packet, Buff buff)
+		{
+			packet.PutInt((int)buff.Id);
```
exectails, line 50: Seems like this is part of the buff structure, couldn't you use it from AddBuff then? The name AddBuffData could potentially be misunderstood though, since Data is our suffix for the data entry classes.

src/ChannelServer/Network/ChannelPacketSender.cs
SalmanTKhan: Oh, you're correct lol, that should be clear.

src/ChannelServer/Network/ChannelPacketSender.cs
SalmanTKhan: Lol, I had the curly braces but removed them.

src/ChannelServer/Network/Helpers/BuffHelper.cs
SalmanTKhan: I've renamed AddBuff to AddTargetedBuff and renamed AddBuffData to AddBuff. AddTargetedBuff uses AddBuff in it.

src/ChannelServer/Skills/General/SelfSkillHandler.cs
SalmanTKhan: That's a great idea, but not for this PR.

src/ChannelServer/Network/Helpers/BuffHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+			packet.PutInt(0);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			packet.PutInt(buff.OverbuffCounter);
+			packet.PutInt(buff.Duration);
+			packet.PutInt(0);
+			packet.PutInt(0);
+			if (entity is Character character)
+			{
+				packet.PutShort(character.Name.Length + 1 + 4);
```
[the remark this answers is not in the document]
SalmanTKhan: We've talked about what this was and I've commented in the code on it so it's apparent why it's done in such a way.

src/ChannelServer/Network/ChannelPacketSender.cs
```diff
@@ -2740,21 +2847,17 @@ public static void ZC_NORMAL_UpdateSkillUI(Character character)
...
-			//packet.PutInt(character.Handle);
-			packet.PutInt(entity?.Handle ?? 0);
-			//packet.PutShort(skill.Id);
+			packet.PutInt(entity.Handle);
 			packet.PutShort(0);
-			packet.PutFloat(position.X);
-			packet.PutFloat(position.Y);
-			packet.PutFloat(position.Z);
-			packet.PutFloat(70f);
+			packet.PutFloat(character.Position.X);
```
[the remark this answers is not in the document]
SalmanTKhan: I've reverted this change, wasn't the scope of this PR.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Added an existence check.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: I've made the changes to check for existence, thanks.

src/ChannelServer/World/Entities/Components/Buffs.cs
SalmanTKhan: Mental notes making into code comments ^^.

PR #192, opened 2021-09-20
Character slot must have a default value

vegax87:
otherwise it causes this error on LoginServer:

[Exception] - MySql.Data.MySqlClient.MySqlException (0x80004005): Field 'slot' doesn't have a default value


exectails:
Whoops. Thank you for the pull request. Though modifying update SQL files after they were already added to the repo is suboptimal, because they might've already been executed for a server, in which case the bug would persist. Instead, a new update file should be created. Though I think a more appropriate way to fix this issue, is to actually set the slot on creation, because new characters are added the end of the list. I have just pushed this change. Thank you for making me aware of this issue.

PR #193, opened 2021-09-21
String Properties Update

SalmanTKhan:
* Updated string property cases in common.bt
* Updated string property cases in Properties.cs


exectails:
👍

PR #200, opened 2022-03-31
Web Registration Password Fix

### LINKED ISSUE #199: Web Registration Script doesn't work
If you try to register via the Web registration script, it fails due to the password being passed as a plain text password that's being encrypted, instead of an MD5 hash to be encrypted.

SalmanTKhan:
* ToS Client requires an MD5 password, so added a hashing function to the script, prior to submitting an account creation request to the Database, we create an MD5 hash of the password and then send that to the Database's CreateAccount function.
* Closes #199

### DIFF
--- a/system/web-optional/register/index.cs
+++ b/system/web-optional/register/index.cs
@@ -6,6 +6,8 @@
 //---------------------------------------------------------------------------
 
 using System.IO;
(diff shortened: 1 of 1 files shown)


exectails:
👍

PR #214, opened 2023-05-28
Add a basic social server

SalmanTKhan:


src/SocialServer/Database/Character.cs
```diff
@@ -0,0 +1,186 @@
...
+		/// <summary>
+		/// Gets or sets the character's current stamina.
+		/// </summary>
+		public int Stamina { get; set; }
+
+		/// <summary>
+		/// Gets or sets the amount of STR the character receives from
+		/// their job.
+		/// </summary>
+		public int StrByJob { get; set; }
```
exectails, line 144: Does Social need all this? Iirc even Barracks doesn't anymore.

src/SocialServer/Database/ChatGroup.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Melia.Social.Database
+{
+	/// <summary>
+	/// Represents a chat group.
```
exectails, line 10: What does that mean though?^^

src/SocialServer/Database/ChatGroup.cs
```diff
@@ -0,0 +1,44 @@
...
+
+		/// <summary>
+		/// Chat receiver's name
+		/// </summary>
+		public string ReceiverName { get; set; }
+
+		/// <summary>
+		/// A list of messages sent in the chat.
+		/// </summary>
+		public List<string> Messages { get; } = new List<string>();
```
exectails, line 42: I'm generally not a big fan of exposing raw lists or dictionaries (chat groups in Account), as it necessitates knowledge of the class' inner workings to use it and makes using it more complicated. I believe it would be better to implement appropriate methods to work with the group's data.

src/SocialServer/Database/SocialDb.cs
```diff
@@ -0,0 +1,68 @@
...
+					if (!reader.Read())
+						return null;
+
+					var account = new Account();
+					account.Id = reader.GetInt64("accountId");
+					account.Name = reader.GetStringSafe("name");
+					account.TeamName = reader.GetStringSafe("teamName");
+					account.Password = reader.GetStringSafe("password");
+
+					// Upgrade MD5 hashes
```
exectails, line 32: This shouldn't be necessary any longer, as we only used it to upgrade accounts that were created before we added BCrypt. Further, we already have it in Barracks, we should always be the first server the client connects to.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+				return;
+			}
+
+			// Logged in
+			conn.Account = account;
+			conn.LoggedIn = true;
+
+			Log.Info("User '{0}' logged in.", conn.Account.Name);
+
+			Send.SC_NORMAL_Unknown_00(conn);
```
exectails, line 52: You could create a dedicated file for SC_NORMAL now, so we don't have to change it later.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CS_CHAT)]
+		public void CS_CHAT(ISocialConnection conn, Packet packet)
+		{
+			var size = packet.GetShort();
+			var chat = packet.GetString();
+			var account = conn.Account;
+
+			Log.Debug("CS_CHAT: {0}", chat);
```
exectails, line 69: Random debug output should ideally not make it into master.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+			// TODO: Save chat and send it after saving it?
+			// TODO: Decide to use ChatCommands?
+			var splitChat = chat.Split(' ');
+			if (splitChat.Length < 3)
```
exectails, line 73: This code could use a little more spacing/grouping to improve readability.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+		[PacketHandler(Op.CS_CHAT)]
+		public void CS_CHAT(ISocialConnection conn, Packet packet)
+		{
+			var size = packet.GetShort();
+			var chat = packet.GetString();
+			var account = conn.Account;
+
+			Log.Debug("CS_CHAT: {0}", chat);
+			// TODO: Save chat and send it after saving it?
+			// TODO: Decide to use ChatCommands?
```
exectails, line 71: `Yggdrasil.Util.Commands.Arguments` might be useful here.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+			var splitChat = chat.Split(' ');
```
exectails, line 72: If you split strings, make sure to remove empty entries, or your code might break just by someone putting multiple spaces in their message.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+			if (splitChat.Length < 3)
+				return;
+			var command = splitChat[0];
+			var targetTeamName = splitChat[1];
+			var text = string.Join(" ", splitChat.Skip(2));
+			if (!SocialServer.Instance.Database.TeamNameExists(targetTeamName))
+			{
+				Send.SC_NORMAL_SystemMessage(conn, 101080, 1, 0);
```
exectails, line 80: `101080` is the client message `TargetUserNotExist`. We should probably add a client message db, maybe an enum, but for the moment this should at least get a comment about what this magic number is about.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+				return;
+			}
+			var target = SocialServer.Instance.Database.GetAccountByTeamName(targetTeamName);
```
exectails, line 83: You aren't checking target for null, what if the account doesn't exist?

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+			var chatGroup = conn.Account.ChatGroups.Values.FirstOrDefault(a => a.ReceiverId == target.Id);
+			if (chatGroup == null)
+			{
+				chatGroup = new ChatGroup()
+				{
+					ChatId = SocialServer.Instance.GetNewChatId(),
```
exectails, line 89: The chat group could do all this in a constructor with just two arguments, cleaning up this code and improving readability and usability.

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+					SenderId = account.Id,
+					SenderName = account.TeamName,
+					ReceiverId = target.Id,
+					ReceiverName = target.TeamName,
+				};
+				conn.Account.ChatGroups.Add(chatGroup.ChatId, chatGroup);
+			}
+			chatGroup.Messages.Add(text);
```
exectails, line 97: I don't know anything about the chat, but are you saving only the raw message text here? Don't you need to save who said it? Maybe also when?

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+		/// </summary>
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CS_CREATE_GROUP_CHAT)]
+		public void CS_CREATE_GROUP_CHAT(ISocialConnection conn, Packet packet)
+		{
+			var account = conn.Account;
+
+			var chatId = SocialServer.Instance.GetNewChatId();
+			Send.SC_NORMAL_CreateChatRoom(conn, account, chatId, "!@#$NewRoomHasBeenCreated#@!");
```
exectails, line 113: Is there a reason to pass both connection and account when you can get the account from the connection?

src/SocialServer/Network/PacketHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+		/// <summary>
+		/// Invite someone to a group chat
+		/// Dummy Handler
+		/// </summary>
+		/// <param name="conn"></param>
+		/// <param name="packet"></param>
+		[PacketHandler(Op.CS_GROUP_CHAT_INVITE)]
+		public void CS_GROUP_CHAT_INVITE(ISocialConnection conn, Packet packet)
+		{
+
```
exectails, line 127: Even dummy handlers should do _something_, unless they're literally expected to do nothing, in which case a comment should state as much. If they simply do nothing, the intention behind the code is unclear, and users might wonder why nothing is happening.

src/SocialServer/Network/Send.cs
```diff
@@ -0,0 +1,225 @@
...
+				case 0x14:
+					packet.PutLong(conn.Account.Id);
+					packet.PutInt(0); // 728
+					break;
+			}
+			conn.Send(packet);
+		}
+
+		/// <summary>
+		/// Login related
```
exectails, line 46: More elaborate comments might help us all with understanding and categorizing code in the future^^ "Login related" is a start, but is it about? When is it sent? What does it do? What happens if it's not sent?

src/SocialServer/Network/Send.cs
```diff
@@ -0,0 +1,225 @@
...
+			packet.PutInt(4723);
+			packet.PutInt(1);
+			packet.PutInt(0);
+			packet.PutByte(0);
+
+			conn.Send(packet);
+		}
+
+		/// <summary>
+		/// Dummy
```
exectails, line 87: Is it a dummy? To me that would indicate that it doesn't do anything, but that's clearly not the case. Could you elaborate on what these are?

src/SocialServer/Network/Send.cs
```diff
@@ -0,0 +1,225 @@
...
+		/// </summary>
+		/// <param name="conn"></param>
+		/// <param name="account"></param>
+		/// <param name="message"></param>
+		public static void SC_NORMAL_CreateChatRoom(ISocialConnection conn, Account account, long chatId, string message = "!@#$NewRoomHasBeenCreated#@!")
```
exectails, line 134: Another case of "we should probably use client messages"^^ Does it send the raw message if you don't wrap it in the lookup code or why did you decide to make the called include it?

src/SocialServer/SocialServer.cs
```diff
@@ -0,0 +1,106 @@
...
+		/// </summary>
+		public PacketHandler PacketHandler { get; } = new PacketHandler();
+
+		/// <summary>
+		/// Returns reference to the server's database interface.
+		/// </summary>
+		public SocialDb Database { get; } = new SocialDb();
+
+		private static long ChatId = 0x1FB0F00000000;
+		public long GetNewChatId()
```
exectails, line 35: Doesn't it feel like a proper chat manager might be cleaner? This method could also use a comment about what exactly it does and returns.

start-all.bat
```diff
@@ -46,5 +46,5 @@ exit
 :NO_ARGS
 start start-barracks
 start start-zone-1
-start start-zone-2
+start start-social
```
exectails, line 49: Why did you replace Zone 2?

exectails (changes requested):
Thanks for the pull request. I found a few minor issues that should ideally be addressed before it's merged.

src/BarracksServer/Network/PacketHandler.cs
```diff
@@ -108,9 +108,13 @@ public void CB_START_BARRACK(IBarracksConnection conn, Packet packet)
 		{
 			var unkByte = packet.GetByte();
 			var str1 = packet.GetString(64); // [i373230 (2023-05-10)] Might've been added before, same as CB_LOGIN
+			var socialServers = BarracksServer.Instance.ServerList.GetSocialServers();
```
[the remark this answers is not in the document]
SalmanTKhan: Would you like me to keep the unrelated var separate from the reading packet values?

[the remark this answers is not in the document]
exectails: Yea, otherwise your eyes might parse it all as one piece of code that looks like packet reading at a glance, when one line is something wholly different in reality. Proper grouping makes it easier to read.

src/BarracksServer/Network/PacketHandler.cs
```diff
@@ -108,9 +108,13 @@ public void CB_START_BARRACK(IBarracksConnection conn, Packet packet)
...
 
 			Send.BC_IES_MODIFY_LIST(conn);
-			Send.BC_SERVER_ENTRY(conn, "127.0.0.1", 9001, "127.0.0.1", 9002);
+			if (socialServers.Length >= 2)
```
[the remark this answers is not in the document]
SalmanTKhan: Initially, I thought it was a redundancy server, but you had mentioned that it wasn't, so the second server (9002) received CS_ADD_RELATION_SCORE, while the first (9001) dealt with chat packets. In the picture sent a whisper, the top server is the "chat" server and the bottom is the "relation" server.
image

src/SocialServer/Database/ChatGroup.cs
SalmanTKhan: image
Whatever these are xD?

src/BarracksServer/Network/PacketHandler.cs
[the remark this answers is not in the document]
exectails: Confirmation, nice =) The current design of the social server handling both or either is pretty good, but we should make a note of this somewhere.

src/SocialServer/Database/ChatGroup.cs
exectails: So... chat rooms between two players? Potentially more?

src/SocialServer/Network/Send.cs
exectails: Actually, this opens up the question of whether we should even use client messages at all... We decided we wanted to handle the localization on the server-side, and client messages are the opposite. We're sending a string key here that gets turned into a localized message.

SalmanTKhan: So it's just my coding habit, when I have different parts of a packet, I use them as defaults, so when I test the packet, I change part at a time without having to rewrite all the parameters. I have changed it in newer commits.

src/SocialServer/Network/Send.cs
SalmanTKhan: Excellent question, I usually don't test what happens if I don't send it but I'll take a look, to see what the behavior is.

PR #215, opened 2023-05-28
Patch notes up to Player Rank 10

KytheLXI:
Notable Patch notes up to Player Rank 10. Tried to limit enumeration to features, maps, classes, some progression (no numbers).


exectails:
Thanks for the pull request, a nice start. Could you move the file to the docs folder though? Since it's not directly related to Melia it might cause confusion when placed in the root folder as "patch notes".

PR #224, opened 2023-06-23
Added mob spawns

kenedos:
- Added fully functional mob spawns
- Mob spawns with usage of monster properties
- Added monster population class
- Changed mob spawner behaviour
- Changed "test" package name to "default"


exectails:
Thank you for your contribution! I looked your PR over, but before we go more in depth, I have a few questions and concerns.

First up, the number of scripts. You're creating a script for every map, which is fine, but you're creating them for _every_ map, regardless of whether monsters spawn there or not. Since every script file and class adds a little bit more overhead, we shouldn't create a bunch of essentially empty scripts that the server has to load, compile, and initialize.

Next is your monster population class. Could you tell me more about what you are trying to accomplish with it? Honestly, it feels a little convoluted and more difficult to control than simply setting min and/or max values. I believe that's a straight forward system that's easy to use and understand by users, while being flexible enough to support different types of spawns. For example, your current code doesn't allow a user to specify a spawner with a static amount of 10, nor one as ridiculous as 1~100. Additionally, your population calls relying on strings for identification adds another point of failure that didn't exist before.

There are a couple more things, though we should talk about these points first. Though I'd ask you to also take a look at our [coding conventions](https://github.com/NoCode-NoLife/melia/blob/master/CONTRIBUTING.md#coding-conventions), as I noticed a few things that aren't quite in line with it, such as not utilizing the `var` keyword and creating methods instead of properties for simple getters.

kenedos:
Sure, I can re-generate the data without the empty scripts. I wasn't sure whether to keep them as placeholders or remove them altogether.

The monster population class is indeed a bit more convoluted solution (I agree with you), but it seems to better emulate the data mined from the client along with official behaviour. If you look at the anchor and gentype files for f_siauliai_west, for example, you'll notice there are 78 anchors for rootcrystal_01 class, but all of them point to the same gentype 30 that has a max population of 12. If you take a look at official behaviour, you'll notice there aren't 78 * 12 root crystals around the map but rather only 12 in total (in the entire map). This also happens for normal monsters and seems to be the standard for most of client data: multiple anchor points in a single room sharing the same gentype and max population.

This means each gentype defines a max population which is shared among all anchors that use it. In our implementation, each anchor is analogue to the monster spawner, but we did not have an entity to emulate the gentype's max population. In essence, if we want to use client data, we need a monster population count that is shared among multiple respawners. That, or create a complicated clustering algorithm that would convert multiple anchor points into a single respawner.

As for the coding conventions, I will take a look and update my PR. Please feel free to ask any further questions meanwhile.

exectails:
Thank you for clarifying. I hadn't taken that close a look at the client data yet and this actually explains a lot. You're right that we couldn't easily mimic the behavior of the R1 server's spawn system with our current spawners. Though truth be told, I wonder if we should actually replace our spawners with this. Me going "Huh...?" over this, despite having the code in front of me should tell us something about how straight forward it is to use^^ And while some people will be happy with the default spawns, many will also want to extend or even revamp them entirely, in which case this system seems a little inflexible.

How do you feel about making this a separate system that is allowed to very closely mimic the client data, while leaving the original spawners as a more customizable solution? We could perhaps create "spawn points" instead of anchors, and "generators" instead of spawners, setting up areas the monsters can spawn in and then spawners that use that data.

```cs
AddSpawnPoint("Onions", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));
AddSpawnPoint("Onions", "f_siauliai_west", Spot(-575, -1300));

AddGenerator("Onions", MonsterId.Onion, 10);
```

If we created a spawner parent class or interface we could also put either version to use, as well as allow users to write new types of spawners easily.

kenedos:
Feels like an overcomplication to have a base class and two child classes that implement spawn points differently. I'm all in for more flexibility but I also don't want to make things more complicated. Can I instead propose creating a new shortcut that will allow Spawners to be called with just monster class and quantity which will internally create a MonsterPopulation/Generator of its own?

For example:
- Users can create a spawn point with either:
`AddGenerator("f_siauliai_west", "Onions", MonsterId.Onion, 10);
AddSpawnPoint("Onions", TimeSpan.FromSeconds(0), "f_siauliai_west", Spot(-575, -1300));`
OR
`AddSpawnPoint(MonsterId.Onion, 10, TimeSpan.FromSeconds(0), "f_siauliai_west", Spot(-575, -1300));`
- In first example, we create the generator and then add it to the spawn point, making usage of MonsterPopulation/Generator directly.
- In second example, the MonsterPopulation/Generator would be created internally only for that SpawnPoint and set its max population to 10.

Also notice I still kept the map reference in "AddGenerator" on first example. If we follow this proposed structure, I believe we cannot get rid of the map parameter in the generator call. If the user happens to create two generators with same name on two different files the SpawnPoint wouldn't know which one to use. Also, it would be very weird to use a same generator on two different maps (kill mob in one map and it spawns in another?), not to mention, a lot of bugs could arise if we have things like dynamic mob generation or something similar.

For identifying the MonsterPopulation/Generator, I would propose we use either **both** the name string and the map, **OR** we change the structure entirely to make the "AddGenerator" call instead return a Generator object. With the latter solution, we could have things like:
`var OnionGenerator = AddGenerator(MonsterId.Onion, 10);
AddSpawnPoint(OnionGenerator, TimeSpan.FromSeconds(0), "f_siauliai_west", Spot(-575, -1300));`
OR
`AddSpawnPoint(AddGenerator(MonsterId.Onion, 10), TimeSpan.FromSeconds(0), "f_siauliai_west", Spot(-575, -1300));`

As for the naming convention, we can use either 'MonsterPopulation' or 'Generator', as well as either 'AddSpawnPoint' or 'AddSpawner' please let me know which ones you prefer.

kenedos:
I have updated the code so that now users can either create a MonsterGeneration (to have a shared max population) or create a SpawnPoint directly by providing the amount of population you want in that spawn and the monster class.

As it stands, you can create an Onion spawn with either:

`var Onion = AddMonsterGenerator(MonsterId.Onion, "f_siauliai_west", 35);`
`AddSpawnPoint(Onion, TimeSpan.FromMilliseconds(0), "f_siauliai_west", Spot(-1545.54, -1109.33, 80));`
**OR**
`AddSpawnPoint(MonsterId.Onion, 20, TimeSpan.FromMilliseconds(0), "f_siauliai_west", Spot(-1545.54, -1109.33, 80));`

I have also updated the coding conventions and removed files with empty respawn points.

I have done a small refactor and renamed the "MonsterSpawner" to "MonsterSpawnPoint" and "MonsterPopulation" to "MonsterGenerator", as you have suggested. I have also removed necessity of providing a name string to the MonsterGenerator and use it as a variable instead. I have kept the map parameter in the generator and ended up not creating multiple child classes with an abstract base as you suggested, please let me know if this is undesirable.

Furthermore, I also lowered the FlexMeterDecreasePerMillisecond by 10x, because it was off by one decimal case.

exectails:
Creating generators behind the scenes for what was previously a spawner is a neat solution to keep backwards compatibility, but I'm a little concerned about the complexity and customizability, and I'm also not sure that your current solution actually supports everything the game supports.

It's my understanding that the game creates anchors that define positions on a map and generators that spawn monsters around those positions. This is pretty straightforward and it's what I tried to emulate in my example. I should mention that I didn't intend for cross-map-spawners to be a thing, even though there isn't really anything preventing us from doing that. I believe the population limit is not a limit per map, but per generator, and the generator/spawner is already keeping track of monsters it spawns. It would just randomly pick a spawn point/area from a list instead of having only one to spawn monsters in. This might actually be a desirable feature to have, as you could, for example, define 10 random spawn points around the world for a special monster to appear in, with a max count of 1. But if we didn't want that, we could simply enforce that all spawn points with the same name are on the same map.

Honestly, your new version feels even more complex than the old one to me, because with the new naming scheme it now seems like you're creating spawners that are then assigned to spawn areas. On the surface it seems like the exact opposite of what the game appears to be doing, which actually confused me for a moment until I wrapped my head around the fact that the "generator" is just the population data. Though it seems like you were trying to follow my suggestion there, which leads me to wonder whether we're on the same page in regards to what the game is doing.

My previous suggestion was basically to change the current spawners to accept multiple spawn areas by way of creating "spawn points" (anchors). First you define where a monster can spawn and then you create a spawner/generator that spawns/generates the monsters. The generator would choose a random location from its list of spawn areas when spawning monsters and it would limit the amount to its maximum population, which is the same as the current max amount in our spawners.

As I'm writing this message, I realize I made a mistake. If you were wondering why my description of the anchors and gentypes didn't make sense, it's because it's wrong^^ I stumbled over this before, but forgot about it. Now I finally get where this design decision came from, you were following the client data perfectly. The anchors actually are the spawners and use the monster and population data from the gentypes... Such a weird system <__> I get that for data generation purposes it would be easiest to mimic this system, but do we really want to spring that on Melia's (future) users? =/

How do you feel about the system I suggested? Would it be easy to understand and support the things needed to emulate the game's spawns? Unless I'm missing something, I think it'd be much easier to wrap your head around, would not require shared data in Map, and require very little changes to the current spawners, which could also support the old spawning method without additional work, by defining only one spawn area for them on-the-fly.

kenedos:
I like your idea. Honestly, it feels a lot more organized to have 1 generator associated to N spawn points than N spawn points associated to 1 generator. In terms of architecture it may mean the same, but in terms of implementation, it means we have a list of spawn points in the generator class, rather than a property in each spawn point defining which generator that spawn point belongs to. It is a design option I didn't consider up until you mentioned it.

As for the pseudo-code, I think it is a great idea to have your suggested architecture as we could get rid of defining the map both in the generator and the spawn point (we can define only in the generator now). Also, something I didn't realize up until now is that we can define the respawn delay only in the generator, as you suggested, instead of per spawn point (it would still be compatible with client data).

One slight different thing I would prefer, however, is to add spawn points via method call, instead of defining them on spawner creation or associating them to the "OnionSpawnLocations" string which then links it to the spawner. As we take that and also move the map definition to the spawner (instead of the spawn point) we would have the pseudo code similar to:

`var OnionSpawner1 = AddSpawner(MonsterId.Onion, 10, TimeSpan.FromSeconds(5), "f_siauliai_west")`
`OnionSpawner1.AddSpawnPoint(Spot(-575, -1300));`
`OnionSpawner1.AddSpawnPoint(Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));`

I don't know, feels more organized that way, and it also keeps open the possibility of adding other methods to that spawner in future. For example, I know we'll have to eventually implement hidden mob spawn points as well (those that fall from sky when you're moving around the map). So we could use like "spawnerObj.AddHiddenSpawnPoint(...)", etc.

And I believe I can still keep support for calling the spawn point directly, without having to explicitly define a spawner, by internally creating the spawner and associating the spawn point to it via shortcuts (as I did on my previous implementation). Hence, the pseudo-code would be like what we originally wanted:

`AddSpawnPoint(MonsterId.Onion, 20, TimeSpan.FromMilliseconds(5), "f_siauliai_west", Spot(-1545.54, -1109.33, 80));`

What do you think?

exectails:
> ```cs
> var OnionSpawner1 = AddSpawner(MonsterId.Onion, 10, TimeSpan.FromSeconds(5), "f_siauliai_west")
> OnionSpawner1.AddSpawnPoint(Spot(-575, -1300));
> OnionSpawner1.AddSpawnPoint(Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));
> ```

I feel like this design complicates the spawner creation slightly. It's not quite as easy to grasp at a glance as `add spawn point, add spawn point, add spawner` and it makes the scripts appear more like core code than scripts if that makes sense. You also couldn't easily reuse spawn points. For example, one could run around the map and set up a series of locations and then use those for all monsters. That's not possible if you assign them separately to each spawner, and you could end up with a ton of duplicated code.

kenedos:
Alright, so, I feel pretty satisfied with current implementation that include the changes you suggested.

Let me try and summarize everything I did:
- We have Spawner and SpawnerPoint classes.
- Spawner is now stored in WorldManager and is not tied to any particular map.
- SpawnPoint is tied to a map and an area.
- To create a SpawnPoint, you need a Spawner, which is identified by its name string (as we discussed).
- Respawn delay is tied to the Spawner, not the SpawnPoint.
- Spawner has a list of SpawnPoints in it and mobs spawn randomly in any of its SpawnPoints.
- Now mobs of a same population can spawn in different maps (i.e. world boss random map spawn support)
- Also created shortcuts to allow scripts to add SpawnPoints directly without explicitly creating Spawners (Spawner is created internally).

So you can either add Onions with:
`AddSpawner("Spawner1", MonsterId.Onion, 15, TimeSpan.FromMilliseconds(0));`
`AddSpawnPoint("Spawner1", "f_siauliai_west", Spot(-978, 291, 200));`
OR
`AddSpawnPoint("Spawner1", MonsterId.Onion, 10, "f_siauliai_west", TimeSpan.FromSeconds(5), Spot(-978, 291, 200));`

I feel we have basically everything covered now, perhaps aside from the spawn point collection idea, but not sure if that's going to be used much as maps are very different and don't share coordinates that much.

exectails:
How do you feel about moving the spawn point name to the back of the arguments? I feel like we don't need it for spawners with only one point, so you could basically use either a shape _or_ a spawn point name as the "location arguments". And if you also name that overload `AddSpawner`, we don't introduce breaking changes. The old code will keep working and the creation calls will be consistent.

The way I would probably have done it is simply add another overload that takes a string instead of the map name and shape, turned the one shape in the spawner class into a list of shapes, and if we're coming from the spawn point name overload, I'd get all spawn points by name and initialize the spawner with them.

kenedos:
I think we may be having different views and opinions regarding this mob spawning architecture. The SpawnPoint class is supposed to represent the Anchor and the Spawner supposed to represent the GenType. The SpawnPoint is not just an area or list of areas to the Spawner, it is a different entity. 

Right now we're only implementing the area and map the SpawnPoint belongs to, but in future, (if we are to emulate official behaviour) we also need it to have other properties such as how far mobs can wander off and trigger "OnEnter" events to spawn hidden mobs with certain chance whenever players pass through it.

From my understanding, you want to accomplish the entire architecture with a single class, the Spawner, and have a list of areas and maps in it representing the "spawnpoints". This would work for now, yes, but it'll become more complicated when we need to associate a mob to its anchor or when we want to trigger events upon entering it.

As for creating the SpawnPoint shortcut overload that does not take the name string, I agree with you. I will do that and create more overloads, which will internally "auto-generate" a Spawner name and instantiate it. I'm not really fond of sending the name parameter to the back of the arguments, it is basically the most important parameter the SpawnPoint class receives and it is also what associates it to the Spawner. We can have shortcut overloads where the name is hidden, but sending it to the back of the arguments would give the impression the name is not as important as it is.

And regarding backwards compatibility with old code, I am honestly a bit confused. What old code are you referring to? The Spawner class was completely overhauled in this PR. I changed the way the flex spawns work and also added bugfixes to it because the Spawner in master branch was not working correctly.

exectails:
> What old code are you referring to?

Well, the existing code. Granted, it's primarily test code, but there are people who already created and/or generated spawners. Imo it's always a good idea not to break anything if you can easily avoid it, and it would also keep the function calls consistent.

> I changed the way the flex spawns work and also added bugfixes to it because the Spawner in master branch was not working correctly.

Could you elaborate on this? You didn't mention any fixes before and I don't see a problem with the spawner on my local copy.

> I think we may be having different views and opinions regarding this mob spawning architecture.

I guess so. I worry we're talking past each other, because after our last exchange I thought we were in agreement that simply extending the current system to support multiple spawn areas would allows us to recreate the spawns we're seeing in-game while keeping the system comprehensible.

Btw, I'm wholly opposed to a "spawn point" doing anything but defining where something is and a "spawner" not spawning anything but defining spawn parameters. This is primarily a naming issue, but to me the spawner would need to do the spawning and the spawn area would define where the spawning takes places. Am I the only one who finds the game's approach and data really confusing?^^

> we also need it to have other properties such as how far mobs can wander off

I'm not sure about that. The gentype data includes a range, yes, but the way I understand it, the range + the anchor point gives us the area in which the monsters can spawn. In Melia we define this area with a shape to have more control over it. The range a monster can move away from that location before deciding to return to its spawn area is not defined in that data, but primarily comes from either the map or the AI data. Personally I'd want to put it in the AI.

> and trigger "OnEnter" events to spawn hidden mobs with certain chance whenever players pass through it

I would've made that an entirely separate mechanic instead of trying to shoehorn it into any existing spawner systems. It's a simple trigger area that spawns monsters on contact, or perhaps on certain other events. Does the client data actually use the anchors and gentypes for this as well? I think we could get a lot more versatility out of it by separating these systems.

kenedos:
> Could you elaborate on this? You didn't mention any fixes before and I don't see a problem with the spawner on my local copy.

The current code (in your repo) makes use of a list of respawn delays but the methods that use that list (RespawnMonsters and FlexSpawnMonsters) were not working correctly. Monsters wouldn't respawn to their max amount.

To be honest, I was confused with the initial implementation of how the respawns and flex respawns work. In my PR, I changed logic to no longer make use of that list. Also, I changed how the FlexSpawnMonsters method works to be more similar to RespawnMonsters in attempt to make it more intuitive. You can check the changes in these methods for more information.

> I guess so. I worry we're talking past each other (...)

I don't want to waste either of our times discussing solutions, so instead, let me just see if I understand correctly what you're proposing:

1. I should create a list in Spawner with type List<Tuple<string, Area>> where the first type (string) is the spawn point map and second type (area) is the spawn location.
2. Adding new spawn points adds elements to that list
3. Bring all SpawnPoint class methods into Spawner class
4. Change Spawner class code to use that list instead of SpawnPoint class
5. Completely delete SpawnPoint class, including all references of it in Map.cs
6. Send Name parameter to the back of the parameters list (should it be the last parameter?)
7. Add shotcut overload for SpawnPoint that does not receive the Name parameter at all

If those items are correct, I will make necessary changes and update this PR. Please advise on any further required change.

exectails:
> I changed logic to no longer make use of that list

You need a list to get correct respawn times per monster. For example, if I have a spawner with 3 Kepa and a fixed respawn delay of 10 seconds, and I kill them 3 seconds apart from each other, I want them to also respawn 3 seconds apart. The first on the 10th second after the first kill, the second after the 13th second, and the last on the 16th second after the first kill. This allows scripters to have control over the respawning, which can be useful if you don't spread the spawns out over the entire map.

If you remove the list you can't keep track of this and have to resort to filling up the spawns on a timer, which will "re"spawn monsters before their unique respawn delay. That's also why there's two spawn methods. One respawns monsters on their own timers so-to-speak, while the other exists only to fill up the spawns when the flex amount increases.

> Monsters wouldn't respawn to their max amount.

Was this perhaps a misunderstanding stemming from the confusion over the current system? I just tested it again, and I don't see any issues with the flex amount rising to the max. Though I could imagine that it would be difficult to reach the current threshold with the monsters potentially being spread across the map.

> I don't want to waste either of our times discussing solutions

I hope I didn't come across as dismissive or anything. I very much believe in discussing solutions, particularly because you spent more time with the game's spawn data than I did. I could simply propose a solution and you could implement it, but if my approach were to not match our needs going forward, we'd need another refactor later on.

> let me just see if I understand correctly what you're proposing

Your understanding of my proposal is basically correct, though I'd say it's easier to describe with the current system as the baseline.

1. Convert `MonsterSpawner.Area` from a single shape to a list, `MonsterSpawner.Areas`, from which one is chosen randomly on spawn.
2. Add `AddSpawnPoint` or `AddSpawnArea` shortcut to create lists of spawn areas.
3. In `AddSpawner`, take either a map name and a shape, _or_ the name of, or reference to, a spawn area list.
4. Move spawners to world.

This proposal assumes that our current system can mimic the game's system by simply adding an array of spawn areas. But it seems to me like this would allow for more flexibility.

kenedos:
> I just tested it again, and I don't see any issues with the flex amount rising to the max.

Well, I checked out to 2938202521caa98a25da924b2f1c5d07945bbebd and checked the location. I only see 1 kepa and 1 big kepa. Even if I set big kepa max amount to 10, I still only see 1 of each.

image

> Your understanding of my proposal is basically correct, though I'd say it's easier to describe with the current system as the baseline.

We will need to support crossmap spawn points too, hence why I suggested the List<Tuple<string, Area>> for map and area. We will need it for world boss spawning support (they spawn randomly on X different maps). What's your approach on this?

exectails:
> Well, I checked out to https://github.com/NoCode-NoLife/melia/commit/2938202521caa98a25da924b2f1c5d07945bbebd and checked the location. I only see 1 kepa and 1 big kepa. Even if I set big kepa max amount to 10, I still only see 1 of each.

That's what the flex spawn system is about. It's a load balancer, it doesn't spawn all monsters at once, but only a subset, and the amount increases based on the activity in the area. The idea is you don't overpopulate an area unless players are actively killing monsters there, in which case the amount increases up to the max. In the original code you can see that we default the min amount to 5% of the max amount.
```cs
public static MonsterSpawner AddSpawner(int monsterClassId, int maxAmount, TimeSpan respawn, string map, IShape area, TendencyType tendency, PropertyOverrides propertyOverrides)
{
	var initialSpawnDelay = TimeSpan.FromSeconds(0);
	var minRespawnDelay = Math2.Max(TimeSpan.FromSeconds(3), respawn);
	var maxRespawnDelay = minRespawnDelay.Multiply(3);
	var minAmount = Math.Max(1, (int)(maxAmount * 0.05));

	return AddSpawner(monsterClassId, minAmount, maxAmount, initialSpawnDelay, minRespawnDelay, maxRespawnDelay, map, area, tendency, propertyOverrides);
}
```
The reason we chose such a small amount for the moment was that the data Akinue saw in the client wasn't quite compatible with our spawner and we ended up with dozens of monsters inside very small spawn areas. Lowering the min amount to 5% was a temporary fix until we were to look into the spawn data again. If you kill the spawning Kepas, the flex meter will slowly rise until the flex amount increases, spawning 2 Kepas, then 3, then 4, etc, until you reach the max of 10. And if you stop killing them, the flex amount slowly falls again, not respawning monsters until the current amount falls below that number again.

> We will need to support crossmap spawn points too, hence why I suggested the List<Tuple<string, Area>> for map and area. We will need it for world boss spawning support (they spawn randomly on X different maps). What's your approach on this?

How we store the spawns is an implementation detail that depends on how we create the spawn areas, isn't it? You could create named spawn area lists that you store somewhere as a `List<Tuple<string, Area>>`, or you could make it a `Dictionary<string, List<SpawnArea>>`, or you could create objects that you pass to spawners, and so on. For example:
```cs
AddSpawnArea("Onions", "f_siauliai_west", Spot(-575, -1300));
AddSpawnArea("Onions", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));

AddSpawner(MonsterId.Onion, 10, TimeSpan.FromSeconds(5), "Onions");

void AddSpawnArea(...)
    someListSomewhere.Add(new SpawnArea(...));

// or

var onionSpawns = CreateSpawnAreaGroup();
onionSpawns.Add("f_siauliai_west", Spot(-575, -1300));
onionSpawns.Add("f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));

AddSpawner(MonsterId.Onion, 10, TimeSpan.FromSeconds(5), onionSpawns);
```
Regardless, the spawner would then randomly take one of these areas to spawn monsters in every time it needs to spawn something. It would simply get the map and add the new monster to it. Am I mistaken in assuming that there doesn't need to be a direct association between map and spawner?

I was approaching this from a point of least resistance and maximum flexibility. Our spawners are limited to one spawn area, which isn't compatible with the game's data, where monsters can spawn in multiple locations. The solution: Let one spawner have multiple potential spawn locations to choose from. Minimal changes, backwards compatibility to existing code, and (hopefully) easy to use.

I feel like I didn't get my idea across very well. I apologize if my description was unclear .__.

kenedos:
Closing this, will open new PR

PR #225, opened 2023-06-25
Fixed issue that was blocking log in on the server.

felipecastagnarodecarvalho:
It seems like the index was being wrongly get from the packet CB_START_GAME.


felipecastagnarodecarvalho:
I have gather the packet data:

{------------------------------------------------------------------------------
Op: 000C CB_START_GAME, Size: 32 (Table: 26, Garbage: 6)
------------------------------------------------------------------------------
00000000   0C 00 0A 00 00 00 05 00 00 00 00 00 00 00 00 00   ................
00000010   00 00 00 00 00 00 00 00 02 01 36 0E D4 66 0A 99   ..........6..f..
------------------------------------------------------------------------------}

for some reason the last bytes were being get: 0x01 0x02 causing it to seem like 258 (decimal)

felipecastagnarodecarvalho:
by the way, I remember that there were a discord for this project, can you send me the invitation?
Thanks.

SalmanTKhan:
hey @felipecastagnarodecarvalho,
Welcome back; the discord link is on the project "readme" page. But here's a direct link: https://discord.gg/5sszEgw
Yeah, I had a similar issue. Do you have a companion/pet?

exectails:
Thank you for your PR and for bringing this issue to my attention. I took a look at your log and it does indeed appear as if `characterIndex` is a byte. Though the question now is whether the last byte of the packet is padding or a yet unknown value. We wouldn't want to discard and forget about that byte in case we might need it in the future, so we should document and read it regardless.

Since you accidentally pushed a change to delete the example database conf file, I'll close the PR and push a fix for this problem. Thanks again!

PR #226, opened 2023-06-27
Adding ObliqueFire and DaggerSlash Skills (Scout)

felipecastagnarodecarvalho:
This is my second PR but the first that I am starting to really implemented something new somehow.

About the implementations:
 - Some values are hard coded.
 - It can be optimized.


src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -0,0 +1,125 @@
...
+				skillHit.ForceId = ForceId.GetNew();
+
+				hits.Add(skillHit);
+			}
+
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, farPos, hits);
+
+			await Task.Delay(secondHitTime);
+
+			hits = new List<SkillHitInfo>();
```
exectails, line 107: In cases where you want to reuse a list I would recommend `.Clear`ing it instead of creating an entirely new list, which might cause an allocation.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -0,0 +1,125 @@
...
+			var secondHitTime = TimeSpan.FromMilliseconds(550);
+			var damageDelay = TimeSpan.FromMilliseconds(270);
+			var damageSecondDelay = TimeSpan.FromMilliseconds(550);
+			var skillHitDelay = TimeSpan.Zero;
+			
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			if (targets.Any() && targets.Count > 6)
+			{
+				targets = targets.GetRange(0, AoeAttackRatio);
```
exectails, line 85: After briefly reading up on what exactly the AoE Attack Ratio is supposed to do I must ask, is this actually correct for the current version of the game or is it more of an approximation because we don't have this implemented yet?

The way I understand it, we'd have to factor in the monsters' AoE Attack/Defense Ratios (SDR) and not only stop hitting targets once we reach our own SDR, but also sort them by this value and potentially their distance, right?

Reference: https://www.reddit.com/r/treeofsavior/comments/3w3kjv/aoe_attack_ratio/

Additionally, the skill states it has an SDR of **+6**. That sounds to me like a modifier on top of the character's default SDR. Should the ratio be higher than 6 then?

It's fine if you just added this as a temporary solution, though a comment about it can be helpful for reference in such cases.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+			skillHit.ForceId = ForceId.GetNew();
+
+			skillHits.Add(skillHit);
+
+			var originPos = caster.Position;
+			var hits = new List<HitInfo>();
+			var splashArea = new Circle(designatedTarget.Position, skill.Data.SplashHeight * 2);
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			if (targets.Any())
```
exectails, line 70: May I make a suggestion?^^
```cs
var nearestTarget = targets.Where(a => a != designatedTarget).OrderBy(a => a.Position.Get2DDistance(designatedTarget.Position)).FirstOrDefault();
```

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+						
+						distancesList.Add(target.Position);
+					}
+				}
+
+				var nearstTargetDistance = distancesList.OrderByDescending(v2 => GetDistance(designatedTarget.Position, v2)).LastOrDefault();
+
+				var nearstTarget = targetDistanceDic[nearstTargetDistance];
+
+				if (nearstTarget != null)
```
exectails, line 92: It seems you're doing this a bit differently from what I'm seeing in the packet logs, is this intentional? The way the skill appears to work is that there's no hits sent via `ZC_SKILL_FORCE_TARGET`, `ZC_SKILL_HIT_INFO` is sent with one hit for the main target ~120ms later and the additional hit is sent via `ZC_HIT_INFO` at the same time.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+				return;
+			}
+
+			if (!caster.Position.InRange2D(designatedTarget.Position, skill.Data.MaxRange))
+			{
+				caster.ServerMessage(Localization.Get("Too far away."));
+				return;
+			}
+
+			var damageDelay = TimeSpan.FromMilliseconds(200);
```
exectails, line 52: I have a `damageDelay` of 270 in my skill info packet, though I believe this value is affected by the attack speed, so 200 might be fine.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+			designatedTarget.TakeDamage(skillHitResult.Damage, caster);
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+			skillHit.ForceId = ForceId.GetNew();
+
+			skillHits.Add(skillHit);
+
+			var originPos = caster.Position;
+			var hits = new List<HitInfo>();
+			var splashArea = new Circle(designatedTarget.Position, skill.Data.SplashHeight * 2);
```
exectails, line 67: I feel like we shouldn't make arbitrary calculations on data values to make the data work, as it seems unlikely that that's the intended way. If they had wanted the radius to be that amount, they probably would've made SplashHeight twice as big. It's not pretty, but as long as we don't have a reliable source for these values we should rather use raw numbers. That way it seems less like this is presumed to be official behavior.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+using Yggdrasil.Extensions;
+using static Melia.Zone.Skills.SkillUseFunctions;
+
+namespace Melia.Zone.Skills.Handlers.Scout
+{
+	/// <summary>
+	/// Handles the Scout skill Oblique Fire.
+	/// </summary>
+	[SkillHandler(SkillId.Scout_ObliqueFire)]
+	public class ObliqueFire : ITargetSkillHandler
```
exectails, line 23: This handler doesn't seem to include the buff/debuff. Our current implementation of it is almost certainly incorrect, but at the very least the buff should be started. Ideally it would also be fixed up to do what it's supposed to do.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,112 @@
...
+			var damageDelay = TimeSpan.FromMilliseconds(200);
+			var skillHitDelay = skill.Properties.HitDelay;
+
+			var skillHits = new List<SkillHitInfo>();
+
+			var skillHitResult = SCR_SkillHit(caster, designatedTarget, skill);
+			designatedTarget.TakeDamage(skillHitResult.Damage, caster);
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+			skillHit.ForceId = ForceId.GetNew();
```
exectails, line 61: Same as Dagger Slash, I don't see a force id in my logs.

exectails (changes requested):
Thank you for your PR! I noticed a couple points that should be addressed before it's merged.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -0,0 +1,125 @@
...
+
+			foreach (var target in targets)
+			{
+				var skillHitResult = SCR_SkillHit(caster, target, skill);
+				var damage = skillHitResult.Damage * SkillFactor / 2;
+
+				target.TakeDamage(damage, caster);
+
+				var skillHit = new SkillHitInfo(caster, target, skill, skillHitResult, damageDelay, skillHitDelay);
+				skillHit.ForceId = ForceId.GetNew();
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I misunderstood its purpose. Its not necessary.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
felipecastagnarodecarvalho: I saw that the SlashHeight was being used to calculate the circle radius in the source files.
But in the case of this skill it, the range doesn't seem to match the range in the official server.
Multiplying by 2 give me a good satisfied result.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
felipecastagnarodecarvalho: great!

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,102 @@
...
+
+			if (nearestTarget != null)
+			{
+				var skillHitResultPost = SCR_SkillHit(caster, nearestTarget, skill);
+				nearestTarget.TakeDamage(skillHitResultPost.Damage, caster);
+
+				var skillHitPost = new SkillHitInfo(caster, nearestTarget, skill, skillHitResultPost, damageDelay, skillHitDelay);
+				var hit = new HitInfo(caster, nearestTarget, skill, skillHitResult.Damage, skillHitResult.Result);
+
+				nearestTarget.Components.Get<BuffComponent>().Start(BuffId.ObliqueFire_Buff, skill.Level, 0, TimeSpan.FromSeconds(10), nearestTarget);
```
exectails, line 91: You're applying the buff to the targets, but is this actually the correct behavior? Wouldn't this increase the monsters' movement speed on attack? I recall the skill description saying that it increases _your_ speed, while decreasing the targets' speeds after a certain number of stacks. I'm not sure about the exact behavior myself, but I don't think that we're supposed to buff the targets^^

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,102 @@
...
+
+				Send.ZC_HIT_INFO(caster, nearestTarget, skill, hit);
+			}
+		}
+
+		private static double GetDistance(Position initialPosition, Position mobPosition)
```
exectails, line 97: This method can be removed, right?

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
exectails: Right. That's totally fine, but this way it reads like this is intended, official behavior, but that seems rather unlikely and could lead to issues or confusion down the line. That's why it would be better to hardcode the value and potentially make a note of it instead of doing calculations on arbitrary data that happened to contain values that somewhat worked.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
```diff
@@ -0,0 +1,102 @@
...
+			var secondHitTime = TimeSpan.FromMilliseconds(100);
+
+			var skillHitResult = SCR_SkillHit(caster, designatedTarget, skill);
+			designatedTarget.TakeDamage(skillHitResult.Damage, caster);
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+
+			designatedTarget.Components.Get<BuffComponent>().Start(BuffId.ObliqueFire_Buff, skill.Level, 0, TimeSpan.FromSeconds(10), designatedTarget);
+
+			Send.ZC_SKILL_FORCE_TARGET(caster, designatedTarget, skill, skillHit);
```
exectails, line 73: > I could`t make it work with Z_KILL_HIT_INFO, in game the client did not respond to the skill altough the mob dies. My solution was Send.ZC_SKILL_FORCE_TARGET(caster, designatedTarget, skill, skillHit);

I understand that this solution works for you, but we should really aim to emulate the game's packets and packet sequences as closely as possible, as to not risk running into any unexpected situations where things break and we don't understand why.

You're saying it didn't work with `ZC_SKILL_HIT_INFO`, but did you try to send `ZC_HIT_INFO` for the second hit?

exectails (changes requested):

src/ZoneServer/Buffs/Handlers/DaggerSlash_Buff.cs
```diff
@@ -0,0 +1,32 @@
...
+	{
+		/// <summary>
+		/// Starts buff, modifying the attack.
+		/// </summary>
+		/// <param name="buff"></param>
+		public override void OnStart(Buff buff)
+		{
+			var target = buff.Target;
+
+			// TODO: Something with the attack based on number of daggers equiped.
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I do not know how to get the number of daggers equiped yet so I will leave like that for now.

src/ZoneServer/Buffs/Handlers/ObliqueFire_Buff.cs
```diff
@@ -12,42 +12,61 @@ namespace Melia.Zone.Buffs.Handlers
 	[BuffHandler(BuffId.ObliqueFire_Buff)]
 	public class ObliqueFireBuffHandler : BuffHandler
 	{
-		private const string VarName = "Melia.StatModifier";
-		private const float BuffBonus = 3f;
-		private const float DebuffPenaltyRate = -0.05f;
+		private const string ModifierVar = "Buff:ObliqueFire_Buff/Modifier";
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I guess there was a problem with this PR, I was using an old `ObliqueFire_Buff.cs` file.

src/ZoneServer/Buffs/Handlers/ObliqueFire_Buff.cs
```diff
@@ -12,42 +12,61 @@ namespace Melia.Zone.Buffs.Handlers
...
+				{
+					var modifier = targetCharacter.Variables.Perm.Get<float>(ModifierVar, 0);
+					targetCharacter.Variables.Perm.SetFloat(ModifierVar, modifier + add);
+				}
 			}
 			else
 			{
-				var mspd = target.Properties.GetFloat(PropertyName.MSPD);
-				var penalty = mspd * DebuffPenaltyRate;
+				// TODO: This part needs more research. There doesn't seem
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: yeah that was what happened.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
felipecastagnarodecarvalho: Good question. From what I have tested in game it applies on the mob instead of applying a buff on yourself.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
felipecastagnarodecarvalho: ZC_SKILL_HIT_INFO doesn't work unfortunately, maybe I missed some other packet that I must send before it?
The thing is that ZC_SKILL_FORCE_TARGET fits well and ZC_HIT_INFO does as well.

> You're saying it didn't work with ZC_SKILL_HIT_INFO, but did you try to send ZC_HIT_INFO for the second hit?

I am trying to follow the server packet flow. It does it on the official but with ZC_SKILL_HIT_INFO instead of ZC_SKILL_FORCE_TARGET.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -0,0 +1,119 @@
...
+		public void Handle(Skill skill, ICombatEntity caster, Position originPos, Position farPos, ICombatEntity designatedTarget)
+		{
+			if (!caster.TrySpendSp(skill))
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.TurnTowards(designatedTarget);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: On my tests in game turning worked well without affecting the calculations, it seems to hit whatever was on the direction that he turned.

src/ZoneServer/Buffs/Handlers/DaggerSlash_Buff.cs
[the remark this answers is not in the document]
exectails: The way I would interpret the skill description is that you always get the buff with a dagger equipped, but _if_ you have weapons equipped in the secondary slot you only get the bonus if both main weapons are daggers. Is that not how it works? If it is, it would be fine to implement it to always apply for now and come back to it later.

Regarding how to get the equipped items, you can simply query them from the inventory if the entity is a `Character`.
```cs
character.Inventory.GetEquip(EquipSlot.RightHand)
character.Inventory.GetEquip(EquipSlot.LeftHand)
character.Inventory.GetEquip(EquipSlot.RightHandSub)
character.Inventory.GetEquip(EquipSlot.LeftHandSub)
```

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
exectails: I just checked my logs again and the buff is definitely applied to the attacker, which makes, because that's what the skill description says^^ I guess we have to pivot to the target after reaching a certain overbuff. This should ideally get logged, so we know exactly what's going on. It's very much possible that a second, other buff is applied to the monster once that happens.

src/ZoneServer/Skills/Handlers/Scout/ObliqueFire.cs
exectails: Well, we need to figure out the missing part. I'm all for not blindly following the packet logs, but the packets we do send should be the right ones. We don't want to hack our way around such issues, as something is clearly going wrong with our code if the correct packets don't work as intended.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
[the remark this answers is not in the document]
exectails: That might be, but you didn't address my question^^ Is this actually what happens based on your logs? See, if my character looks straight up, and the designated target is at an angle of 10° from it, and it turns 10° towards the target, that means it will potentially hit different targets than it would've if the character hadn't turned. It's not a major issue, but a fine detail that has the potential to feel a bit wrong in-game.

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -0,0 +1,119 @@
...
+			var skillHitDelay = TimeSpan.Zero;
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			if (targets.Count > AoeAttackRatio)
+			{
+				targets = targets.GetRange(0, AoeAttackRatio);
+			}
+
+			foreach (var target in targets)
```
exectails, line 86: There's now a way to limit the iteration over the targets by the SDR. Please check other skills for an example.

exectails:
I made a couple minor changes and hopefully fixed up the buffs correctly before merging. Feel free to make another PR if I made any mistakes. Thanks again for the PR.

PR #227, opened 2023-06-29
Add ObliqueShot and TwinArrows skills (archer)

felipecastagnarodecarvalho:
There a few minor things to adjust like adding a effect that goes from the target hit to the new hit enemy.


src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
```diff
@@ -0,0 +1,102 @@
...
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+			skillHit.ForceId = ForceId.GetNew();
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, caster.Position);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
+			Send.ZC_SKILL_FORCE_TARGET(caster, designatedTarget, skill, skillHit);
+
+			// It has 50% of being giving slow of 7 seconds to the target enemy			
+			Random random = new Random();
```
exectails, line 73: If you just quickly need a `Random` instance, please use `RandomProvider.Get()` to get it, instead of creating it manually. That static class automatically creates them with a random seed, improving the numbers, and on a per-thread basis, so we create less of them.

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
```diff
@@ -0,0 +1,102 @@
...
+			bool shouldExecute = random.Next(2) == 0;
```
exectails, line 74: We encourage the use of the the `var` keyword, as per our coding conventions.

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
```diff
@@ -0,0 +1,102 @@
...
+			var secondHitTime = TimeSpan.FromMilliseconds(100);
+
+			var skillHitResult = SCR_SkillHit(caster, designatedTarget, skill);
+			designatedTarget.TakeDamage(skillHitResult.Damage, caster);
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+			skillHit.ForceId = ForceId.GetNew();
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, caster.Position);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
```
exectails, line 69: Did you intend to send both `ZC_SKILL_MELEE_GROUND` and `ZC_SKILL_FORCE_TARGET`? Only the latter appears in my logs.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
```diff
@@ -0,0 +1,79 @@
...
+		/// <param name="caster"></param>
+		/// <param name="designatedTarget"></param>
+		private async void Attack(Skill skill, ICombatEntity caster, ICombatEntity designatedTarget)
+		{
+			var damageDelay = TimeSpan.Zero;
+			var skillHitDelay = skill.Data.DefaultHitDelay;
+
+			// TODO: Add more 50% damage to enemies using cloth armor type
+
+			var skillHitResult = SCR_SkillHit(caster, designatedTarget, skill);
```
exectails, line 60: I see you're generating two skill hits here, but the way the game typically handles double-hits is with one modified skill hit that splits the damage in half and shows two hits. You can take a look at how we currently handle this for other skills in `SCR_CalculateDamage`.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
```diff
@@ -0,0 +1,79 @@
...
+			designatedTarget.TakeDamage(skillHitResult.Damage, caster);
+
+			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, caster.Position);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
```
exectails, line 66: According to my logs this should be `ZC_SKILL_FORCE_TARGET`. What about yours?

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
```diff
@@ -0,0 +1,102 @@
...
+
+		/// <summary>
+		/// Executes the actual attack.
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="designatedTarget"></param>
+		private async void Attack(Skill skill, ICombatEntity caster, ICombatEntity designatedTarget)
+		{
+			var damageDelay = TimeSpan.Zero;
```
exectails, line 58: I logged a delay of 45, does it work correctly with 0?

exectails (changes requested):

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
felipecastagnarodecarvalho: Yeah I just tested again, those seems not to be necessary at all.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
felipecastagnarodecarvalho: gotcha

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
felipecastagnarodecarvalho: it seems to do so, 45ms is almost imperceptible.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
felipecastagnarodecarvalho: Well, if I don`t send the ZC_SKILL_READY and ZC_SKILL_MELEE_GROUND before the ZC_SKILL_HIT_INFO, the character does`t correct does the animations + sound effect. If you find a better way feel free to tell me.

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
exectails: That might be, but do keep in mind that we're still in the process of figuring out how all of this works, and there is a big difference between "no delay" and "some delay", because "some delay" must come from somewhere. Any information we might get, like what kinds of values to expect, will help us figure out more down the line.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
exectails: What I mean is that you presumably shouldn't be sending `ZC_SKILL_MELEE_GROUND` but `ZC_SKILL_FORCE_TARGET`.

src/ZoneServer/Skills/Handlers/Archer/ObliqueShot.cs
```diff
@@ -0,0 +1,100 @@
...
+			var skillHitDelay = TimeSpan.Zero;
+			var secondHitTime = TimeSpan.FromMilliseconds(100);
+
+			var skillHitResult = SCR_SkillHit(caster, designatedTarget, skill);
+			designatedTarget.TakeDamage(skillHitResult.Damage / 2, caster);
```
exectails, line 64: Is splitting the damage between main target and sub-target official behavior?

felipecastagnarodecarvalho: I probably miss understood, I remember seeing a comment about it somehwer. But its wrong.
The damage should split at all.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
felipecastagnarodecarvalho: From my tests on the server, it seems like ZC_SKILL_FORCE_TARGET is not being received, altough ZC_SKILL_HIT_INFO is,
but for it to work we should use the other two packets that I mentioned.
I am getting redundant but I guess you get it.

I am gonna try to simple switch to ZC_SKILL_FORCE_TARGET to see if I get a better result.

src/ZoneServer/Skills/Handlers/Archer/TwinArrows.cs
```diff
@@ -62,9 +62,7 @@ private async void Attack(Skill skill, ICombatEntity caster, ICombatEntity desig
 
 			var skillHit = new SkillHitInfo(caster, designatedTarget, skill, skillHitResult, damageDelay, skillHitDelay);
 
-			Send.ZC_SKILL_READY(caster, skill, caster.Position, caster.Position);
```
exectails: I think you might be misunderstanding what I'm trying to say. I didn't mean you should remove `ZC_SKILL_READY` and `ZC_SKILL_HIT_INFO`, but only replace `ZC_SKILL_MELEE_GROUND` with `ZC_SKILL_FORCE_TARGET`, because that's what the game usually sends. And you also still need to take a look at the aforementioned scriptable function for how multi-hit skills works, because they generally don't send two hits, but only one that has a special property set to animate a double-hit.

exectails:
I made a couple minor changes and hopefully fixed up the buffs correctly before merging. Feel free to make another PR if I made any mistakes. Thanks again for the PR.

PR #228, opened 2023-07-02
Add Enchanter Skills

felipecastagnarodecarvalho:
List of enchanter skills implemented:
- Enchant Aura
- Enchant Glove
- Enchant Earth
- Agility
- Over Reinforce

Missing:
- Enchant Weapon (applied when learned)

Considerations:
- A few todo are insert like applying the buff to part members and applying attribute passives.
- The Enchant Aura should be modified since the effect is not being displayed


src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,68 @@
...
+
+				if (target != null && target.Inventory.GetEquip(EquipSlot.Gloves) != null)
+				{
+					if (buff.Data.Level == 1 || buff.Data.Level == 2)
+					{
+						speedBonus = 2;
+					}
+					else if (buff.Data.Level == 3 || buff.Data.Level == 4)
+					{
+						speedBonus = 3;
```
exectails, line 33: It seems like this code is incorrect. The skill description mentions a bonus of 1 for levels 1\~3, 2 for 4~6, etc. Am I missing something? This structure could also be simplified a little to make it more concise. What do you think?
```cs
speedBonus = (int)Math.Ceiling(level / 3f);
```

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,68 @@
...
+					}
+					else if (buff.Data.Level == 7 || buff.Data.Level == 8)
+					{
+						speedBonus = 5;
+					} else {
+						speedBonus = 6;
+					}
+
+					// TODO: Apply "Agility: Enchant" (passive adquire through attribute points)
+					// TODO: Reduce the STAMINA consuption rate
```
exectails, line 47: Based on the available data I assume the property to use for that is `MOVESTA_RATE_BM`.

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,68 @@
...
+						speedBonus = 4;
+					}
+					else if (buff.Data.Level == 7 || buff.Data.Level == 8)
+					{
+						speedBonus = 5;
+					} else {
+						speedBonus = 6;
+					}
+
+					// TODO: Apply "Agility: Enchant" (passive adquire through attribute points)
```
exectails, line 46: You can already do this, even if getting the abilities is currently not working.
```cs
if (character.Abilities.Has(AbilityId.Enchanter10))
    // ...
```

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,48 @@
...
+	/// </summary>
+	[BuffHandler(BuffId.EnchantEarth_Buff)]
+	public class EnchantEarth_Buff : BuffHandler
+	{
+		private float blockPenetrationBonus = 0;
+
+		public override void OnStart(Buff buff)
+		{
+			var target = buff.Target as Character;
+			var SCR_Get_Character_MAXPATK = ScriptableFunctions.Character.Get("SCR_Get_Character_MAXPATK");
```
exectails, line 19: Unless you have a reason to directly access the scriptable function, you should use the property instead, as it's simpler and reduces the risk for errors, because the property getter will never change.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,48 @@
...
+			if (target != null)
+			{
+				var caster = buff.Caster as Character;
+
+				// Apply penality when the CASTER Max Physical Attack is lower than the TARGET Max Physical Attack
+				// TODO: Find out the exacly value of the penality (We are applying 50%)
+				var penalityValue = SCR_Get_Character_MAXPATK(caster) < SCR_Get_Character_MAXPATK(target) ? 0.5f : 1f;
+
+				// Based on in game values (tooltip)
+				blockPenetrationBonus = (8f + (1.6f * buff.Data.Level)) * penalityValue;
```
exectails, line 30: Here you're using the static buff level once more.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,48 @@
...
+
+			if (target != null)
+			{
+				var caster = buff.Caster as Character;
+
+				// Apply penality when the CASTER Max Physical Attack is lower than the TARGET Max Physical Attack
+				// TODO: Find out the exacly value of the penality (We are applying 50%)
+				var penalityValue = SCR_Get_Character_MAXPATK(caster) < SCR_Get_Character_MAXPATK(target) ? 0.5f : 1f;
+
+				// Based on in game values (tooltip)
```
exectails, line 29: It's very possible that I'm missing something, but I don't quite understand your formula in relation to the in-game description. Enchant Earth is supposed to raise your Block Penetration by X% per skill level, where X appears to be `6 + (level -1)`. Shouldn't this buff be as simple as adding that value to `BLK_BREAK_RATE_BM`?

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Melia.Zone.Buffs.Handlers
+{
+	/// <summary>
+	/// Handle for the Enchant Earth Buff, which increases the target's block penetration
+	/// </summary>
+	[BuffHandler(BuffId.EnchantEarth_Buff)]
+	public class EnchantEarth_Buff : BuffHandler
+	{
+		private float blockPenetrationBonus = 0;
```
exectails, line 14: You can't save the modifier on the buff handler class, because this class is used globally by every player who gets this buff. You need to save it on the character or the buff, as other buffs are already doing.

src/ZoneServer/Buffs/Handlers/OverReinforce_Buff.cs
```diff
@@ -0,0 +1,41 @@
...
+		private float attackBonus = 0;
+
+		public override void OnStart(Buff buff)
+		{
+			var target = buff.Target as Character;
+
+			if (target != null)
+			{
+				// Based on in game values (tooltip)
+				attackBonus = 1308 + (349 * buff.Data.Level);
```
exectails, line 23: I'm starting to wonder whether we're using different games for referencing bonuses, because Over Reinforce is said to raise ATK by X on my client. I can't tell where the 1308 is coming from 🤔 That seems way, way too high. I'll stop the review here for the moment because there are clearly several issues to work out still with this PR.

Screenshot 2023-07-03 132349

exectails (changes requested):

src/ZoneServer/Buffs/Handlers/OverReinforce_Buff.cs
felipecastagnarodecarvalho: image
 Good point, I need to figurate out this formula.

felipecastagnarodecarvalho: I will follow the calc_property_skill.lua:

```
function SCR_GET_OverReinforce_Ratio(skill)
    local pc = GetSkillOwner(skill)
    local maxpatk = SCR_Get_DEFAULT_MAXPATK(pc, 0)
    local minpatk = SCR_Get_DEFAULT_MINPATK(pc, 0)
    local rate = 0.015 + skill.Level * 0.004

    local vibora_rate = GetExProp(pc, 'ITEM_VIBORA_Empowering') / 1000
    if vibora_rate > 0 then
        rate = rate + vibora_rate
    end

    local value = ((maxpatk + minpatk)/2) * rate
    
    return math.floor(value);
end
```

exectails: Can you tell me which version of the game that is? Because as you can see, my client shows _vastly_ different values, both for the bonus as well as for the SP usage.

felipecastagnarodecarvalho: I think its showing different values because I am max level with that character.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
felipecastagnarodecarvalho: `var penalityValue = SCR_Get_Character_MAXPATK(caster) < SCR_Get_Character_MAXPATK(target) ? 0.5f : 1f;`

This part is related to:

`The final efficiency is reduced when the highest Physical ATK of the caster is lower than the highest ATK of the target.`

src/ZoneServer/Buffs/Handlers/OverReinforce_Buff.cs
exectails: Ah, right. I overlooked that this value is dependent on your own PATK.

felipecastagnarodecarvalho:
I found a weird bug where the Enchant Gloves buff are applying 10 minutes instead of 30.
Other buffs are being applied at 30 minutes as the code describes it.

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -11,45 +14,44 @@ namespace Melia.Zone.Buffs.Handlers
...
 			{
-				// It is not applyed to characters without Gloves
+				var skillLevel = buff.NumArg1;
+				var caster = buff.Caster as Character;
 
-				if (target != null && target.Inventory.GetEquip(EquipSlot.Gloves) != null)
+				// Algorithm retrieved from client files.
+				var speedBonus = 1 + (skillLevel * 0.3);
+
+				if (caster.Abilities.Has(AbilityId.Enchanter8))
```
exectails, line 33: Is this ability actually used? I don't see it on my Enchanter.

system/scripts/core/calc_ability.cs
```diff
@@ -0,0 +1,65 @@
...
+//--- Description -----------------------------------------------------------
+// Functions that calculate ability-related values, such as properties.
+//---------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Melia.Zone;
+using Melia.Zone.Scripting;
+using Melia.Zone.Skills;
+
+public class AbilityCalculationsScript : GeneralScript
```
exectails, line 12: This has presumably become obsolete with my recent change to add the reinforce ability data to skills and the bonuses to the skill calculations.

doc/bt/ZC_NORMAL.bt
```diff
@@ -878,15 +878,13 @@ switch (type)
...
-        int targetHandle;
         int characterHandle;
-        int packetStringId;
-        float argNum;
-        short argStrLen; // -1
-        if (argStrLen > -1)
-            char argStr[argStrLen]; // defaults to "None"
-        int i1;
-        int i2;
+        int targetHandle;
```
exectails, line 882: You accidentally reverted some of my changes again.

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,76 @@
...
+			{
+				var skillLevel = buff.NumArg1;
+				var caster = buff.Caster as Character;
+
+				// Algorithm retrieved from client files.
+				var speedBonus = 1 + (skillLevel * 0.3);
+
+				if (caster.Components.Get<AbilityComponent>().Has(AbilityId.Enchanter10))
+				{
+					var SCR_Get_SkillFactor = ScriptableFunctions.Skill.Get("SCR_Get_SkillFactor");
```
exectails, line 36: You should prefer using properties over explicitly calling scriptable functions. It's simpler and you can be sure that you actually get the property for this skill and not some partial value.

src/ZoneServer/Buffs/Handlers/Common_Slow.cs
```diff
@@ -12,25 +10,15 @@ namespace Melia.Zone.Buffs.Handlers
...
-
-			Send.ZC_SHOW_EMOTICON(target, "I_emo_slowdown", buff.Duration);
-			Send.ZC_NORMAL.PlayTextEffect(target, caster, "SHOW_BUFF_TEXT", (float)BuffId.Common_Slow, null);
 
-			var reduceMspd = target.Properties.GetFloat(PropertyName.MSPD) * MspdDebuffRate;
-			buff.Vars.SetFloat("Melia.ReduceMspd", reduceMspd);
-
-			target.Properties.Modify(PropertyName.MSPD_BM, -reduceMspd);
-			Send.ZC_MSPD(target);
+			// TODO: Something with Move Speed
```
exectails, line 21: =|

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,67 @@
...
+
+				// Apply penality when the CASTER Max Physical Attack is lower than the TARGET Max Physical Attack
+				// TODO: Find out the exacly value of the penality (We are applying 50%)
+				var casterMaxPAtk = caster.Properties.GetFloat(PropertyName.MAXPATK);
+				var targetMaxPAtk = target.Properties.GetFloat(PropertyName.MAXPATK);
+				var penalityValue = casterMaxPAtk < targetMaxPAtk ? 0.5f : 1f;
+
+				var skillLevel = buff.NumArg1;
+
+				var data = ZoneServer.Instance.Data.SkillDb.Find("Enchanter_EnchantEarth");
```
exectails, line 33: You should use the `SkillId` enum over strings, for type safety and to prevent typos.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,67 @@
...
+
+				var initialBlockPenBonus = data.Factor + (skillLevel * data.FactorByLevel);
```
exectails, line 35: Could you explain the idea behind this formula? The skill description says it increases the block penetration by X%, but what you're doing here is calculating some percentage of the skill factor that you then use as the modifier for `BLK_BREAK_BM`? I'm not sure how all this relates.

src/ZoneServer/Buffs/Handlers/OverReinforce_Buff.cs
```diff
@@ -0,0 +1,72 @@
...
+	/// Handle for the Over Reinforce Buff, which increases the target's physical and magic attack
+	/// </summary>
+	[BuffHandler(BuffId.OverReinforce_Buff)]
+	public class OverReinforce_Buff : BuffHandler
+	{
+		private const string VarName = "Melia.MagicAndPhysicalAttackBonus";
+
+		public override void OnStart(Buff buff)
+		{
+			var target = buff.Target as Character;
```
exectails, line 21: Is there a reason for these `Character` checks? If you remove them we could use this for monsters as well if we wanted to.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -98,7 +98,10 @@ protected IEnumerable Say(string message)
 		/// <exception cref="ArgumentException"></exception>
 		protected IEnumerable Emoticon(string packetString)
 		{
-			Send.ZC_SHOW_EMOTICON(this.Entity, packetString, TimeSpan.FromSeconds(2));
+			if (!ZoneServer.Instance.Data.PacketStringDb.TryFind(packetString, out var data))
```
exectails, line 101: It seems like you reverted quite a few of my changes, so I'll once more postpone looking over the rest of this PR^^" Please check the comments I made and try not to revert recent commits when you merge.

felipecastagnarodecarvalho: I am sorry about it, I am trying to save this PR somehow.

felipecastagnarodecarvalho:
I am gonna close this PR and make a new branch, sorry about it.
I don`t know how it happened exactly.

felipecastagnarodecarvalho:
New branch and PR incoming....

PR #232, opened 2023-07-06
Refactored implementation of mob spawns

kenedos:
This is a refactor and a more reasonable update to the monster spawner class.

- Retrocompatibility with old spawner scripting code (but parameters changed positions)
- Changed MonsterSpawner to be part of WorldManager
- Added minAmount, spawnerName, tendency and propertyOverrides as optional parameters on shortcuts
- Added List of SpawnPoints to Spawner class
- Changed "test" package into "default"
- Spawner and SpawnPoint are now in Melia.Zone.World.Spawner namespace
- Changes as minimal as possible in Spawner class to allow default spawns


kenedos:
<iframe width="560" height="315" src="https://www.youtube.com/embed/tCtcxM8m4oQ" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>

src/ZoneServer/World/Spawner/MonsterSpawner.cs
```diff
@@ -110,35 +110,30 @@ public class MonsterSpawner
...
 		/// <param name="maxAmount"></param>
-		/// <param name="mapClassName"></param>
-		/// <param name="area"></param>
 		/// <param name="initialSpawnDelay"></param>
 		/// <param name="minRespawnDelay"></param>
 		/// <param name="maxRespawnDelay"></param>
 		/// <param name="tendency"></param>
 		/// <param name="propertyOverrides"></param>
-		public MonsterSpawner(int monsterClassId, int minAmount, int maxAmount, string mapClassName, IShape area, TimeSpan initialSpawnDelay, TimeSpan minRespawnDelay, TimeSpan maxRespawnDelay, TendencyType tendency, PropertyOverrides propertyOverrides)
+		public MonsterSpawner(string name, int monsterClassId, int minAmount, int maxAmount, TimeSpan initialSpawnDelay, TimeSpan minRespawnDelay, TimeSpan maxRespawnDelay, TendencyType tendency, PropertyOverrides propertyOverrides)
```
exectails, line 122: Hm... So you're assigning names to the spawners and then assign spawn points to them. That is certainly one way of doing it, but wouldn't it be more flexible to treat the spawn points as a separate thing, a list of areas that can be used for arbitrary purposes? For example, let's assume that the spawners and ambushes use the same areas. With the current approach, we would have to assign the same areas again to the ambushes, but if the spawn points are just a list, stored somewhere in the server, we could access them for ambushes as well, and whatever other things we might come up with. What do you think?

src/ZoneServer/World/Spawner/MonsterSpawner.cs
```diff
@@ -149,15 +144,44 @@ public MonsterSpawner(int monsterClassId, int minAmount, int maxAmount, string m
...
 		public void Spawn(int amount)
 		{
+			// Gets random spawn point
+			var index = _rnd.Next(0, this.SpawnPoints.Count);
+			var spawnPoint = this.SpawnPoints[index];
+
 			for (var i = 0; i < amount; ++i)
 			{
-				if (!this.TryGetRandomPosition(out var pos))
+				if (!this.TryGetRandomPosition(spawnPoint, out var pos))
```
exectails, line 182: Would it perhaps make more sense to move the random position getter to the spawn point class?

src/ZoneServer/Scripting/Shortcuts.cs
```diff
@@ -154,109 +155,104 @@ public static WarpMonster AddWarp(string warpName, double direction, Location fr
...
 		/// <param name="tendency"></param>
 		/// <param name="propertyOverrides"></param>
 		/// <returns></returns>
 		/// <exception cref="ArgumentException"></exception>
-		public static MonsterSpawner AddSpawner(int monsterClassId, int minAmount, int maxAmount, TimeSpan initialSpawnDelay, TimeSpan minRespawnDelay, TimeSpan maxRespawnDelay, string map, IShape area, TendencyType tendency, PropertyOverrides propertyOverrides)
+		public static MonsterSpawner AddSpawner(int monsterClassId, int maxAmount, TimeSpan initialSpawnDelay, TimeSpan minRespawnDelay, TimeSpan maxRespawnDelay, string spawnerName = "", int minAmount = -1, TendencyType tendency = TendencyType.Peaceful, PropertyOverrides propertyOverrides = null)
 		{
-			if (!ZoneServer.Instance.World.TryGetMap(map, out var mapObj))
-				throw new ArgumentException($"Map '{map}' not found.");
+			if (spawnerName.IsNullOrWhiteSpace())
```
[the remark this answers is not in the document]
kenedos: If you just create the spawner but don't add spawn points to it it won't be able to spawn anything, yes.

We opted for this architecture where spawners are not necessarily associated to a map but the spawn points are. We can't have spawners also associated to the map because then we can't do cross-map spawning (required for world bosses).

It still makes sense, however, to create a spawner with an empty name string (which will be internally generated) because that shortcut returns the object and the "Name" property is public, so they can always find what name the spawner was generated with.

src/ZoneServer/World/Spawner/MonsterSpawner.cs
kenedos: I thought a lot about this but I don't see how to implement your suggestion in a neat, organized way.

Let's assume, for example, that the spawn point list is stored as a property of Map class, because that would make the most sense. Then we arrive at the following issues:

- The spawner class is not tied to a map, so it won't know which map to look for when iterating over spawn points.
- The spawner class would then have to have a list of spawn points in its property, making it redundant to have list of spawn points also in map class property (not that I'm against this redundancy, but I think the list in map will just not be used).
- Spawn points are identified by their map and area. There isn't a name string to associate them to (like the spawner class). So re-using them on scripts as you intend is quite complicated. I also wouldn't want to add name strings to spawn points, it would be too unecessarily confusing. Perhaps on spawn point collections, but that's not under the scope of this PR.
- I imagine the spawn point class as a "component" of spawner or (the future) ambush spawner classes. It should not exist on it's own, as it makes no sense to have a spawn point in a map that does nothing.
- For future classes that would use same spawn point (i.e. ambush spawner class) we could just have a shortcut that adds the spawn point to the 'spawner list' and 'ambush spawner list'. No real need to add spawn point to the world or map and then an additional effort to identify and retrieve it to be re-used.

In short, I believe having each spawner class (spawner & ambush spawner) having their own internal list of spawn points as a better architecture. But what do you think?

kenedos:
Note: Added tendency of monsters on the spawners, which was missing on your last review of the scripts.

src/ZoneServer/World/Spawner/MonsterSpawner.cs
exectails: I agree that each spawner should have its own list, particularly because I also imagine that it could be useful to be able to customize a pre-defined list. Though I imagine it wouldn't be very complicated. The way I picture it, the spawn points would be placed on World because they don't necessarily have a relation to any one map. It'd basically be like a database that we generate at run-time and then reference to build a spawner's list of potential spawn areas.
```cs
World:
    SpawnAreaCollection SpawnAreas
    SpawnerCollection Spawners

Shortcuts:
    AddSpawnArea(string spawnerAreaListName, string mapClassName, ...)
        World.SpawnAreas.GetOrCreateList(spawnerAreaListName).Add(...)

    SpawnArea[] SpawnAreas(string spawnAreaListName)
        return World.SpawnAreas.Get(spawnAreaListName)

    AddSpawner(int monsterClassId, ... string mapClassName, IShapeF area, ...))
        AddSpawner(monsterClassId, ... new[] { new SpawnArea(mapClassName, area) }, ...)

    AddSpawner(int monsterClassId, ... SpawnArea[] spawnAreas, ...))
        World.Spawners.Add(...)

// Usage Examples
AddSpawnArea("FSiauliaiWestOnions", "f_siauliai_west", ...)
AddSpawnArea("FSiauliaiWestOnions", "f_siauliai_west", ...)

AddSpawner(MonsterId.Onion, ... "f_siauliai_west", Area(...), ...)
AddSpawner(MonsterId.Onion, ... SpawnAreas("FSiauliaiWestOnions"), ...)
```
An alternative to using named lists would be to create local lists that can be directly passed to `AddSpawner`.
```cs
var onionAreas = CreateSpawnerAreas()
onionAreas.Add("f_siauliai_west", ...)

AddSpawner(MonsterId.Onion, ... onionAreas, ...)
```
The disadvantage is that we wouldn't be able to easily reuse spawn area lists from other scripts, but I guess that would be rare. Well, unless you wanted to add a custom spawner that reuses an existing list 🤔 For example, I might want to add a regular spawn for a Large Kepa, as a kind of mini boss, and have it spawn on one of the Kepa spawns in West Siauliai. If it's not possible to reference the existing list, I'd have to duplicate it. Granted, that's not an uncommon thing to do in such a case, but the list _could_ be made accessible.

> Spawn points are identified by their map and area

Each independent spawn point doesn't need an identifier I don't think, but yea, the _list_ of spawn points would require one if we wanted to be able to query it as needed for whatever purposes, be it as a reference for the spawn areas or for whatever other purposes someone might come up with.

> There isn't a name string to associate them to (like the spawner class)

Right, but we don't actually _need_ to name the spawners, do we? I can't think of a general reason to do that off the top of my head. It's just the implementation that made sense if you want to assign spawn points to the spawners instead of the other way around. Not saying this doesn't work, but you could just as well name spawn area lists and remove the spawner names.

> as it makes no sense to have a spawn point in a map that does nothing

I hope my comparison to a database makes sense here. It's somewhat like the resurrection point database, just a list of named lists of map/area values. We could technically also actually put that data into a database, like the game does, though I think we want to create it at run-time for customization purposes.

> For future classes that would use same spawn point (i.e. ambush spawner class) we could just have a shortcut that adds the spawn point to the 'spawner list' and 'ambush spawner list'. 

Would you then give the spawners and the ambushes the same name? You'd have to assign the same lists to the same spawners and ambushes after all. Doesn't that seem confusing? And another potential problem I just remembered mentioning before, what if you want to use the same spawn areas for multiple monsters?
```cs
AddSpawner("FSiauliaiWestOnions", MonsterId.Onion, ...)
AddSpawner("FSiauliaiWestOnions", MonsterId.Leaf_diving, ...)

AddAmbush("FSiauliaiWestOnions", MonsterId.Onion, ...)

AddSpawnPoint("FSiauliaiWestOnions", "f_siauliai_west", ...)
AddSpawnPoint("FSiauliaiWestOnions", "f_siauliai_west", ...)
AddSpawnPoint("FSiauliaiWestOnions", "f_siauliai_west", ...)
AddSpawnPoint("FSiauliaiWestOnions", "f_siauliai_west", ...)
```
I mean, that _will_ work if you iterate over all possible options inside `AddSpawnPoint`, but doesn't it feel a little weird for names not to be unique?

kenedos: I see what you're saying. I will add a 'MonsterSpawnPointCollection' class that has a list of 'MonsterSpawnPoint' in it and an identifier string. It makes more sense to have the identifier stirng in the spawn point collection than on the spawner. That way, you can associate multiple spawners to a spawn point collection and spawn different mob classes or add ambushes.

'MonsterSpawnPointCollection' will also be stored in world, as you recommended, because it can contain spawn points on different maps.

I was trying to get away with this PR without implementing spawn point collections, but I think you're right, that is the best architecture.

kenedos: I have added the functionality you suggested. Now you can assign multiple spawners to a same spawn point collection:
image

And you can also do it even if you add the spawn point after:
image

Essentially, the spawn point collection has a string name identifier and the spawner class receives that string and tries to resolve the object on runtime. The name identifier is stored at world so you can even use it cross-script. I also removed name from spawner class, as you recommended (now only the spawn point collection has a name, but the users making scripts don't need to know how it works internally).

kenedos: Maybe I should rename the "spawnPointCollectionName" variables to something more intuitive for users that produce scripts? I mean, for them, they don't even need to know spawn point collections exist. Maybe I should name it "spawnIdentifier" or something generic and abstract? What do you think?

exectails: Sounds good!

exectails: One thing I'm not sure about is the order of the `AddSpawnPoint` arguments. Don't you think putting the name at the end makes taking in the information more difficult? Isn't the name of the list the most important information in a call that adds a new element to said list? Say you're staring at a block of spawn point definitions, it becomes difficult to determine whether they even all go into the same list.
```cs
AddSpawnPoint("f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205), "SpawnPointCollection10.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Spot(-575, -1300), "SpawnPointCollection10.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205, -555, -1205), "SpawnPointCollection11.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Spot(-575, -1300), "SpawnPointCollection10.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Rect(-555, -1105, 500), "SpawnPointCollection10.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205), "SpawnPointCollection10.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205, -555, -1205), "SpawnPointCollection11.f_siaulia_west");
AddSpawnPoint("f_siauliai_west", Spot(-575, -1300), "SpawnPointCollection10.f_siaulia_west");
```
The only information you can glean at a glance is that all of those are in West Siauliai, but not which lists they belong to. On the other hand, putting the name as the first argument lets you parse both the lists they belong to and the maps at once.
```cs
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Spot(-575, -1300));
AddSpawnPoint("SpawnPointCollection11.f_siaulia_west", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205, -555, -1205));
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Spot(-575, -1300));
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Rect(-555, -1105, 500));
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205));
AddSpawnPoint("SpawnPointCollection11.f_siaulia_west", "f_siauliai_west", Area(-555, -1105, -655, -1105, -655, -1205, -555, -1205, -555, -1205));
AddSpawnPoint("SpawnPointCollection10.f_siaulia_west", "f_siauliai_west", Spot(-575, -1300));
```

kenedos: I thought of that too, and already changed it, just reviewing some spawn points now because some NPCs are being spawned because client data has them with "Monster" type ^^

image

kenedos: I've finished my updates and removed some spawn points that were generating hidden triggers and npcs. Let me know if there's any other changes to be done in this PR.

exectails:
Finally merged^^ Thanks for the PR. Though I did notice a couple things after merging. There are some minor formatting/style "errors", such as the spaces in the `AddSpawnPoint` calls. Example:
```cs
AddSpawnPoint( "d_abbey_22_4.Id1", ...
```
Technically that's not an issue, but if someone with an auto-formatter were to go in there, make a change, and save that file, they'd end up with a bunch of unnecessary changes, which is unfortunate. It also feels to me like it would be nice to have the actual map names in the descriptions in the headers, and unnecessary empty lines could potentially be removed as well, such as after the spawn points.
```cs
		AddSpawnPoint( "f_siauliai_west.Id9", "f_siauliai_west", Spot(1464.408, -694.63, 20));
		AddSpawnPoint( "f_siauliai_west.Id9", "f_siauliai_west", Spot(592.6201, -804.8381, 20));
		// <<< empty
	}
}
```
I also noticed that quite a few of your coordinates (X/Z) don't match the values I see in the anchor files. They are basically rounding errors, so it's again not a technical issue, and floating points can be finicky, but I still wonder what happened there, because this, too, could cause a lot of unnecessary changes. Rounding them to perhaps 2 digits after the decimal point should fix this without causing issues and would also make the calls a smidge more concise.

Another potential issue is the order of the spawns. I believe the creation was tool-assisted, which is of course fine, but the order does not appear to match the order of the gentypes found in the client. Looking at your code I couldn't discern a clear pattern for how the data was sorted/read, which might make eventual updates a little tricky. If someone else were to make their own tool and recreate these scripts, they would most likely cause changes in every single file just because the order is different, which will make tracking the actual changes difficult. One simple solution would be to stick to the gentype order, another even more reliable one would be sorting by monster id or similar. Something reproducible that ideally also groups together the monsters.

Aside from these styling issues I also noticed that you appear to be missing a couple things, such as property overrides on some maps. The map "ep13_2_d_prison_1" for example doesn't have any overrides in your scripts, but it does have a fieldstatus file, though it's incorrectly named "field_monster_status_ep13_ep13_2_d_prison_1". I assume you used the file names instead of the namespaces to match these up. But there are also overrides missing that don't have a corresponding spawn, presumably because you only looked up the overrides for the monsters that are actually there. That might not appear like an issue at first glance, but Large Kepa is missing in West Siauliai for example, so if it's spawned by an event or quest script or something it would have its default stats. And while we could put the overrides there, we'd end up with duplicated data if we have more than one of those.

This should be it though =P

PR #233, opened 2023-07-07
Adding Wugushi Skills

felipecastagnarodecarvalho:
Wugushi Skills:
- Wugong Gu
- Zhendu
- Latent Venom
- Golden Frog
- WideMiasma

### DIFF
--- a/src/Shared/Network/NormalOp.cs
+++ b/src/Shared/Network/NormalOp.cs
@@ -64,7 +64,8 @@
 			public const int WigVisibilityUpdate = 0x1AC;
 			public const int Unknown_1B4 = 0x1B7;
 			public const int SubWeaponVisibilityUpdate = 0x1C5;
-			public const int ShowAnimationEffect = 0xE3;
+			public const int Skill_E3 = 0xE3;
+			public const int Skill_06 = 0x06;
 		}
 	}
 }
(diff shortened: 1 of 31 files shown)


felipecastagnarodecarvalho:
Missing:
 - Golden Frog skill effect not being displayed
  - Throw Pot skill effect not being displayed

exectails:
Thank you for the PR, but I see you're listing missing effects. Before I review the actual changes, what's the problem with the effects? I'd prefer not to merge half-implemented skills, since we'd run the risk of forgetting about finishing them.

felipecastagnarodecarvalho:
Yeah the idea is to fully display the effects before you try to merge it.
So it was being displayed before I updated the client.
And for some reason now it doesn't show up.
I am gonna try again tho.

exectails:
Updates often times change the packet strings and their ids, which also affects the effects that use them. Could that be the issue?

felipecastagnarodecarvalho:
I could be it! How can we test it out?

exectails:
Update your client and see if that fixes it^^

felipecastagnarodecarvalho:
I did but did not fix it properly.
Do you think the packetstrings.txt file could be outdated/old ?

felipecastagnarodecarvalho:
this was fixed by using exclusive the steam client for tests

felipecastagnarodecarvalho:
Current issues/bugs:
 - Can't display two effects at the same time (based on the skill_59 function) - If you use PoisonPot the GoldenFrog effect will not be showed.

felipecastagnarodecarvalho:
Closing since we are making another PR with this content

PR #236, opened 2023-07-08
Add Skills (Enchanter, Ardito, Wugushi, Sapper)

felipecastagnarodecarvalho:
TODO/Missing features:
 - Apply buffs on party members and pet as well
 - Add Enchant Weapon Skill (the effect is applied once the player learns the skill)
 - All Buff handler and Skill handlers can be improved as we learn more and add more properties/features. The code is not definitive and there is always open doors for improvements.


felipecastagnarodecarvalho:
TODO:
 - Make the skill rotate within the mouse position properly (how?)
  - Add knockback to Taglio Skill (I don`t know why inst working right now)

felipecastagnarodecarvalho:
List of skills added:

Ardito:
- Granata
- Invasione
- Recupero
- Ritirarsi
- Taglio
- TreGranate

Enchanter:
- Agility
- EnchantAura
- EnchantEarth
- EnchantGlove
- OverReinforce

Sapper:
- Claymore
- LegHoldTrap
- PunjiStake
- SpikeShooter
- SpringTrap

Wugushi:
- CrescendoBane
- GoldenFrog
- LatentVenom
- PoisonPot
- WideMiasma
- WugongGu
- Zhendu

src/ZoneServer/Skills/Handlers/Sapper/Claymore.cs
```diff
@@ -0,0 +1,162 @@
...
+				return;
+			}
+
+			if (!caster.Position.InRange2D(farPos, skill.Data.MaxRange))
+			{
+				caster.ServerMessage(Localization.Get("Too far away."));
+				return;
+			}
+
+			if (trap != null)
```
exectails, line 50: This is another case that doesn't work because handlers aren't created for every skill use. I should note that I don't mention every instance of every issue I find, so this one for example exists in most of the new handlers.

src/ZoneServer/Skills/Handlers/Sapper/Claymore.cs
```diff
@@ -0,0 +1,162 @@
...
+
+			Send.ZC_SKILL_READY(caster, null, caster.Position, caster.Position);
+			Send.ZC_NORMAL.UpdateSkillEffect(caster, 0, caster.Position, trap.Direction, trap.Position);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, trap.Position, ForceId.GetNew(), null);
+
+			await Task.Delay(350);
+
+			Send.ZC_DEAD(trap, trap.Position);
+
+			caster.Map.RemoveMonster(trap);
```
exectails, line 84: The trap wouldn't be removed if the caster changed maps or DCed in between casting and this happening.

src/ZoneServer/Skills/Handlers/Sapper/DetonateTraps.cs
```diff
@@ -0,0 +1,89 @@
...
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, ForceId.GetNew(), null);
+
+			var character = caster as Character;
+			var removeList = new List<Mob>();
+
+			foreach (var trap in character.PlacedTraps)
+			{
+				Send.ZC_DEAD(trap, trap.Position);
+				caster.Map.RemoveMonster(trap);
+				if (character.PlacedTraps.Contains(trap))
```
exectails, line 55: Isn't this check pointless, since you're iterating over the traps? It has to be in the list.

src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -260,6 +261,11 @@ public int ClassLevel
 		/// </summary>
 		public bool IsDead => (this.Hp == 0);
 
+		/// <summary>
+		/// Returns the character's list of placed traps
+		/// </summary>
+		public List<Mob> PlacedTraps { get; set; } = new List<Mob>();
```
exectails, line 267: You should never expose a non-thread-safe list in a multi-threaded environment. If anybody decides to suddenly modify this list from somewhere, or it just happens accidentally, you'll get exceptions. And since you're using the list in async code that doesn't have try/catches, your server will crash if that happens.

src/ZoneServer/Skills/Handlers/Sapper/LegHoldTrap.cs
```diff
@@ -0,0 +1,212 @@
...
+
+			while (!cancellationToken.IsCancellationRequested)
+			{
+				var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+				foreach (var target in targets.LimitBySDR(caster, skill))
+				{
+					// Oficial server skills apply slow while the mob is inside the area
+					// But for now we gonna apply 10 seconds slow
+					if (!target.Components.Get<BuffComponent>().Has(BuffId.Common_Slow))
```
exectails, line 188: Seems like this would be a good use case for trigger areas, no?

src/ZoneServer/Skills/Handlers/Sapper/LegHoldTrap.cs
```diff
@@ -0,0 +1,212 @@
...
+					{
+						var duration = TimeSpan.FromSeconds(10);
+						target.StartBuff(BuffId.Common_Slow, skill.Level, 0, duration, caster);
+					}
+
+					if (!target.Components.Get<BuffComponent>().Has(BuffId.LegHoldTrap_Debuff))
```
exectails, line 194: Is this check necessary? The buff is defined with an overbuff of 1, so it shouldn't get started more than once either way.

src/ZoneServer/Skills/Handlers/Sapper/PunjiStake.cs
```diff
@@ -0,0 +1,279 @@
...
+			// Duration changed to 5 seconds{ nl}
+			// *Movement speed increase effect removed{ nl}
+			// *Cooldown increased by 10 seconds
+			if (!caster.Components.Get<AbilityComponent>().IsActive(AbilityId.Arditi19))
+			{
+				caster.Properties.Modify(PropertyName.MSPD_BM, 10);
+				Send.ZC_MSPD(caster);
+			}
+
+			stopwatch = new Stopwatch();
```
exectails, line 59: There's a nice shortcut for this, Stopwatch.StartNew().

src/ZoneServer/Skills/Handlers/Sapper/PunjiStake.cs
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		/// <summary>
+		/// Apply Slow to the targets and restore it after certain time
+		/// </summary>
+		/// <param name="target"></param>
+		/// <param name="oldValue"></param>
+		private async void ApplySlow(ICombatEntity target, float oldValue)
+		{
+			target.Properties.SetFloat(PropertyName.MSPD, 10);
```
exectails, line 274: You can't change MSPD directly because it's a calculated property. But even if you could change it, that would be a pretty bad idea to do from a skill like this, because if the target were to log out in between setting and resetting the value, they'd get stuck with the lower MSPD when it's saved to the db.

src/ZoneServer/Skills/Handlers/Sapper/SpikeShooter.cs
```diff
@@ -0,0 +1,211 @@
...
+						target.TakeDamage(skillHitResult.Damage, caster);
+
+						var hit = new HitInfo(caster, target, skill, skillHitResult);
+
+						Send.ZC_HIT_INFO(caster, target, skill, hit);
+					}
+
+					Send.ZC_NORMAL.PlayForceEffect(caster, "I_arrow009", ForceId.GetNew(), "FAST", 500, leftPos, leftMovPos);
+					Send.ZC_NORMAL.PlayForceEffect(caster, "I_arrow009", ForceId.GetNew(), "FAST", 500, rightPos, rightPos);
+				} else //Stops on the first hit
```
exectails, line 147: This else should have it's own line, as per our coding conventions.

src/ZoneServer/Skills/Handlers/Wugushi/Zhendu.cs
```diff
@@ -0,0 +1,72 @@
...
+
+			var effectId = ForceId.GetNew();
+
+			Send.ZC_NORMAL.GroundEffect_59(caster as Character, "M_GTOWER_STAGE_15", skill.Id, caster.Position, effectId, true);
+			Send.ZC_NORMAL.UpdateSkillEffect(caster, caster.Handle, caster.Position, caster.Position.GetDirection(caster.Position), Position.Zero);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
+
+			var casterCharacter = caster as Character;
+
+			// Reduces the Poison property resistance of enemies within 150 by 10% per attribute level when [Zhendu] is used
```
exectails, line 49: I assume this is unfinished?

src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -705,6 +711,12 @@ public void Heal(float hpAmount, float spAmount)
 			if (hpAmount == 0 && spAmount == 0)
 				return;
 
+			// 30% of healing reduced
```
exectails, line 714: Putting this in the Heal method is potentially problematic. It's convenient that it works automatically on everything that heals, but at the same time you might call this method expecting to get exactly what you put in, and then be confronted with unexpected side effects. I believe we would need both, and at the _very_ least it should be documented.

system/scripts/core/calc_skill.cs
```diff
@@ -67,15 +67,15 @@ public float SCR_Get_SkillFactor(Skill skill)
 	public float SCR_Get_AbilityReinforceRate(Skill skill)
 	{
 		if (skill.Data.ReinforceAbility == 0)
-			return 0;
+			return 1;
```
exectails, line 70: This change is incorrect. The result is used in this formula:
```cs
var value = sklFactor + (sklFactorByLevel * (skill.Level - 1));
// ...
value += value * byReinforceRate;
```
Meaning that if you return 1, the value would always double, because it's added to itself. We only want to return and add the bonus.

exectails (changes requested):

src/ZoneServer/Buffs/Buff.cs
```diff
@@ -36,6 +37,11 @@ public class Buff : IUpdateable
 		/// </summary>
 		public SkillId SkillId { get; } = SkillId.Normal_Attack;
 
+		/// <summary>
+		/// Returns the buff's associated skill.
+		/// </summary>
+		public Skill Skill { get; }
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Wow that is a big issue!

src/ZoneServer/Skills/Handlers/Wugushi/Zhendu.cs
felipecastagnarodecarvalho: Done!

src/Shared/Network/NormalOp.cs
```diff
@@ -39,9 +39,28 @@ public static class Zone
 			public const int AttackCancel = 0x41;
 			public const int Skill_45 = 0x45;
 			public const int AccountUpdate = 0x4C;
+			public const int Skill_4E = 0x4E;
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I have removed the particular OP but there are a few other that are unknow purposes for me and I will leave then without a real name for now until we can finally figurate out.

src/ZoneServer/World/Actors/Characters/Character.cs
felipecastagnarodecarvalho: which side effects are you thinking on?

src/ZoneServer/Skills/Handlers/Sapper/PunjiStake.cs
felipecastagnarodecarvalho: Worked around it: Now it applies 70% move speed reduction.

src/ZoneServer/Buffs/Buff.cs
```diff
@@ -173,6 +179,49 @@ public Buff(BuffId buffId, float numArg1, float numArg2, TimeSpan duration, ICom
...
+		/// </summary>
+		/// <param name="caster"></param>
+		/// <param name="buffId"></param>
+		/// <param name="numArg1"></param>
+		/// <param name="numArg2"></param>
+		/// <param name="duration">Use MinValue to use the buff's default duration.</param>
+		/// <param name="target"></param>
+		/// <param name="caster"></param>
+		/// <param name="skill">Id of the skill associated with this buff.</param>
+		public Buff(BuffId buffId, float numArg1, float numArg2, TimeSpan duration, ICombatEntity target, ICombatEntity caster, Skill skill)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: removed.

src/ZoneServer/Buffs/Buff.cs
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Resolved.

system/scripts/core/calc_skill.cs
felipecastagnarodecarvalho: I was looking into the SCR_REINFORCEABILITY_TOOLTIP from the client, I guess I was wrong then...

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,76 @@
...
+
+				// Algorithm retrieved from client files.
+				var speedBonus = 1 + (skillLevel * 0.3);
+
+				if (caster.Components.Get<AbilityComponent>().Has(AbilityId.Enchanter10))
+				{
+					var SCR_Get_SkillFactor = ScriptableFunctions.Skill.Get("SCR_Get_SkillFactor");
+					if (caster.Skills.TryGet(buff.SkillId, out var skill))
+					{
+						var abilityBonus = SCR_Get_SkillFactor(skill);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: got it, although I think using those NumberArgs are very dirty....

src/ZoneServer/Buffs/Handlers/DecreaseHeal_Debuff.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Melia.Zone.Buffs.Handlers
+{
+	internal class DecreaseHeal_Debuff
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: yep. I though it needed to be declared to appear as a buff but that`s not the case.

src/ZoneServer/Buffs/Handlers/LegHoldTrap_Debuff.cs
```diff
@@ -0,0 +1,59 @@
...
+		public override void OnStart(Buff buff)
+		{
+			var target = buff.Target;
+
+			// We may need to add "CanMove" variable to this component.
+			if (!buff.Target.Components.TryGet<AiComponent>(out var component))
+			{
+				var movementComponent = target.Components.Get<MovementComponent>();
+				movementComponent?.Stop();
+				// movementComponent.MoveSpeedType = MoveSpeedType.Stop;
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: yeah! that's a interesting approach.

src/ZoneServer/Buffs/Handlers/LegHoldTrap_Debuff.cs
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			// Well I don't know a better way of doing this
+			// Lets just decrease the movespeed to zero?
+			if (target as Character != null)
+			{
+				var reduceMspd = target.Properties.GetFloat(PropertyName.MSPD);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: it can't be zero so I leave it as 0.0001f

src/ZoneServer/Buffs/Handlers/Zhendu_Buff.cs
```diff
@@ -0,0 +1,63 @@
...
+			{
+				buff.Target.Properties.Modify(PropertyName.SkillFactor, dmgIncreasePercentage);
+			}
+		}
+
+		public override void OnEnd(Buff buff)
+		{
+			if (buff.Vars.TryGetFloat(varName, out var bonus))
+			{
+				if (buff.Target.Properties.Has(PropertyName.SkillFactor))
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: So yeah this is currently incomplete/unfinished.
The ideia is that should increase all SKILLS from WUGUSHI (and not all skills as it is now) by a percentage

src/ZoneServer/Network/PacketHandler.cs
```diff
@@ -1301,20 +1302,45 @@ public void CZ_SKILL_SELF(IZoneConnection conn, Packet packet)
...
+				Log.Warning("CZ_DYNAMIC_CASTING_START: User '{0}' tried to use a skill that's on cooldown ({1}).", conn.Account.Name, skillId);
+				character.ServerMessage(Localization.Get("You may not use this yet."));
 				return;
+			}
 
-			handler.StartDynamicCast(skill, character);
+			// Try to use skill
+			try
+			{
+				if (!ZoneServer.Instance.SkillHandlers.TryGetHandler<IDynamicCastingSkillHandler>(skillId, out var handler))
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: What happened is that I was implementing it and you did as I was doing so.
So we get a conflict.

src/ZoneServer/Network/Send.cs
```diff
@@ -463,7 +463,7 @@ public static void ZC_SKILL_MELEE_GROUND(ICombatEntity entity, Skill skill, Posi
 			packet.PutInt(1);
 			packet.PutFloat(shootTime);
 			packet.PutFloat(1);
-			packet.PutInt(0);
+			packet.PutInt(1);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I think it was a mistake.
I was desperate  to find the effect to show.

felipecastagnarodecarvalho:
Sapper:
- Can't show the explosion effect of the Claymore skill.
I spend a lot of time on this but I can't identify why its not being displayed.
- Can't show the pillar (effect) for the Spike Shooter skill.
Other effect that I am having issue to display somehow.
- Desync between the actual hitbox and skill effect for the skill Broom Trap
It seems like the skill effect doesn't have a linear movement and it causes a desync.
- Detonate Trap doesn't show the explosion effect.

Wugushi
- Zhendu Buff effect is not present.
- Crescendo Bane Buff effect is not present.

Ardito:
- Everything seems to work fine with all the effects included.

Enchanter:
- Enchant Light Skill implementation missing (It should apply forever once the skill is learned)

src/ZoneServer/World/Actors/Characters/Character.cs
exectails: The side effect I was referring to is that you're changing the amount of healing received. As a hypothetical example, let's say I wanted to create an NPC that heals the character for exactly 50% of their HP and SP. The way I would try to do that might be to get the max HP and SP, half them, and then call `Heal` with those values. This addition of yours would change the amount of HP the character is healed by, and the users who called the function might rightfully wonder why they didn't get the full amount. And since you didn't document this behavior in any way, i.e. didn't change the comments on the `Heal` method, the only way for them to figure out what happened is to dig into the source code, which is never ideal.

exectails:
What's the current status of this PR? I see there's a couple conflicts that need to be resolved, and based on your last comment there appear to be several issues still remaining. While a missing effect here and there might be acceptable for now, I'm somewhat concerned about the desyncs mentioned.

felipecastagnarodecarvalho:
I am updating this PR right now. I should be done by the end of this week (15/06/2024).
I see a lot of things that annoying me a bit. I will try to improve it as much as I can.

src/ZoneServer/Skills/Handlers/Enchanter/EnchantAura.cs
```diff
@@ -0,0 +1,169 @@
...
+			}
+
+			await Task.Delay(TimeSpan.FromSeconds(1));
+
+			while (buffComponent.Has(BuffId.EnchantAura_Buff))
+			{
+				await Task.Delay(2000);
+
+				// Cancel if the caster has not enough SP
+				if (!caster.TrySpendSp(54f))
```
Terotrous, line 109: According to the ingame tooltip, the upkeep cost is 21, not 54.  If this changes based on the skill's level, you might want to make this a variable.

src/ZoneServer/Skills/Handlers/Enchanter/EnchantAura.cs
```diff
@@ -0,0 +1,169 @@
...
+				{
+					this.RemoveSkillEffect(skill, caster, position, effectId);
+					break;
+				}
+
+				var character = caster as Character;
+
+				if ((character != null && !character.Connection.LoggedIn) || caster.IsDead)
+				{
+					break;
```
Terotrous, line 119: It also needs to remove the skill effect here.  You might want to move the RemoveSkillEffect from the previous bracket to the end of the loop, as in all of the situations where the execution escapes the loop (these two brackets and the while condition failing) the effect should be removed.

src/ZoneServer/Skills/Handlers/Wugushi/WideMiasma.cs
```diff
@@ -0,0 +1,82 @@
...
+
+			var hits = new List<SkillHitInfo>();
+
+			foreach (var target in targets.LimitRandom(10))
+			{
+				var skillHitResult = SCR_SkillHit(caster, target, skill);
+
+				Send.ZC_SHOW_EMOTICON(target, "shootpad_spector", TimeSpan.FromMilliseconds(2000));
+				target.StartBuff(BuffId.WideMiasma_Debuff, skill.Level, skillHitResult.Damage, TimeSpan.FromSeconds(15), caster);
+				target.StartBuff(BuffId.DecreaseHeal_Debuff, skill.Level, skillHitResult.Damage, TimeSpan.FromSeconds(20), caster);
```
Terotrous, line 78: This is not what the numarg2 for this buff should be, it needs to be set to 1000x the reduction amount.  So for example, if you need to reduce healing by 3%, you pass in 3000f.  Note that for this skill it scales based on the skill level, so you'll also need to calculate that.  See Doppel Punish for an example of this debuff.

src/ZoneServer/Skills/Handlers/Wugushi/WideMiasma.cs
```diff
@@ -0,0 +1,82 @@
...
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+			caster.StartBuff(BuffId.WideMiasma_Buff, skill.Level, 0, TimeSpan.FromSeconds(10), caster);
+
+			Send.ZC_SKILL_READY(caster, skill, caster.Position, caster.Position);
+			Send.ZC_NORMAL.UpdateSkillEffect(caster, caster.Handle, caster.Position, caster.Position.GetDirection(caster.Position), Position.Zero);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
+
+			// Radius seems precise
+			var radius = 150;
```
Terotrous, line 47: It looks like the radius is 120 as per the skill_bytool.  It also looks like it's cut by half if it's in a PVP area.

src/ZoneServer/Skills/Handlers/Wugushi/WideMiasma.cs
```diff
@@ -0,0 +1,82 @@
...
+		/// Executes the actual attack after a delay.
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="splashArea"></param>
+		private async void Attack(Skill skill, ICombatEntity caster, ISplashArea splashArea)
+		{
+			await Task.Delay(TimeSpan.FromSeconds(1));
+
+			Debug.ShowShape(caster.Map, splashArea, edgePoints: false);
```
Terotrous, line 65: Need to remove debug statements.

src/ZoneServer/Skills/Handlers/Wugushi/Zhendu.cs
```diff
@@ -0,0 +1,65 @@
...
+			{
+				var radius = 150;
+				var center = caster.Position;
+				var splashArea = new Circle(center, radius);
+
+				var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+				foreach (var target in targets)
+				{
+					var positonRes = target.Properties.GetFloat(PropertyName.Poison_Def);
```
Terotrous, line 59: Spelling error here.  Also, this should probably be a buff, as among other things, it doesn't look like you ever remove it.

src/ZoneServer/Skills/Handlers/Ardito/Invasione.cs
```diff
@@ -0,0 +1,99 @@
...
+		/// <param name="splashArea"></param>
+		private async void Attack(Skill skill, ICombatEntity caster, ISplashArea splashArea)
+		{
+			var hitDelay = TimeSpan.FromMilliseconds(200);
+
+			await Task.Delay(hitDelay);
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
```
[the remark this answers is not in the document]
exectails: > Once again I think you're trying to do a multihit, this should just be one hitbox with a hitcount applied to it. Also, this skill has some logic related to a Gunpowder buff, which is consumed to hit extra targets, which I don't see here.

I could be mistaken, but based on the data this skill appears to hit 4 times, with 1 hit each and 250ms between hits. I don't think you'd implement that with multi hits. The actual problem is the missing delays.

src/ZoneServer/Skills/Handlers/Ardito/Invasione.cs
```diff
@@ -0,0 +1,95 @@
...
+			if (!caster.TrySpendSp(skill))
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			var splashParam = skill.GetSplashParameters(caster, originPos, farPos, length: 70, width: 30, angle: 60);
```
exectails, line 41: Since you're using the square splash type, the angle value won't actually do anything. The length value also doesn't quite match the values found in the skill_bytool data (`skill_bytool.ipf/arditi.xml`), which lists 55 as the length. Could that value be more accurate?

src/ZoneServer/Skills/Handlers/Ardito/Recupero.cs
```diff
@@ -0,0 +1,100 @@
...
+			// RECOVERY STAMINA
+			var character = caster as Character;
+			if (character != null)
+			{
+				var maxSta = target.Properties.GetFloat(PropertyName.MaxSta);
+				var addStamina = maxSta * this.GetStaminaRatio(skill);
+				var stamina = character.Properties.Stamina;
+				var addSta = Math.Min(Math.Max(0, maxSta), addStamina - maxSta);
+				character.ModifyStamina((int)addSta);
+			}
```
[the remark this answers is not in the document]
exectails: > This also has the ability to remove a debuff, should be noted with a TODO.

I don't see this mentioned anywhere, where did you read it?

src/ZoneServer/Skills/Handlers/Ardito/Recupero.cs
```diff
@@ -0,0 +1,99 @@
...
+			target = caster;
+
+			Send.ZC_SKILL_READY(caster, skill, originPos, farPos);
+			Send.ZC_NORMAL.UpdateSkillEffect(caster, 0, originPos, caster.Position.GetDirection(farPos), Position.Zero);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, farPos, ForceId.GetNew(), null);
+
+			// RECOVERY HP
+			var maxHP = target.Properties.GetFloat(PropertyName.MHP);
+			var healAmount = maxHP * this.GetMaxHPRatio(caster, skill);
+			var addHp = Math.Min(Math.Max(0, maxHP * 0.5f), healAmount - maxHP);
```
exectails, line 48: We have a clamp method, doing min and max at once, in `Math2.Clamp`.

src/ZoneServer/Skills/Handlers/Ardito/Ritirarsi.cs
```diff
@@ -0,0 +1,120 @@
...
+			Send.ZC_NORMAL.LeapJump(caster, targetPos, 0, 0, 0.5f, 0.35f, 0.7f);
+			
+			var splashParam = skill.GetSplashParameters(caster, originPos, farPos, length: 50, width: 50, angle: 0);
+			var splashArea = skill.GetSplashArea(SplashType.Circle, splashParam);
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				var skillHitResult = SCR_SkillHit(caster, target, skill);
```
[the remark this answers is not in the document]
exectails: > Once again, the game specifies 6 hits for this, which are probably all in one hit packet.

You should not just assume that all skills use the multi hit functionality. It is a fact that ToS uses manual multi hits a lot.

src/ZoneServer/Skills/Handlers/Ardito/Taglio.cs
```diff
@@ -0,0 +1,179 @@
...
+
+			if (buffComponent != null)
+			{
+				if (!buffComponent.Has(BuffId.Taglio_Buff))
+				{
+					buffComponent.Start(BuffId.Taglio_Buff, TimeSpan.FromSeconds(maxCastingTime));
+				}
+			}
+
+			this._effectId = ForceId.GetNew();
```
exectails, line 112: You can't save skill values in instance fields, because the skill handlers are globally used classes. They aren't specifically created for each character, so you would be utilizing this exact variable for every single time this skill is casted by any entity, which could also happen multiple times at once.

src/ZoneServer/Skills/Handlers/Ardito/Taglio.cs
```diff
@@ -0,0 +1,179 @@
...
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="maxCastingTime"></param>
+		/// <param name="cancellationToken"></param>
+		private async void AreaOfEffect(Skill skill, ICombatEntity caster, float maxCastingTime, CancellationToken cancellationToken)
+		{
+			await Task.Delay(150);
+
+			var buffComponent = caster.Components.Get<BuffComponent>();
```
exectails, line 102: The component collection has a `TryGet` method as well that you might want to utilize, as it lets you ditch the null check.

exectails:
I just spent another 2 hours reading through only half of this PR, and the number of little issues and the amount of changes make the repeated reviews rather taxing and time consuming. For this reason, I'd like you to go over everything again, fix the mentioned issues across all of the code, and then try to cut the PR up a little into smaller chunks, maybe even down to one per skill if feasible.

I appreciate your work, and I apologize for the inconvenience, but as is, I can't even leave a comment without my browser lagging because of the size of the changes page^^"

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
```diff
@@ -0,0 +1,96 @@
...
+			// End the Buff if the Caster is Dead
+			if (buff.Caster.IsDead)
+			{
+				buff.End();
+			}
+
+			// Ends the Buff if the Caster and Target are not in the same map
+			if (buff.Caster.Map.Id != buff.Target.Map.Id)
+			{
+				var differentMapHelper = buff.Vars.GetInt(DifferentMapVarName);
```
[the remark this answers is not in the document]
Terotrous: "And the problem with the counting tick approach is that you have no control over the buff tick time, which a user or even an update might change at any time, which would could trigger this code much earlier or much later than expected."
This is true, but it's how the game actually works - it specifically notes in the tooltip that the buff expires 21-30 seconds after separating from the caster (the buff's update time is 10 seconds)

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+using Melia.Zone.Network;
+using Melia.Zone.World.Actors.Characters;
+using Melia.Zone.World.Actors.Characters.Components;
+using Melia.Zone.World.Items;
+
+namespace Melia.Zone.Buffs.Handlers
```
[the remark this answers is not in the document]
Terotrous: I agree that the "standard" status conditions can stay in the normal folder, though as far as I know this buff is tied to a specific skill.

src/ZoneServer/Skills/Handlers/Ardito/Recupero.cs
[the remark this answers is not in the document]
Terotrous: In the skill tooltip ingame.

src/ZoneServer/Skills/Handlers/Ardito/Ritirarsi.cs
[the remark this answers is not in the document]
Terotrous: Is there any meaningful difference between a multihit and sending multiple single-hit packets (besides the fact that the latter is more programmatically complicated and noisier from a network perspective)?

src/ZoneServer/Buffs/Handlers/Agility_Buff.cs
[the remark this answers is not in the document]
exectails: That's a matter of interpretation. Is a buff tied to a specific skill and class if, under normal circumstances, you only get it from that skill? What if they add an item that gives a certain buff, not necessarily this one, that was previously skill only? Do we move the file? What if a buff already belongs to multiple skills/items/whatever, indicated by not including a class name? What about the buffs that do include a class name, _because_ they're intended to be tied to a skill, do they go into that class' folder by default, no matter what? What if these buffs no longer belong to that class, because they were shuffled around with ReBuild? What about monster (skill) buffs, do we sort these into monster sub-folders?

These are rhetorical questions. I'm just trying to illustrate that giving order to the ~4600 buffs currently in the game is a bigger discussion than saying "sort them by what class you found them with first".

src/ZoneServer/Skills/Handlers/Ardito/Recupero.cs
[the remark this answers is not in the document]
exectails: Could be more precise? I don't see this in the skill's description.

src/ZoneServer/Skills/Handlers/Ardito/Ritirarsi.cs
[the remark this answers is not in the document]
exectails: A multi-hit-hit is visual only, while sending multiple hits actually allows for variety, delays, and checks between hits. We're  going to be implementing skills exactly as logged.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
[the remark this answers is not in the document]
Terotrous: Right, but the default value of updateTime is 10, so it is equivalent unless tinkered with, and the updateTime for this buff has no use except for this calculation.  I feel like the only reason you'd change it is if you wanted a longer grace period, in which case maybe you'd set it to 30 or something for a 60-90 second window.  In this case, if we instead hardcoded the time, you'd have to make a code change to get this behavior.

src/ZoneServer/Skills/Handlers/Ardito/Ritirarsi.cs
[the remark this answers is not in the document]
Terotrous: This kind of feels like a rabbit hole that's probably overkill at this stage in development.  If we know a skill is a manual multi-hit we can implement it that way, but zenmyaing every hitbox in the game on official to see if the server is doing some strange things with hit packets is probably something that can be left until later.

I would also note that our hit packet timing is inaccurate anyway (see the comments on Thrust for example), so attempting to get hit packets accurate to the byte is probably beyond what we can achieve right now.

src/ZoneServer/Buffs/Handlers/EnchantEarth_Buff.cs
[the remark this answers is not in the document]
exectails: What matters is not what we find likely or sensible, but reality and future changes.

Fact is, if you change the update time, regardless of the reason, the description will no longer match the behavior, and we can't easily fix that from the server. Meaning the best solution for us to keep server and description in sync, as developers of the server alone, is arguably to specifically check for a duration.

Additionally, this was also meant as general advice, because I assume Felip mostly followed your instructions, potentially without considering these consequences or side effects, which could result in him or others doing this again on the next buff, where the effects could be much more critical from the get go, without users doing unsensible modifications.

src/ZoneServer/Skills/Handlers/Ardito/Ritirarsi.cs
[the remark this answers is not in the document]
exectails: This is neither a rabbit hole nor overkill, it's simply what you do. You implement what you see in the packets. That doesn't mean that you can't make assumptions here and there, and once you've implemented a couple skills you can in many cases see from videos alone how they _probably_ work, but we're not going to start dumping dozens and hundreds of skills into the server that are based on nothing but guesses and good wishes.

PR #237, opened 2023-07-11
Update web.conf

Terotrous:
The download link in web.conf doesn't exist, so the server fails with an error.  I updated it to a working link, though I have some concerns about being forced to choose between 32 and 64 bit.


Terotrous:
I updated it to use the x86 nts version, though this isn't the version I'm using so I can't personally confirm that it works.

exectails:
It will, thanks 👍

PR #239, opened 2023-07-12
New Monsterinfo

Terotrous:
Updates the MonsterInfo command to work more like the one from RO.  Also fixes a bug in the Monster DB where the Amount values were not being read properly.


exectails:
Nice, thanks for PR!

PR #241, opened 2023-07-13
Drop rate customization

Terotrous:
Adds customized drop rates based on item type.  Of course, more item types can be added if desired.  Also updates MonsterInfo to reflect this.

### DIFF
--- a/src/Shared/Configuration/Files/World.cs
+++ b/src/Shared/Configuration/Files/World.cs
@@ -7,7 +7,13 @@ namespace Melia.Shared.Configuration.Files
 	/// </summary>
 	public class WorldConfFile : ConfFile
 	{
-		public float DropRate { get; protected set; }
+		public float SilverDropAmount { get; protected set; }
+		public float SilverDropRate { get; protected set; }
+		public float EquipmentDropRate { get; protected set; }
+		public float BlueOrbDropRate { get; protected set; }
+		public float RedOrbDropRate { get; protected set; }
+		public float GemDropRate { get; protected set; }
+		public float EtcDropRate { get; protected set; }
 		public int DropRadius { get; protected set; }
 		public int PickUpRadius { get; protected set; }
 		public int LootPrectionSeconds { get; protected set; }
@@ -36,7 +42,13 @@ namespace Melia.Shared.Configuration.Files
 		{
 			this.Include(filePath);
 
-			this.DropRate = this.GetFloat("drop_rate", 100);
+			this.SilverDropAmount = this.GetFloat("silver_drop_amount", 100);
+			this.SilverDropRate = this.GetFloat("silver_drop_rate", 100);
+			this.EquipmentDropRate = this.GetFloat("equipment_drop_rate", 100);
+			this.BlueOrbDropRate = this.GetFloat("blue_orb_drop_rate", 100);
+			this.RedOrbDropRate = this.GetFloat("red_orb_drop_rate", 100);
+			this.GemDropRate = this.GetFloat("gem_drop_rate", 100);
+			this.EtcDropRate = this.GetFloat("etc_drop_rate", 100);
 			this.DropRadius = this.GetInt("drop_radius", 25);
 			this.PickUpRadius = this.GetInt("pick_up_radius", 100);
 			this.LootPrectionSeconds = this.GetInt("loot_protection", 100);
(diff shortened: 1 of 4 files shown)


src/Shared/Configuration/Files/World.cs
```diff
@@ -7,7 +7,13 @@ namespace Melia.Shared.Configuration.Files
 	/// </summary>
 	public class WorldConfFile : ConfFile
 	{
-		public float DropRate { get; protected set; }
+		public float SilverDropAmount { get; protected set; }
+		public float SilverDropRate { get; protected set; }
+		public float EquipmentDropRate { get; protected set; }
+		public float BlueOrbDropRate { get; protected set; }
```
exectails, line 13: Honest question because I haven't played this game enough to know, do we need this level of control over the rates? Will there be a lot of people wanting to modify the orb drop rate? It doesn't hurt to have the option of course, but my concern is that everybody who wants a specific type of item to drop more frequently might advocate for that type to be added to this list, and we probably have to draw a line somewhere.

Terotrous: To be honest I think it is probably necessary.  Virtually every monster in the game drops an item from all of these types.  The only one you could maybe get rid of I think is the silver drop rate, but I added that based on our conversation yesterday.

Like in theory you could combine gems and orbs, but gems are much more powerful than orbs (currently), so I think it makes sense for orbs to be more common than gems.  Maybe you could combine the orbs types together, but their function is very different.  If we do end up making the blue orb into something that functions like a pet egg, then it should probably be rarer than a red orb.

Generally speaking I based the way this was set up based on RO's system (which lets you specifically set rates for consumables, equipment, cards, and other items), but TOS has even more items and systems compared to RO, so it makes sense that there might be slightly more categories.

kenedos: I think the orb drop rate having separated rates is a reasonable thing. They are unique enough to be treated as a category of item of their own and they drop from so many different monsters that manually editing each one would be painful.

I would say remove either SilverDropRate or SilverDropAmount though, as it may be confusing to have both. If keeping only SilverDropRate, make it multiply by how many times the actual drop goes past 100%. 
Example: Mob has 80% chance to drop silver, but server settings are SilverDropRate=300, so it would end up having 80 * 3 = 240% chance to drop silver. So it drops 2.4x the normal amount with 100% chance.

If keeping only SilverDropAmount, no 'complicated' math required. Statistically, you'll get same amount of silver anyways.

If you keep both, however, things get confusing because some mobs can cap at 100% silver drop chance _before_ others. So having SilverDropRate=200 and SilverDropAmount=200 does not equate final silver drop rate to be 4x.

exectails: > Example: Mob has 80% chance to drop silver, but server settings are SilverDropRate=300, so it would end up having 80 * 3 = 240% chance to drop silver. So it drops 2.4x the normal amount with 100% chance.

I'm honestly a bit confused how 240% chance to drop silver results in 2.4x the amount being dropped at 100%. I mean, I understand the idea, but I would never get that without an explicit explanation. As it is, silver has a drop amount and a drop chance. Wouldn't it make more sense to have both be configurable? Everybody knows the drop chance is not 100%, so you can tweak it to be more or less. And the other factor is the amount, which you could tweak as well if you so choose. Say I wanted 100% drop chance at vanilla amounts, or much reduced chances with much higher payouts, how would I do that if it's not two separate values?

Terotrous: I agree that I think that's a bit confusing, I prefer it to work in the more straightforward manner where one controls the drop rate and one controls how much drops and there's no crossover.  Note that for all other items, if the drop chance is over 100, you don't suddenly get multiples of them, it just becomes a guaranteed drop.

As for the orbs, I don't really agree that it's that similar, they're pretty distinct in both value and function.  I'd compare it to pet egg vs dead branch in RO.  Both do cause an entity to appear but they do very different things.

exectails: I view them both as summons, like you summon monsters in RO with dead branches, but one is potentially hostile towards you while the other assists you. But I guess that doesn't really matter^^

I might change the name of the options a little to create a clearer distinction between the two silver values, but other than that I think this is good. If there's no strong arguments against the current split approach I think it's ready to be merged.

Terotrous: Sure, if you have any suggestions for naming changes or anything like that, just list them and I'll make the changes, or it can be done later.

kenedos: I don't have strong arguments against the split approach, I think it's a decent architecture too ^^
Good work @Terotrous

exectails: > Good work

Seconded^^ Thanks for the PR!

PR #243, opened 2023-07-17
Random spawn point minor fix

kenedos:
Minor fix to make spawners spawn monsters on truly random spawn points when spawn amount > 1.


src/ZoneServer/World/Spawning/MonsterSpawner.cs
```diff
@@ -158,11 +158,11 @@ public void InitializePopulation()
 		/// <param name="amount"></param>
 		public void Spawn(int amount)
 		{
-			if (!_spawnAreas.TryGetRandom(out var spawnArea))
```
exectails, line 161: Oh. Yea, that will do it xD"

Issue #209, opened 2023-05-23
Add GUI


exectails:
As some users aren't fond of a series of console windows cluttering their screens, it might be beneficial to create a host application that can run and hide the windows in the background for these users. This could be accomplished by simply running the server applications hidden and redirecting their standard output to be able to show the servers' log output somewhere in the GUI.

In addition to hiding the consoles, this application could also provide some QoL features in the future, such as an account creator, authority setter, server (re)start/stop options, access to the future web API, etc.

Since this application isn't going to be part of the core, and depending on the implementation might also not work on all operating systems, it would probably make sense to create a dedicated project/repo for it.

exectails:
Alternative idea: We could make this application a "server manager" with a web interface. From there, you would be able to spin up the servers and control them, and it would work on all platforms. It could also be capable of connecting multiple machines to launch zones on different systems as needed.

I don't believe that this should necessarily be the default option, but it could work really well for people who dislike the console or have more advanced requirements than "launch five applications and be done with it".

felipecastagnarodecarvalho:
Some suggestion done by @Voudi

`- Adjusting basic server settings ('meta settings' like droprate or exp modification)
- GM settings (player list and commands as buttons like warp, mute, freeze, kill)
- Map settings (setting up spawn points, spawn zones/waymarks/surprise attacks, warps)
- Mob settings (altering mobs hp, skills, respawn time)
- Exporting/Importing settings to other clients`

My plan for the future:

- Broadcast message (send message to all players)
- Send mail (select item/items and the target player/player to receive a post box mail.
- Restore database backup
- Inspect inventories (from players including warehouse)

exectails:
Don't forget that the main point of this issue was to have a server manager though^^

PR #245, opened 2023-07-18
GUI host application

felipecastagnarodecarvalho:
Tecnologies/Libraries envolved:

- Laravel (backend)
- Vue (frontend)
- Tailwind (css)
- Inertia (Server Sided Rendering)

image


felipecastagnarodecarvalho:
image

exectails:
I checked out your app and would like to thank you for your efforts so far, as it seems like a rather neat admin panel. That being said, I'm not sure it's a perfect fit to include in the main repo, due to its broad functionality and reliance on several third-party frameworks and applications.

As is, I don't think it's simple enough to use for the people who would benefit most from it, as the installation alone is arguably even more complex than running Melia. Both a hypothetical admin panel as well as the proposed host application (#209) would ideally be usable by default and not require any additional steps to run. At most, they should require copying the folder from the optional web directory.

Furthermore, you're shaping this app into a one-stop shop for managing Melia, which, to be clear, is a great thing to have. But it feels too large and broad for my linking if that makes sense. I believe it's essential for the core application to maintain a certain level of simplicity and focus, whereas the creation and maintenance of your admin panel is a rather large undertaking in itself, not to mention that it adds almost 10 MB to the repo, which is twice as large as the entire source code. As such, I believe it would be better suited to exist as a separate project that users may download and add to their installation at their discretion, instead of shipping all versions of Melia with it.

I will leave this PR open for the moment, because there's some valuable code to be found here, like the API additions we might want to include in the core, for applications such as yours to utilize, but I don't think we'll merge the entire app.

exectails:
Since the PR has been sitting for a while, I'm going to close it now. We should still evaluate which changes to the API might make good candidates for the core, though for this to happen we'll have to pick the PR apart anyway. In the meantime, I'd like to encourage you to keep your branch updated and post your work on our [forum](https://nocodenolife.org/forum/65-melia/), for others to discover and potentially use.

PR #246, opened 2023-07-19
Mob Spawns Formatting Fixes

kenedos:
Opening this PR to fix previous mob spawn formats and some other issues:
- Added property overrides even for mobs that do not naturally spawn in map (but are in mob stats files)
- Fixed 'ep13_2_d_prison_1/2/3' not picking correct stat file
- Removed break line at end of script
- Removed space on "AddSpawnPoint" call
- Added GenType number to each spawn point group
- Ordered spawn points and spawners by gentype
- Changed "Spot" into "Rectangle"
- Added minimum mob spawn of 75% value of max mob spawn (rounded up). 75% felt like a decent inbetween because 50% was too low for certain mob groups.
- Added real map name at beggining of each script


exectails:
Thanks for the PR 👍 Though I'm still curious why your coordinates are slightly different from the ones I see in the IES files. Can you tell me how you're reading those?

exectails:
I just noticed there appears to be something going on with the ranges as well. In the test zone you have all area sizes as 20, but in my `gentype_test_zone` the `GenRange`s are all 1 and the `Range`s are almost all 40. Same thing for the `Rootcrystal_05`s on `d_abbey_22_4`. You have them as 20, in my files they're 40. I'm starting to wonder if we're using different files |D

exectails:
I did assume that you were using `GenRange`, because the vast majority of these values matched, but the `< 5` rule explains why some 1s turned into 20s. But in that case, why crank them up that high? Wouldn't make `min(5, genRange)` or  `min(20, genRange)` make more sense?

kenedos:
This is the code I created to generate those scripts putting together your IES tool and Akinue's older code. Be wary, it is a super ultra mess ^^;

It reads all IES files from "Data" folder though, so in theory should execute without needing any configurations.

https://github.com/kenedos/DataConverter

exectails:
I wonder if range 1 is supposed to be a fixed position spawn 🤔 In that case we'd want to use `Spot` instead.

kenedos:
> I wonder if range 1 is supposed to be a fixed position spawn 🤔 In that case we'd want to use `Spot` instead.

It would make sense tbh, it seems when the 'GenRange == 1' the mobs indeed have a VERY specific spawn (like behind a door, next to a rock, etc) but when I tried directly making their spawn in the given coordinates they would always be inside a wall or something. Maybe there's a special flag or something that changes the spawn behaviour when Genrange == 1.

> (...) Wouldn't `min(5, genRange)` or `min(20, genRange)` make more sense?

It was an arbitrary choice, but walking around the maps and taking a look at the spawns I just felt the ones with low values (like 10) were reasonable enough. The ones with '1' were way too far off, some would still not spawn the mob even setting them as 10 instead, so I decided upon 20.

kenedos:
I have removed the decimals from the spawn area coordinates and there were no issues with spawning the monsters.

Let me know if there's anything else to be done here.

exectails:
Now that the data is more uniform I'm noticing just a couple more oddities, like the ranges on the root crystals in `f_siauliai_2` getting increased from 10 to 100. What's the reasoning there? I'm also seeing a few (potentially) missing spawn points, such as another for `Pokubu` on `f_siauliai_2` at `X 2037, Z -30`, and spawners, like for `Wood_Carving` on `f_flash_64`.

kenedos:
Some data was 'manually' adjusted. Some of them were not spawning at all due to invalid positions (like the pokubu spawn point you mentioned) and some were not actually monsters (like the wood_carving you mentioned). These wood carvings are like training dummies but specific for musketeer quest so I imagined we wouldn't want them.

Adding the pokubu spawn back will probably not cause any issues, but they'll fail silently since the 'error' log when trying to spawn monsters in invalid positions are being surpressed now.

Any adjustments I did are within that project's code.

exectails:
Ah, I figured since we were spawning other dummies we would want these too, but if they're for some quest that might not be the case. Alright 👍

PR #247, opened 2023-07-24
Feature - Personal Storage

kenedos:
Adds feature of personal storage.

**Includes:**
- Open/Close personal storage
- Store/Retrieve items in personal storage
- Stacking of items to/from personal storage
- Swapping of items in personal storage
- 20 silver cost for personal storage
- Saving personal storage in SQL DB
- 'Multistack' toggleable feature
- Chatcommand to open personal storage anywhere
- Rita storage keeper in klaipeda

**Does not include:**
- Team storage
- Personal storage expansion

**Overview:**
Basic storage implementation using arrays. Index of array is it's position in storage and value is the item. Hardcoded storage size of 60 for time being.

'Multistack' feature allows toggling on/off the possibility of creating multiple stacks of same item class when storing. If it's toggled off, you can only store 1 stack of the same item class. If feature is toggled on and then off after a while, players may see multiple stacks of items in storage but once these are retrieved they'll not be able to create multiple stacks anymore.

Multistack implementation stores first to existing items and then any remainder to a new stack.

Multisack comes enabled by default.


src/ZoneServer/World/Actors/Characters/Components/PersonalStorageComponent.cs
```diff
@@ -0,0 +1,512 @@
...
+			return StorageResult.ItemNotFound;
+		}
+
+		/// <summary>
+		/// Returns specific item in position, or null
+		/// if it doesn't exist.
+		/// </summary>
+		/// <param name="position"></param>
+		/// <returns></returns>
+		public Item TryGetItemPosition(int position)
```
exectails, line 456: Minor thing, but adding an "At" in between Item and Position would more accurately describe what this method does.

system/db/features.txt
```diff
@@ -80,6 +80,9 @@
 		{ name: "ClericHealPartySelect", enabled: true },
 	]},
 	
+	// Enables storages to have multiple stacks of same item type
+	{ name: "StorageMultiStack", enabled: true },
```
exectails, line 84: The features file is intended for things that changed over the course of the game's life, like official changes to skills or NPCs, how properties worked, or things were calculated. The idea is that it should be easy to, for example, enable or disable all rebuild changes, or picking out a couple you like. Custom features on the other hand should go into the conf files.

src/ZoneServer/Network/Send.cs
```diff
@@ -3549,7 +3568,10 @@ public static void ZC_SOLD_ITEM_DIVISION_LIST(Character character, byte type, Li
 						zpacket.PutInt(items[i].Amount);
 						zpacket.PutInt(items[i].Price);
 						zpacket.PutInt(1);
-						zpacket.PutInt(items.Count - i - 1);
+						if (itemsPositions != null)
```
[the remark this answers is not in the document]
kenedos: Well ZC_SOLD_ITEM_DIVISION_LIST is sent to client both when player is browsing a npc store or the storage. The game shares some packets of npc stores for storage. In this case, the npc store cannot have empty items for sale so "items.count - i - 1" for item positions makes sense, but in storage, items can have positions that are not consecutive. Therefore, I also need to send the position of the item.

Basically I want to send that exact same packet but with one different integer logic. What I could do, instead, is duplicating the code and overload it into receiving the type "List<StorageItem>" instead of "List<Item>" which would be a wrapper around 'Item' class but with a property of the position the item is in storage.

src/ZoneServer/World/Actors/Characters/Components/PersonalStorageComponent.cs
```diff
@@ -0,0 +1,512 @@
...
+					}
+				}
+			}
+		}
+
+		/// <summary>
+		/// Creates a new storage component
+		/// </summary>
+		/// <param name="account"></param>
+		public PersonalStorageComponent(Character character) : base(character)
```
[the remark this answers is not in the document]
kenedos: I thought about reusing this code for team storage but then I looked how team storage works and realized the logic there is quite different. Maybe one or two methods would be similar and reused but I think it's better keeping their implementations separated. Team storage doesn't allow dragging items around and allows for multiple item retrieval, for example. It also automatically reorder items positions when you retrieve an item, which personal storage doesn't. Team storage will most likely be implemented as a linked list while personal storage should remain as an array.

I honestly didn't give much thought on custom storages or storages shared between multiple players. If you prefer, I can move it to the world and have checks for character login/logoff to load/unload storage from memory. Tbh my methodology of programming is implementing things on the simplest architecture possible and then refactor them later on as needed, but I'm of course not against the idea of future-proofing it ^^

src/ZoneServer/World/Actors/Characters/Components/PersonalStorageComponent.cs
```diff
@@ -0,0 +1,512 @@
...
+				for (var i = 0; i < existingItems.Count; i++)
+				{
+					var stackedAmount = this.AddItemStack(existingItems[i], itemToStore.Amount - addedAmount);
+					addedAmount += stackedAmount;
+					if (stackedAmount > 0)
+					{
+						Send.ZC_ITEM_ADD(this.Character, existingItems[i], existingPositions[i], stackedAmount, InventoryAddType.New, InventoryType.Warehouse);
+					}
+				}
+				if ((itemToStore.Amount - addedAmount) == 0)
```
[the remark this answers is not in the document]
kenedos: You're not paranoid, in fact, my first verions of code had the check for '<= 0' even knowing it could never happen. I will change it back.

system/conf/commands.conf
```diff
@@ -136,6 +136,9 @@ broadcast: 50,-1
 // Kicks a player
 kick: 50,-1
 
+// Opens personal storage
+storage: 50,50
```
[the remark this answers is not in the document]
kenedos: I'm not sure I understand what you meant. I made this command '50,50' and not '50,-1' to enable GMs to open storage for other players so they can browse it (for whatever reason). But yeah, GMs cannot open another player's storage in their own client, only the in the player's client.

src/ZoneServer/World/Actors/Characters/Components/PersonalStorageComponent.cs
```diff
@@ -0,0 +1,512 @@
...
+		}
+
+		/// <summary>
+		/// Returns true if storage is full.
+		/// </summary>
+		/// <returns></returns>
+		public bool CheckStorageFull()
+		{
+			lock (_syncLock)
+				return _occupiedPositions == _maxStorage;
```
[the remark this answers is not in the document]
kenedos: Yeah I added that as some sort of optimization so you don't have to iterate through the array looking for the item but I was also wondering if that was even necessary since it is basically O(1) anyways.

src/ZoneServer/Database/ZoneDb.cs
```diff
@@ -635,6 +637,91 @@ public void SaveCharacterItems(Character character)
...
+				}
+
+				foreach (var storageItem in character.PersonalStorage.GetStorage())
+				{
+					var newId = 0L;
+
+					// Save the actual items into the items table and the
+					// storage-items into the storage table,
+					// while linking to the items.
+					// TODO: Add generic item load and save methods, for
```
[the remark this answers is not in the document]
kenedos: Well, at least you didn't catch part of my commit where it says "Author: Salman T. Khan"

src/ZoneServer/World/Actors/Characters/Components/PersonalStorageComponent.cs
```diff
@@ -0,0 +1,512 @@
...
+
+			Item itemToStore = new Item(item);
+			itemToStore.Amount = Math.Min(itemToStore.Amount, amount);
+
+			// For stackable items, finds existing storage items.
+			var existingPositions = new List<int>();
+			var existingItems = new List<Item>();
+			if (itemToStore.IsStackable)
+			{
+				this.FindStackableItems(itemToStore, out existingPositions, out existingItems);
```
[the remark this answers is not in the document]
kenedos: I opted to not take the FillStacks() approach. 

The reason for that is that I want to remove item from player **before** adding it to storage. It may be a bit redundant because I do check if item exists in player's inventory at very beggining of "StoreItem()" method but the "character.Inventory.Remove()" method I call has a return state and I want to make use of that and check it before proceeding further with adding the item to storage.

kenedos:
I have refactored the implementation into a more reusable approach.

- 'Storage' class is now abstract and has the core of storage implementation
- 'PersonalStorage' class inherits from it and implements single character storage
- Changed _storageItems data type to SortedList<Item> to be compatible with either PersonalStorage or TeamStorage
- Whether you want an ordered storage or not now can be achieved by either loading storage with Add() or AddAtPosition()
- Removed unecessary fields from 'Character' class, now 'isBrowsing' field is in 'PersonalStorage' class.
- The new implementation allows to easily create shared storages between multiple characters.
- The new implementation also allows for easy creation of other storage types (i.e. guild storage).
- Added a few more methods for better manipulation of storages.

src/ZoneServer/Network/Send.cs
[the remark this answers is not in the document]
exectails: It does seem like creating more wrapper classes around items might be worth considering. That way we could store additional information for them, similar to how we do it in the database, without polluting the `Item` class.

src/ZoneServer/World/Items/Item.cs
```diff
@@ -112,6 +112,25 @@ public Item(int itemId, int amount = 1)
...
+		{
+			this.Id = other.Id;
+			this.LoadData();
+			this.Price = other.Price;
+			this.IsLocked = other.IsLocked;
+			this.OriginalOwnerHandle = other.OriginalOwnerHandle;
+			this.RePickUpTime = other.RePickUpTime;
+			this.OwnerHandle = other.OwnerHandle;
+			this.LootProtectionEnd = other.LootProtectionEnd;
+			this.Properties = other.Properties;
```
exectails, line 129: You shouldn't just copy the reference, because changes to one item will then affect the copies as well and vice versa. Instead, I added a `CopyFrom` method to `Properties` a while back for this purpose.

src/ZoneServer/World/Storage/PersonalStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+	{
+		/// <summary>
+		/// Character that owns this personal storage.
+		/// </summary>
+		public Character Owner { get; private set; }
+
+		/// <summary>
+		/// Whether the owner is currently browsing this storage.
+		/// </summary>
+		public bool IsBrowsing { get; private set; }
```
exectails, line 28: Wouldn't this property be required for other types of storages as well? I assume you would do the same `IsBrowsing` checks for team storage, and we'd definitely need it to check whether globally accessible storages are currently being browsed, right?

src/ZoneServer/World/Storage/PersonalStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+
+		/// <summary>
+		/// Constructor.
+		/// </summary>
+		/// <param name="owner"></param>
+		public PersonalStorage(Character owner) : base()
+		{
+			this.Owner = owner;
+			this.IsBrowsing = false;
```
exectails, line 37: This assigment is unnecessary because the property is initialized to its default value (`false`) automatically.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+					return 0;
+
+				availableAmount = item.Amount;
+			}
+
+			return availableAmount;
+		}
+
+		/// <summary>
+		/// Adds the available amount of an item to storage.
```
exectails, line 195: Unfortunately this comment and the name of the method aren't very descriptive. What is the "available amount", what happens to the item, what are the potential results? You essentially have to read the method to understand what it does.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+		/// Updates client.
+		/// </summary>
+		/// <param name="character">Character that is performing interaction</param>
+		/// <param name="objectId">ObjectId of item to store</param>
+		/// <param name="amount">Amount of item to store</param>
+		/// <param name="invType">Storage inventory type</param>
+		/// <returns></returns>
+		protected StorageResult StoreItem(Character character, long objectId, int amount, InventoryType invType)
+		{
+			var item = character.Inventory.GetItem(objectId);
```
exectails, line 246: Would it not feel more natural to get the item you want to store and pass it in instead of getting it from the character in here? That way you could also add arbitrary items, and not only ones found on the character already.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+		/// <returns></returns>
+		protected StorageResult StoreItems(Character character, List<long> objectIds, InventoryType invType)
+		{
+			foreach (var objectId in objectIds)
+			{
+				var item = character.Inventory.GetItem(objectId);
+				if (item == null)
+					return StorageResult.ItemNotFound;
+
+				var result = StoreItem(character, objectId, item.Amount, invType);
```
exectails, line 289: You're missing a couple more `this.` on a few method calls.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+		/// <summary>
+		/// Removes an item from storage and gives to a character.
+		/// Updates client.
+		/// </summary>
+		/// <param name="character"></param>
+		/// <param name="objectId"></param>
+		/// <param name="amount"></param>
+		/// <param name="invType"></param>
+		/// <returns></returns>
+		protected StorageResult RetrieveItem(Character character, long objectId, int amount, InventoryType invType)
```
exectails, line 306: Similar to storing this feels somewhat restrictive, not allowing us to get an item without adding it to a character's inventory. I realize this is (probably) not a function we'll need during normal gameplay, but I could think of a couple reasons why someone might want to do that, be it for administrative purposes, support, or even events.

Wouldn't it be relatively simple to make the storage more of a general purpose item list that the child classes than expand to do what they need? That would arguably make for a much more flexible system, and it would also remove certain patterns, like passing the character to the storage for getting items or sending updates.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+			return StorageResult.Success;
+		}
+
+		/// <summary>
+		/// Creates a new storage.
+		/// </summary>
+		/// <param name="account"></param>
+		public Storage()
+		{
+			this.Id = Interlocked.Increment(ref Ids);
```
exectails, line 359: Do we have a need for storages possessing a unique id? Potentially, but it should probably be persistent in that case, right? For example, let's say we wanted to create guild storages, which don't exist in the game to my knowledge, then we'd create a storage for a guild, with an id, and reference that to get/open it. We can't have the ids on the server collide with the database ids in that case though.

Actually, with the potential existence of different kinds of storages in mind, wouldn't it make sense to have a general purpose storage table that different entities can then take ownership of, such as characters, accounts, guilds, or even the server itself, for storages accessible to everyone?

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+		/// </summary>
+		/// <returns></returns>
+		public int GetStorageSize()
+		{
+			return _maxStorage;
+		}
+
+		/// <summary>
+		/// Increases max storage size by the given number.
+		/// Storages can never decrease in size.
```
exectails, line 611: This, too, seems somewhat restrictive and also potentially problematic. What if someone accidentally called this method twice for example, should they then be forced to modify the database directly or attempt to understand the inner workings of this class? Don't you think the storage itself should be capable of decreasing the size, even if it might come at the cost of items? Though we could also simply rearrange the items or pass any potentially removed items to the caller in that case and let them deal with it. For example, they could be added to the inventory or be dropped on the floor.

src/ZoneServer/World/Storage/Storage.cs
```diff
@@ -0,0 +1,691 @@
...
+		/// Swaps items in two positions.
+		/// Moves item into position2 if empty.
+		/// </summary>
+		/// <param name="position1">Item1 object id</param>
+		/// <param name="position2">Item2 object id</param>
+		/// <returns></returns>
+		public StorageResult Swap(int position1, int position2)
+		{
+			var item1 = this.TryGetItemAtPosition(position1);
+			var item2 = this.TryGetItemAtPosition(position2);
```
exectails, line 634: `TryGet` methods usually return a boolean directly and the actual return value in an out parameter, allowing patterns like this:
```cs
if (foo.TryGetX(out var x))
    // got x, do something with it
```
Deviating from that pattern is potentially confusion, making me think that the method should probably be renamed.

system/conf/commands.conf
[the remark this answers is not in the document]
exectails: The line I commented on might've been a bit random, I just thought of this once more here^^ I understand that you allowed GMs to open people's storages for them, I just wanted to mention once more that a general purpose storage system that operates independently of the characters would make it possible for GMs or even other players to also open players' storages, which could potentially be useful.

exectails (changes requested):

src/ZoneServer/Database/ZoneDb.cs
```diff
@@ -635,6 +637,85 @@ public void SaveCharacterItems(Character character)
...
+						// Check item, in case its data was removed
+						if (!ZoneServer.Instance.Data.ItemDb.Contains(itemId))
+						{
+							Log.Warning("ZoneDb.LoadStorageItems: Item '{0}' not found, removing it from storage.", itemId);
+							continue;
+						}
+
+						var item = new Item(itemId, amount);
+
+						character.PersonalStorage.AddAtPosition(item, position, out var addedAmount);
```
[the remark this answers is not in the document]
kenedos: You're right, this was an oversight from my end when I switched the data structure from Array to SortedList. Thanks for noticing it.

src/ZoneServer/World/Storage/PersonalStorage.cs
kenedos: Well, I thought about this one and I could potentially move that property to the parent class but then I imagined if we want to implement, say, shared storages, we would need this "IsBrowsing" to be some sort of list for the current characters browsing it.. I think this is better being left in child class, as the child should actually implement the logic on how users actually 'browse' the storage.

src/ZoneServer/World/Storage/Storage.cs
kenedos: The problem with making things more 'universal' like that is how to manage it. 

Right now, we have personal storage as property inside a character and team storage as property inside account. This makes things pretty straightforward to save/load. Character logs in? Load the storage. Character logs off? Save the storage. The limitation is that others can only access your storage if they have your character/account object reference.

If we had a dictionary of storages in WorldManager for all storage types we would need to:
- Character logs off? lock the dictionary, get his storage, save his items to db.
- Character logs in? read db for his stored items, lock the dictionary, add his storage to it.

Basically same process but we're also locking all storages in game while any update occurs to them. I think this could cause some slow downs in servers with a lot of players, not to mention we're currently also saving storage even when a character changes map. 

I also don't see much of a benefit of having that storage universal table. If someone wants to implement a GM command that adds items to all player's personal storages, they can just grab all online characters and iterate over each character's storage property and add it.

I'm going to be honest and say the unique ID is just there because I followed the same idea on all other actor classes in Melia (thinking about it, storages should be actors too). But giving it a little bit more thought, I think the ID is still good to have in the case someone wants to implement multiple personal storages for a same character (for example).

kenedos: Furthermore, having a system where GMs can open a characters' storage even when the target player is offline would probably mean we need to keep all storages in memory at all times, not sure if that's a good idea. That, or load from SQL DB when the GM attempts to open it, which could be exploitable.

src/ZoneServer/World/Storage/PersonalStorage.cs
exectails: Would we ever allow a storage to be accessed by multiple players at once? It feels like there'd be a big potential for bugs and exploits with that. In RO for example they restricted the guild storage access to one player at a time.

src/ZoneServer/World/Storage/Storage.cs
exectails: Well, it seems like the natural evolution of this idea to me. We'll undoubtedly want to have storages that can (potentially) be accessed by different people, like guild storages for example. Another idea for a potential application might be to open a storage window to retrieve drop items from a monster, like in WoW. And I'm sure people will come up with even more ideas. This will require us to be able to create storages on the fly, as well as load and save them.

At that point we'd have multiple storage systems though. The personal storage, potentially the team storage if it's different enough, and all the dynamic storages. And since these systems wouldn't quite work the same way, we'd probably end up with a lot of redundant code between them. But the dynamic system could technically support every type of storage. We could create a storage and simply assign it to a player. And given the requirements of dynamic storages, handling only one player at a time would be simple to them. We also wouldn't necessarily have to keep all storages in memory at all times, as we could load and unload player storages when they log in/out.

In short, dynamic storage support would future proof this feature.

exectails:
Personal Storage is high on my todo list, but I feel like it's important we get this right early on, as it's been my experience that refactoring item-/inventory-related code can be rather tedious. Seeing how the last couple comments have not been addressed for the most part yet, I might take this promising PR as the base for completing the feature in the near future.

src/ZoneServer/World/Storage/PersonalStorage.cs
kenedos: I'm leaving this as is for now, as this property being in the children allows for more flexible future implementations of other types of storage.

src/ZoneServer/World/Storage/Storage.cs
kenedos: I'm not sure if this was changed or what, but after looking again at my code I realized this is not the case.

In Storage.cs class I am providing 'RetrieveItem()' as an easy way to retrieve the item **for a character** since it will also send packets to the character received in parameter.

However, this is not the only method that allows you to get items from storage. You can also call GetItem() method which is public and allows for retrieval of items in storage without need of character parameter. Same for Add() and AddAtPosition() methods, which are all public.

src/ZoneServer/World/Storage/Storage.cs
kenedos: What is the conclusion to this discussion? Let me know if I should remove the unique IDs or leave them be for now.

src/ZoneServer/World/Storage/Storage.cs
kenedos: The other method in this class allows for that.
This method is just a simple shortcut to do so while also updating a character.

PR #250, opened 2023-07-29
Blue orb pets

Terotrous:
Adds a new config option to make Blue Orbs behave more like pets - you can reuse them and they never expire.


src/ZoneServer/Network/PacketHandler.cs
```diff
@@ -764,8 +764,8 @@ public void CZ_ITEM_USE(IZoneConnection conn, Packet packet)
 					return;
 				}
 
-				// Remove consumeable items on success
-				if (item.Data.Type == ItemType.Consume)
+				// Remove consumeable items on success, unless it's a Blue Orb and the Pet System is turned on
```
exectails, line 767: While this is a simple way to accomplish not consuming certain types of items, I feel like hardcoding this sets a bad precedent. The next person with this problem might think this is the way to go and add their exception as well, then comes another and another and we quickly have quite the mess on our hands. Instead, we should think of a way to let the item usage function specify whether to skip the consumption of the item.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -238,16 +260,18 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
 					// recreate it on the other side, from the summoning
 					// script. All this could need a clean up.
 
-					if (!ZoneServer.Instance.Conf.World.BlueOrbFollowWarp)
+					if (!ZoneServer.Instance.Conf.World.BlueOrbFollowWarp && !ZoneServer.Instance.Conf.World.BlueOrbPetSystem)
```
exectails, line 263: Wouldn't it be confusing if one conf option disables another implicitly? I realize you wouldn't typically want your "blue orb pets" to stay behind, but isn't that an choice the user should make? Maybe someone wants to limit BO pets to one map, in which case they could enable the pet option but leave follow warp disable.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -238,16 +260,18 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
...
 						while (true)
 							yield return this.Wait(10000);
 					}
 					else
 					{
 						if (this.Entity is Mob mob)
+						{
 							this.Entity.Map.RemoveMonster(mob);
-						yield break;
+							yield break;
```
exectails, line 273: The yield break coming either way after the if was intentional, is there a reason for this change? Because now, if the entity were a player for example, it wouldn't break, and the following would continue, which we don't want if the target switched maps.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -190,8 +191,20 @@ private void Disappearances()
 			lock (_monsters)
 				toDisappear = _monsters.Values.Where(a => a.DisappearTime < now).ToList();
 
-			foreach (var monster in toDisappear)
+			foreach (var monster in toDisappear) {
+
+				// If this is a blue orb summon, inform the summoner that it expired
+				if (monster is Mob mob) {
+					if (mob.MonsterType != MonsterType.Mob && mob.Components.Get<AiComponent>()?.Script.GetMaster() != null && mob.Components.Get<AiComponent>()?.Script.GetMaster() is Character master)
```
exectails, line 198: If you're going to call multiple methods on one object, such as an AI script or a character, you should create a reference to shorten the code a little. This makes it easier to read and improves performance.

```cs
if (monster is Mob mob)
{
    var master = mob.Components.Get<AiComponent>()?.Script.GetMaster();

    if (mob.MonsterType != MonsterType.Mob && master is Character masterCharacter)
```

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -190,8 +191,20 @@ private void Disappearances()
 			lock (_monsters)
 				toDisappear = _monsters.Values.Where(a => a.DisappearTime < now).ToList();
 
-			foreach (var monster in toDisappear)
+			foreach (var monster in toDisappear) {
```
exectails, line 194: Please try to conform to the general code style, as well as the coding conventions layed out in [CONTRIBUTING.md](https://github.com/NoCode-NoLife/melia/blob/master/CONTRIBUTING.md). Curly braces should be placed on their own line.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -190,8 +191,20 @@ private void Disappearances()
...
+
+				// If this is a blue orb summon, inform the summoner that it expired
```
exectails, line 196: This comment seems misleading, because what's happening is not that the character gets informed in some way. All that happens is that the orb summon variables are reset. Missing comments are usually preferable over incorrect ones.

exectails (changes requested):
Thanks for the PR, though I gotta say that the code feels a little hacked in. A lot of new checks in a lot of random places, as few changes as possible to accommodate this new feature, and very little documentation about what's actually happening. For example, to understand what the perm variables do now, as well as what resetting them does, you need to actually read all the relevant code, which isn't optimal. I like the option you're introducing here, but I believe it could be fleshed out a little better.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
Terotrous: It actually does still respect the follow warp setting with the pet system turned on, the effect of this is in summoning.cs.  Basically, if the pet system is turned off, when you go through a warp, your monster just stays there until it despawns.  This is not a valid option if the pet system is turned on as the monster has no time limit and will just get stuck there forever, so instead, if follow warp is turned off, it removes your monster (as per what would normally happen with follow warp turned on), but it doesn't respawn it on the other side, forcing you to recall it manually.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
Terotrous: Nope, that was a mistake, at one point I had more code there, it got messed up when I removed it.  I've set it back to the original now.

src/ZoneServer/World/Maps/Map.cs
Terotrous: Looks like I missed those two.  It's been fixed.

src/ZoneServer/World/Maps/Map.cs
Terotrous: I originally did it that way because I wasn't sure if it was safe against null.  It does appear to be, though (I just ran a test on it), so I've simplified it.

src/ZoneServer/World/Maps/Map.cs
Terotrous: I've adjusted the comment to be more clear about what it actually does.

src/ZoneServer/Network/PacketHandler.cs
Terotrous: I agree, though I didn't see any other way to achieve this with the current implementation except for possibly changing the item database, which I think is even worse.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
exectails: Right, but if I were a user and I turned "blue orb pets" on and "follow warp" off I might expect them to stay behind. If that's not going to be the case this behavior should be documented in the conf files to avoid confusion.

Terotrous: That's a good point, I've added more detail to the description now.

src/ZoneServer/Network/PacketHandler.cs
exectails: I've added a third item use result type that can be used to specify that a consumable should not be decremented.

Terotrous: I've updated the code so that it uses it.  This does seem cleaner.

src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -528,6 +529,9 @@ public void Warp(int mapId, Position pos)
 			if (!ZoneServer.Instance.World.TryGetMap(mapId, out var map))
 				throw new ArgumentException($"Map with id '{mapId}' doesn't exist in world.");
 
+			if (ZoneServer.Instance.Conf.World.BlueOrbFollowWarp || ZoneServer.Instance.Conf.World.BlueOrbPetSystem)
```
exectails, line 532: I'm not quite happy with this code duplication. Doesn't it feel more like there should be an event one could subscribe to to be informed that a player warps? That way we could even decouple this from the character, which arguably shouldn't have to care about whether it even has a follower.

src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -1131,6 +1138,13 @@ public bool TakeDamage(float damage, ICombatEntity attacker)
 				return true;
 			}
 
+			// inform the follower that I got hit so it can aggro
+			if (this.Variables.Temp.TryGet<Mob>("Melia.BlueOrbSummon.Monster", out var follower))
```
exectails, line 1142: If we're going to extend the hit information, shouldn't we perhaps directly go for a more general approach? We might also have a need for monsters reacting to other monsters getting hit for example, and approaching this from this direction, where we inform all relevant parties that something happened, feels a little disjointed and tedious.

How about we let the AIs observe the world instead? If we were to add a hit server event we could potentially use it to check for all kinds of combat actions, and all the code would be in the AiScript, keeping things tidy.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -282,6 +282,15 @@ public bool TakeDamage(float damage, ICombatEntity attacker)
 			if (this.Components.TryGet<AiComponent>(out var ai))
 				ai.Script.QueueEventAlert(new HitEventAlert(this, attacker, damage));
 
+			if (attacker is Character attackCharacter)
```
exectails, line 285: Frankly, this doesn't feel like it belongs here. It _works_, the follower of the attacker gets informed about the attack, but, again, don't you think the follower should be able to react on its own?

This pattern has a high risk of quickly getting unwieldy, because the moment we add pets we'd have to notify those too. Then we might want to add something like a clone command like in RO, which would add another type of follower. Afaik there's also skills that summon followers, and so on. If we were to move the handling to the AI script though they could all just do their thing, independently of the Character and Mob classes.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -307,11 +316,31 @@ public void Kill(ICombatEntity killer)
 
 			this.DisappearTime = DateTime.Now.AddSeconds(2);
 
-			if (killer is Character characterKiller)
+			// If I have a master, remove my variables from my master so I can be resummoned
```
exectails, line 319: Minor thing, but I believe we don't use a class' POV to document its behavior anywhere else in the code, making it feel a little out of place.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -307,11 +316,31 @@ public void Kill(ICombatEntity killer)
...
+			if (this.Components.Get<AiComponent>()?.Script.GetMaster() != null)
+			{
+				if (this.Components.Get<AiComponent>()?.Script.GetMaster() is Character master)
```
exectails, line 322: This is another instance where you get the master over and over, here as well as below.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -190,8 +191,23 @@ private void Disappearances()
...
-			foreach (var monster in toDisappear)
+			foreach (var monster in toDisappear) 
+			{
+
+				// If this is a blue orb summon, remove the variable from the summoner
+				if (monster is Mob mob) 
+				{
+					if (mob.MonsterType != MonsterType.Mob && mob.Components.Get<AiComponent>()?.Script.GetMaster() is Character master)
+					{
+						master.Variables.Temp.Remove("Melia.BlueOrbSummon.Monster");
```
exectails, line 202: Not a big fan of this happening in multiple places, as the redundancy makes it more likely that one instance might be changed but not the other. At the very least this should probably be a method, perhaps on Mob. Though it would be even better if we didn't need to explicitly add it in multiple places in the first place.

Hm, I just realized, there is a flaw that needs to be fixed, isn't there? The disappear time is currently two seconds for monsters, which means that a player could summon a new monster in between the kill and the disappearance. The variables would then be reset despite them having a new follower already.

system/scripts/ais/basic_monster.cs
```diff
@@ -49,8 +49,16 @@ protected IEnumerable Attack()
...
 		SetRunning(true);
 
+		var master = GetMaster();
+
 		while (!target.IsDead)
-		{
+		{			
+			if (master != null)
+			{
+				if (!MasterAttackCheck(master))
```
exectails, line 58: Is this the appropriate place to do this or should it be an event like the other checks at the bottom of the script? Seems to me like it would be more consistent to handle all "target loss events" in the same way.

system/scripts/items/summoning.cs
```diff
@@ -55,23 +88,44 @@ public ItemUseResult SCR_USE_SUMMONORB_ENEMY(Character character, Item item, str
...
-		if (!character.Variables.Perm.TryGet<DateTime>("Melia.OrbSummon.DisappearTime", out var disappearTime))
+		if (!character.Variables.Perm.TryGetInt("Melia.BlueOrbSummon.MonsterId", out var monsterClassId))
 			return;
 
-		if (DateTime.Now > disappearTime)
+		// If follow warp is not enabled, your monster is released from you when you leave a map,
+		// functionally causing the master to forget about it and be able to summon another monster.
+		if (!ZoneServer.Instance.Conf.World.BlueOrbFollowWarp) 
+		{
+			character.Variables.Perm.Remove("Melia.BlueOrbSummon.MonsterId");
```
exectails, line 101: Again a lot of redundancy.

exectails (changes requested):

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -194,6 +195,40 @@ protected IEnumerable Animation(string packetString)
...
+		/// It stops attacking if the master dies, leaves the map or forces it to teleport (via getting too far away)
+		/// </summary>
+		/// <param name="followTarget">The target to follow.</param>
+		/// <param name="minDistance">The minimum distance to the target the AI attempts to stay in.</param>
+		/// <returns>true if the monster should continue to attack, false if it should stop</returns>
+		protected bool MasterAttackCheck(ICombatEntity followTarget, float minDistance = 50)
+		{
+			if (followTarget.Map.Id != this.Entity.Map.Id)
+			{
+				RemoveAllHate();
```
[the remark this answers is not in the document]
Terotrous: This is fixed.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -194,6 +195,40 @@ protected IEnumerable Animation(string packetString)
...
 
+
+		/// <summary>
+		/// Makes a monster with a master check to see if it should cease attacking.
+		/// It stops attacking if the master dies, leaves the map or forces it to teleport (via getting too far away)
+		/// </summary>
+		/// <param name="followTarget">The target to follow.</param>
+		/// <param name="minDistance">The minimum distance to the target the AI attempts to stay in.</param>
+		/// <returns>true if the monster should continue to attack, false if it should stop</returns>
+		protected bool MasterAttackCheck(ICombatEntity followTarget, float minDistance = 50)
```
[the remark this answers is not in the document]
Terotrous: I changed it to a much simpler method called MasterGone that calls EntityGone (and also checks the distance).  The code that checks this method resets the hate if it returns true.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -233,22 +268,13 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
...
-					}
-					else
-					{
-						if (this.Entity is Mob mob)
-							this.Entity.Map.RemoveMonster(mob);
-						yield break;
-					}
+					// them get stuck in a loop. Note that if the follow warp
+					// or pet settings are on, the monster gets unsummoned
+					// when the master goes throw a warp, so this logic does
```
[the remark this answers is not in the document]
Terotrous: This is fixed.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -329,6 +340,10 @@ private void HandleEventAlerts()
 						{
 							if (hitEventAlert.Target.Handle == this.Entity.Handle)
 								this.IncreaseHate(hitEventAlert.Attacker, _hatePerHit);
+							else if (hitEventAlert.Target.Handle == this._masterHandle)
```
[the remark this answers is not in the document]
Terotrous: This is fixed.

src/ZoneServer/World/Actors/Characters/Character.cs
Terotrous: This is the one thing I haven't touched.  I think the current implementation is still fairly simple and it's still a big improvement over having the pet constantly check to see if the player left, which caused a number of small issues.  Maybe adding an event could be a future enhancement.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: This is fixed.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: This is fixed.

src/ZoneServer/World/Maps/Map.cs
Terotrous: I made a method that does this.  As for this particular call, this is actually not intended to operate on dead monsters, but rather ones that expire due to the disappear time limit, as such I added a check that makes sure the monster is still alive at the time of removal that fixes that potential issue.

In theory, we could remove the check on death and only check it here, but this introduces a very minor issue where if you leave the map within 2 seconds of your follower dying, you can save it, since that code won't have executed yet.  As such, I've left the check on death for now.

system/scripts/ais/basic_monster.cs
Terotrous: I moved it to an event.

system/scripts/items/summoning.cs
Terotrous: This also uses the new method to clear those variables.

src/ZoneServer/World/Actors/Characters/Character.cs
Terotrous: So are you suggesting that perhaps, all entities within an area are informed about hit events to any target?  This actually could be needed, for example in RO, there are the concept of Assist monsters, which aggro if a monster of the same species is hit within a certain range.  This could be used to make the pet aggro, too, eliminating the need for the player to inform them of the hit.  I feel like this is beyond the scope of this PR, though.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: Yeah, I had a similar thought process about it, though under the current implementation it seems to be the only way it can be done.  As I mentioned above, if anyone nearby was informed whenever something gets hit, neither of those checks would be necessary, but it's a fair bit of a change to the way the hit events are sent.

src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -552,6 +556,9 @@ internal void WarpChannel(int channelId)
 			_warping = true;
 			_destinationChannelId = channelId;
 
+			if (ZoneServer.Instance.Conf.World.BlueOrbFollowWarp || ZoneServer.Instance.Conf.World.BlueOrbPetSystem)
```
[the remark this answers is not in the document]
Terotrous: Yes, this works, I just tested it.  I was initially under the impression that FinalizeWarp occurred after the client disconnected from the map, but it looks like it occurs before, so this allows us to eliminate the duplication.

src/ZoneServer/Scripting/AI/AiScript.cs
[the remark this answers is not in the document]
exectails: I mean the use of `this.` to reference properties and methods, like so `if (this.EntityGone(master))`. This isn't enforced for scripts, but all internal code should adhere to the coding conventions.

src/ZoneServer/World/Actors/Monsters/Mob.cs
exectails: Well, as long as we're in agreement that it should be updated soon^^

src/ZoneServer/Scripting/AI/AiScript.cs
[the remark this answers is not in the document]
Terotrous: Oh, I was thinking the issue related to private variables.  Which classes use this and which don't still slightly confuses me, but I've tried to make sure it's consistent with other methods in the class.

exectails:
Thanks again for the PR. I merged it, cleaned it up a little, and implemented some of the improvements we talked about. I tested the functionality briefly, but I'd welcome you going over it again as well, to make sure everything is in order.

PR #251, opened 2023-08-01
Fix for crashing client on selling certain specific items

kenedos:
Fix for ZC_SOLD_ITEM_DIVISION_LIST crashing client when no property items are received.

- Forces the "Cooldown" (7266) property set to value 0 when no other property is found. This seems to be R1 behaviour.
- We may consider forcing this property onto every item created if this crashing issue happens with other packets as well.


exectails:
Thanks 👍

PR #252, opened 2023-08-03
Adding Elite and Jackpot monsters

Terotrous:
This adds Elite and Jackpot monsters.  A couple things are still todo, in particular elites do not yet have minions and can't use their special puddle attack.  Also, TwinkleBuff might not be the right buff for Gold Monsters.


src/Shared/Configuration/Files/World.cs
```diff
@@ -37,6 +37,22 @@ public class WorldConfFile : ConfFile
 		public bool BlueOrbFollowWarp { get; protected set; }
 		public bool BlueOrbPetSystem { get; protected set; }
 
+		public float BlueJackpotSpawnChance { get; protected set; }
+		public float BlueJackpotExpRate { get; protected set; }
+		public float SilverJackpotSpawnChance { get; protected set; }
+		public int SilverJackpotRolls { get; protected set; }
```
exectails, line 43: What are "rolls" in this context? That seems like a rather unusual term.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -290,8 +290,21 @@ public void Kill(ICombatEntity killer)
...
-			var expRate = ZoneServer.Instance.Conf.World.ExpRate / 100.0;
-			var classExpRate = ZoneServer.Instance.Conf.World.ClassExpRate / 100.0;
+			var worldConf = ZoneServer.Instance.Conf.World;
+
+			var expRate = worldConf.ExpRate / 100.0;
+			var classExpRate = worldConf.ClassExpRate / 100.0;
+
+			if (this.IsBuffActive(BuffId.SuperExp))
+			{
+				expRate *= worldConf.BlueJackpotExpRate;
```
exectails, line 300: The default value of this option is 10000... maybe my math is off, but doesn't this mean you're increasing the EXP rate by 10,000,000%? That seems rather high, doesn't it?

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
 
-			foreach (var dropItemData in this.Data.Drops)
-			{
-				var dropChance = GetAdjustedDropRate(dropItemData);
+			var iterations = 1;
 
-				var dropSuccess = rnd.NextDouble() < dropChance / 100f;
-				if (!dropSuccess)
-					continue;
+			// these should never be simultaneously applied, but just in case we apply gold last
```
exectails, line 401: If you do something for a specific reason you might want to not only mention that you're doing it, but also why, because someone else, or you in 2 months time, might not be able to understand it without actually reading the code.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
+			if (this.IsBuffActive(BuffId.EliteMonsterBuff)) iterations = worldConf.EliteRolls;
+			if (this.IsBuffActive(BuffId.SuperDrop)) iterations = worldConf.SilverJackpotRolls;
+			if (this.IsBuffActive(BuffId.TwinkleBuff)) iterations = worldConf.GoldJackpotRolls;
 
-				if (!ZoneServer.Instance.Data.ItemDb.TryFind(dropItemData.ItemId, out var itemData))
+			for (int i = 0; i < iterations; i++)
```
exectails, line 406: Please use `var` instead of `int` here, as per our coding conventions.

exectails, line 406: I guess a roll is something of a drop multiplier then 🤔 Though you're also calling them "iterations". I feel like a more descriptive name would be good here. Even something not entirely accurate, like `SilverJackpotDropMultiplier`, might be easier to understand than "roll" or "interation" in my mind.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
 
-				if (dropItemData.ItemId == ItemId.Silver || dropItemData.ItemId == ItemId.Gold)
-				{
-					minAmount = Math.Max(1, (int)(minAmount * (ZoneServer.Instance.Conf.World.SilverDropAmount / 100f)));
-					maxAmount = Math.Max(minAmount, (int)(maxAmount * (ZoneServer.Instance.Conf.World.SilverDropAmount / 100f)));
-				}
+			var worldConf = ZoneServer.Instance.Conf.World;
 
-				dropItem.Amount = rnd.Next(minAmount, maxAmount + 1);
+			var spawnRateBoost = 1F;
```
exectails, line 462: Not a big issue, but we're using lower case `f`s for floats everywhere else, and for consistency we should stick with that.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
+
+			// Check to see if this monster can become elite
+			// Note that on official it uses the map level rather than the monster level
+			if (this.Level < worldConf.EliteMinLevel)
+				return;
+
+			if (redOrb)
+				spawnRateBoost = worldConf.RedOrbEliteRate / 100F;
+
+			if (rnd.NextDouble() * 100D < worldConf.EliteSpawnChance * spawnRateBoost)
```
exectails, line 492: The double notation (which should use a lower-case d) isn't necessary and could safely be omitted since it doesn't contribute anything to the code or the reader's ability to understand it.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
 			}
+
+			if (rnd.NextDouble() * 100D < worldConf.SilverJackpotSpawnChance * spawnRateBoost)
+			{
+				this.StartBuff(BuffId.SuperDrop, 1, 0, TimeSpan.Zero, this);
+				return;
+			}
+
+			// Check to see if this monster can become elite
+			// Note that on official it uses the map level rather than the monster level
```
exectails, line 485: Why not implement the official way then/make it configurable?

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
+			if (this.Level < worldConf.EliteMinLevel)
+				return;
+
+			if (redOrb)
```
exectails, line 489: It's commendable that you're always thinking about the customization potential and add the code for it, but I feel like this could be a little more flexible. Instead of essentially hardcoding a check for red orb monsters, which would only ever be useful for this one option, wouldn't be make more sense to let the caller pass in modified rates? This would decouple the code a little and open up the possibility of lowering or increasing the chances for other monster types.

Alternatively, another way to do this without hardcoding a relationship between these classes might be to simply let he monster figure out that it's a red orb monster, by setting a variable on it for example, and then apply the multipliers itself, removing the need for any special arguments.

src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -381,47 +394,132 @@ private void DropItems(Character killer)
...
+				properties.InvalidateAll();
+
+				properties.SetFloat(PropertyName.HP, properties.GetFloat(PropertyName.MHP));
+				properties.SetFloat(PropertyName.SP, properties.GetFloat(PropertyName.MSP));
+
+				if (worldConf.EliteAlwaysAggressive) this.Tendency = TendencyType.Aggressive;
+
+				// TODO: Also needs to be able to summon its minions and have its special attack
+
+				return;
```
exectails, line 520: This appears to be unnecessary.

exectails (changes requested):

src/Shared/Configuration/Files/World.cs
Terotrous: A roll is one pass through the the drop list.  So a Silver Monster iterates through its drop list 100 times (by default).  I'm open to another term for it if you have any suggestions, I couldn't come up with anything else that sounded better.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: Oops, that's an error, it should be dividing that by 100F.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: I added a bit more detail to that comment (the reason we check gold last is because gold gives the most rolls.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: Fixed.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: I fixed a bunch, hopefully I didn't miss any.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: Fixed.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: My main reason for thinking to do it this way is simply because it makes more sense.  If you take a red orb and summon a monster from it, it makes little sense that whether or not it could become elite depends on the map where you summon it.  I also did consider having a customization option for it, but it only really matters if EliteMinLevel is used, so we're basically setting two parameters that only matter if both are used.  I suppose we can just use the map level since if you set the EliteMinLevel to 0 (which I imagine would be the only other setting anyone would use) it does the same thing.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: I actually did have it that way originally, but I changed it to this approach because it simplified the code slightly.  If you think any other code might ever want to override these rates I can return to having the rates be passed in.

src/ZoneServer/World/Actors/Monsters/Mob.cs
Terotrous: The return statement?  You're right about that, I removed it.

src/ZoneServer/World/Actors/Monsters/Mob.cs
exectails: > it makes more sense

Well, I get that, but do keep in mind that we're developing an _emulator_^^ The default behavior should always be the official behavior if possible.

Terotrous: I noticed while trying to make it use the map that I don't think we actually currently have the maplevel, it's not even in maps.txt.

src/Shared/Configuration/Files/World.cs
exectails: Personally I like multiplier better, simply because it's easier to understand, even if it wouldn't be as accurate. But yea, I don't have a suggestion for an actual, objectively better name. And once you know what it actually does, roll makes sense. Unfortunately most users will probably never know how it works^^"

Terotrous: Well, if it helps, I put more detail about what it means into the conf file.  I think it's at least reasonably clear what it does now.

PR #254, opened 2023-08-11
Adds Collection Support

Terotrous:
First commit for Collections.  Thanks to SalmanTKhan for significant help.  Does not include rewards yet, and only the chest in siauliai west is added, though all collections should work if you spawn the item.


src/ZoneServer/Database/ZoneDb.cs
```diff
@@ -1205,5 +1207,113 @@ private void LoadQuests(Character character)
...
+							var collectionId = reader.GetInt32("collectionId");
+							var itemId = reader.GetInt32("itemId");
+
+							if (collectionId != previousCollection)
+							{
+								// we're starting a new collection, save the previous one
+
+								if (previousCollection != -1)
+								{
+									character.Collections.AddSilent(previousCollection, currentItemList);
```
exectails, line 1295: Wouldn't it be easier to add the items one by one instead of building up a list?

src/ZoneServer/Database/ZoneDb.cs
```diff
@@ -1205,5 +1207,113 @@ private void LoadQuests(Character character)
...
+								cmd.Set("collectionId", collectionId);
+								cmd.Set("itemId", collectionItem);
+
+								cmd.Execute();
+							}
+						}
+					}
+					else
+					{
+						// we insert a collection with a blank item if you have the collection
```
exectails, line 1249: Why is this necessary? Can't the collection just be empty? Especially seeing how you just ignore the -1 entries in loading anyway.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
```diff
@@ -0,0 +1,165 @@
...
+							// First need to check how many times this item appears in the collection
+							var itemsNeeded = collectionData.RequiredItems.Count(a => a.Equals(item.Data.ClassName));
+							var itemsAddedAlready = collectionProgress.Count(a => a == item.Id);
+
+							if (itemsNeeded > itemsAddedAlready)
+							{ 
+								collectionProgress.Add(item.Data.Id);
+								if (collectionProgress.Count == collectionData.RequiredItems.Count)
+								{
+									// TODO: Give the award here and re-calculate stats
```
exectails, line 156: Is there a hurdle with adding the rewards?

system/scripts/items/collections.cs
```diff
@@ -0,0 +1,53 @@
...
+	[ScriptableFunction]
+	public ItemUseResult SCR_USE_COLLECTION_GET(Character character, Item item, string strArg, float numArg1, float numArg2)
+	{
+		var collectionName = strArg;
+
+		if (ZoneServer.Instance.Data.CollectionDb.Lookup(strArg, out var collectionData))
+		{
+			if (character.Collections.Add(collectionData.Id))
+			{
+				Send.ZC_SYSTEM_MSG(character, 119006);
```
exectails, line 27: You might want to use `Character.SystemMessage` instead, which takes the class name of the message instead of the id, which is more future proof.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
```diff
@@ -0,0 +1,165 @@
...
+{
+	/// <summary>
+	/// A character's collection manager.
+	/// </summary>
+	/// <remarks>
+	/// An important thing to note about the collection system is that Collections are account-level.
+	/// </remarks>
+	public class CollectionComponent : CharacterComponent
+	{
+		private readonly static TimeSpan AutoReceiveDelay = TimeSpan.FromMinutes(1);
```
exectails, line 25: This isn't necessary for collections, right?

exectails (changes requested):

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
Terotrous: You're right, that's a holdover from a component I used as a baseline, it's been taken out.

system/scripts/items/collections.cs
Terotrous: Good catch on that, I've updated it.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
Terotrous: Is there already a system in place to add bonus stats to a character?  I didn't see it when I looked for it and I felt that creating such a system was outside the scope of this PR.

I think I would prefer to handle this in a separate PR in any case, since the code to add the rewards could be substantial.

src/ZoneServer/Database/ZoneDb.cs
exectails: The loading code would definitely be simpler though, and hence easier to write and read^^ But you do have a point, even though locking in this context is entirely inconsequential, because the lock isn't being contented.

But I also just realized that you're storing the list that you create here inside the collection component... That seems risky to me, since you never know what a user might do with that. While _your_ code is fine, you're giving users access to the internals of the component. Someone might pass in a list that they then keep using for other things, which may have unintended effects. Someone might also pass a list to `AddSilent` with the expectation that the items in the list are _added_ to the collection, when in reality that call will overwrite whatever was there before--something that isn't documented at all. I would highly recommend changing it to add the contents of the list to the internal list.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
```diff
@@ -0,0 +1,162 @@
...
+		/// <summary>
+		/// Gets a collection and returns the list of items registered to it.
+		/// </summary>
+		/// <param name="collectionId"></param>
+		/// <param name="registeredItems"></param>
+		/// <returns></returns>
+		public bool TryGet(int collectionId, out List<int> registeredItems)
+		{
+			lock (_syncLock)
+				return _collectionProgress.TryGetValue(collectionId, out registeredItems);
```
exectails, line 91: You're also directly returning the internal list, which lets users modify it. Again, this is undocumented and might lead to unintended side effects. We also don't do this anywhere else in Melia. Unless you have a good reason to do it, and you document it, you should rather return a copy.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		/// <summary>
+		/// Returns the total number of registered collections
+		/// </summary>
+		/// <returns></returns>
+		public int Count()
```
exectails, line 98: Simple methods like this that only return a "property" of the component should not be a full blown method, but rather a property.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
```diff
@@ -0,0 +1,162 @@
...
+		/// <summary>
+		/// Registers an item to this collection
+		/// <param name="collectionId"></param>
+		/// <param name="item"></param>
+		/// </summary>
+		public bool RegisterItem(int collectionId, Item item)
+		{
+			lock (_syncLock)
+			{
+				if (_collectionList.TryGetValue(collectionId, out var collectionData))
```
exectails, line 140: The nested ifs in this area are a bit difficult to read. It would arguably be clearer if you employed a structure where you check for invalid states first and put the success at the end.

Instead of this:
```text
if (a)
{
	if (b)
	{
		if (c)
		{
			if (d)
			{
				return true;
			}
		}
	}
}

return false;
```
You would do this:
```text

if (!a)
	return false;

if (!b)
	return false;

if (!c)
	return false;

if (!d)
	return false;

return true;
```

This makes it much easier to read, understand, and maintain. Also, you should not omit the curly braces for an `if` if the conditional code is longer than one line, as per our coding conventions.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
exectails: I'm not familiar with what exactly the rewards are, but granting bonus stats might be as simple as just calling modify on the stat you want to increase, which would fairly trivial.

src/ZoneServer/World/Actors/Characters/Components/InventoryComponent.cs
```diff
@@ -597,6 +597,8 @@ public InventoryResult Remove(Item item, int amount, InventoryItemRemoveMsg msg)
 
 				Send.ZC_ITEM_REMOVE(this.Character, item.ObjectId, amount, msg, InventoryType.Inventory);
 
+				Send.ZC_ITEM_INVENTORY_DIVISION_LIST(this.Character);
```
exectails, line 600: Is a full refresh of the inventory actually necessary? We're only decreasing an item's amount here, and a refresh is a relatively expensive operation.

exectails (changes requested):

src/ZoneServer/Database/ZoneDb.cs
Terotrous: Oh, the way it works is that you have to find a hidden treasure chest on each map which gives you an item that unlocks the collection.  The item gets used up, so you'd want the game to remember if you have the collection unlocked, even if it's empty.

Using a separate table is absolutely possible, though I don't think it helps.  The game forces us to send every item in a collection even if it's finished (this is how the game identifies a completed collection), so we still have to query every row in the collection contents table even if we have a separate collections table.  For efficiency, we would probably do this with an outer join (selecting one collection at a time in a loop leads to hundreds of select statements being executed for a single character), but this will now return null for the itemid on an empty collection, forcing us to check for dbNull on every record again.  Overall I think the current implementation is the most optimal.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
Terotrous: There's a bunch of different reward types.

The most simple are derived stats, like physical attack +5.  This is indeed probably fairly trivial.  The only real issue is that if there's any situation where the game recalculates your derived stats from scratch, we'd have to re-add these bonus stats again (probably needs to be a method like "Collections.AddBonusStats" or something like that that a method could call).

You can also get base stats, like STR and INT.  I notice that in the interface, there are 3 categories shown for these stats, base, investment, and other.  This definitely needs to go in "Other".  Does a method of adding to other currently exist?  (If so, equipment probably also uses it).

The third reward type is items.  This is also pretty easy.  When the collection is finished (at the exact time of entering the final item), it gives you the item(s).  When calling the method to get all of your existing bonuses, these ones are ignored, so it doesn't give you the items over and over.

The fourth type is account properties.  It seems like the only one of these is maximum team storage size.  I'm not sure if this is implemented yet?  Might need a separate method from the others to get this.

In any case, it's complex enough that I feel like we might want to save this for another PR.

src/ZoneServer/Database/ZoneDb.cs
exectails: I'll have to take another look at this, but not using a dedicated collections table that connects to a list of items per collection seems like a recipe for refactoring, because the moment we need to save _any_ additional information about a collection, such as a date when it was unlocked for example, we have to redo everything. You're presumably right that it won't give us any boosts in performance, or make the code cleaner in some significant way, but maintainability should always be a concern.

You might've also noticed that the dummy entry implementation is not ideal for conveying intent. Someone who isn't familiar with how exactly the system works will have to dig around the code and database and make several assumptions to come to the conclusion that the dummy actually is a dummy, that's only used to remember a collection is open.

src/ZoneServer/World/Actors/Characters/Components/CollectionComponent.cs
exectails: When it comes to bonuses and rewards I'm of the opinion that a script will usually yield the best results, to make the system flexible for any new official additions, as well as users' customization ideas. The setup is a bit more involved than setting a type and one or two parameters, but in turn you can do pretty much anything with it. Would that seem like a good approach in this case?

I think adding the base system to accomplish granting rewards should ideally be in place before merge, even if not every single reward type is implemented yet.

PR #255, opened 2023-12-15
Highlander Tier 1

Terotrous:
Adds the three skills for Highlander's first tier.  Note the correction to skills.txt.


src/ZoneServer/Skills/Handlers/Highlander/WagonWheel.cs
```diff
@@ -0,0 +1,87 @@
...
+			{
+				var skillHitResult = SCR_SkillHit(caster, target, skill);
+				target.TakeDamage(skillHitResult.Damage, caster);
+
+				var skillHit = new SkillHitInfo(caster, target, skill, skillHitResult, damageDelay, skillHitDelay);
+				skillHit.ForceId = ForceId.GetNew();
+
+				skillHit.KnockBackInfo = new KnockBackInfo(caster.Position, target.Position, skill);
+				skillHit.HitInfo.Type = skill.Data.KnockDownHitType;
+				target.Components.Get<MovementComponent>().Stop();  // Temporary workaround, monsters that get knocked back should not be able to move for a while
```
exectails, line 78: I get the idea of wanting to create a hotfix for a problem, but is this really much of a workaround? Whether stopping their movement does anything at all depends entirely on the AI, and they might start walking again right away. Adding temporary code for specific skills might just end up polluting the skill handlers. There should really be proper handling it instead.

Additionally, this code will throw an exception if a target doesn't have a movement component for some reason.

system/db/skills.txt
```diff
@@ -200,7 +200,7 @@
...
+{ skillId: 10207, className: "Highlander_SkyLiner", name: "Skyliner", useType: "MeleeGround", attackType: "Slash", attribute: "Melee", classType: "Melee", maxLevel: 1, basicSp: 54, enableAngle: 90, maxRange: 100, waveLength: 40, splashType: "Square", splashRange: 13, splashHeight: 0, splashAngle: 0, splashRate: 0, factor: 297.2, factorByLevel: 49.5, atkAdd: 0, atkAddByLevel: 0, defaultHitDelay: 0, deadHitDelay: 50, shootTime: 9999, delayTime: 0, cancelTime: 9999, hitTime: [200], holdTime: [400], speedRate: 1, speedRateAffectedByDex: true, speedRateAffectedByBuff: false, enableCastMove: false, reinforceAbility: "Highlander29", hiddenReinforceAbility: "Highlander40", hiddenReinforceAbilityFactorByLevel: "1.25", cooldownGroup: "SkyLiner", cooldownTime: 15000, overheatGroup: "SkyLiner_OH", overheatCount: "0", overheatDelay: "15000" },
```
exectails, line 203: Have you confirmed that this is actually the overheat group sent for this skill via `ZC_OVERHEAT_CHANGED`? While it does seem like it probably should be `SkyLiner_OH`, the official skill data says it's `SkyLiner`.

exectails (changes requested):

src/ZoneServer/Skills/Combat/KnockBackInfo.cs
```diff
@@ -103,6 +103,15 @@ public KnockBackInfo(Position attackerPosition, Position targetPosition, HitType
 				this.Time = TimeSpan.FromMilliseconds(6747);
 				this.ToPosition = this.FromPosition.GetRelative(this.Direction, distance);
 			}
+			else if (this.Velocity == 250 && this.VAngle == 85) // Wagon Wheel
+			{
+				var distance = 30.570992087511f;
+
+				this.Time = TimeSpan.FromMilliseconds(6747);
+				var finalPosition = this.FromPosition.GetRelative(this.Direction, distance);
+				finalPosition.Z = finalPosition.Z + 40;
```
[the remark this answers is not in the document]
Terotrous: You're right, that was a mistake.  I thought I had to add the vertical component of the attack manually (which would have been the Y coordinate), but you don't, the angle handles that.  It's been removed.

src/ZoneServer/Skills/Handlers/Highlander/WagonWheel.cs
Terotrous: You're right, and this doesn't really even work.  This definitely does need to be fixed elsewhere in code (being hit with a knockdown attack should presumably make you unable to do anything at all until you get up), but it's not the place to do it here.  I've removed it.

system/db/skills.txt
Terotrous: The only confirmation I can give is that if it is set to just Skyliner, the skill doesn't go on cooldown when it is used, but with Skyliner_OH it does.

src/ZoneServer/Skills/Handlers/Highlander/CrossCut.cs
```diff
@@ -0,0 +1,98 @@
...
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			var splashParam = skill.GetSplashParameters(caster, originPos, farPos, length: 50, width: 100, angle: 90);
+			var splashArea = skill.GetSplashArea(SplashType.Fan, splashParam); // DB states that this should be Square
```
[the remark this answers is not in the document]
Terotrous: This is noted, though in this particular case this line has been changed and the comment is no longer there.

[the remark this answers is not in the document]
Terotrous: I was mainly using fan because from my initial testing, Square splash areas were working strangely, but after fiddling with them a bit more I think I got it working reasonably, so I've switched it to Square as the skill database indicates.  It's worth noting though that the size of the area is not specified at all in skill-db, so I've tweaked the values as best I can to match the animation.

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
```diff
@@ -0,0 +1,130 @@
...
+namespace Melia.Zone.Skills.Handlers.Highlander
+{
+	/// <summary>
+	/// Handler for the Highlander skill SkyLiner.
+	/// </summary>
+	[SkillHandler(SkillId.Highlander_SkyLiner)]
+	public class SkyLiner : IGroundSkillHandler, IDynamicCasted
+	{
+		private const int TotalHits = 4;
+		private bool keepAttacking = false;
```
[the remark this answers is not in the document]
Terotrous: This is a good catch.  I've rewritten it to use the Character variables.  This does force a cast to Character (since ICombatEntity doesn't have variables), so if this skill is used by any other type of entity, they can't repeat it.  I'm not sure if this is relevant.

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
```diff
@@ -0,0 +1,130 @@
...
+		public void StartDynamicCast(Skill skill, ICombatEntity caster)
+		{			
+			keepAttacking = true;
+			Send.ZC_PLAY_SOUND(caster, "voice_archer_multishot_cast"); // Probably not the right voice clip
```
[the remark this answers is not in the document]
Terotrous: I set it to use the first one, I'm not quite sure why two are listed.  I'm actually not convinced the sound plays anyway, it seems to sound the same as it did before.

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
```diff
@@ -0,0 +1,130 @@
...
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			var splashParam = skill.GetSplashParameters(caster, originPos, farPos, length: 50, width: 100, angle: 90);
+			var splashArea = skill.GetSplashArea(SplashType.Fan, splashParam); // DB states that this should be Square
```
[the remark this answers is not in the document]
Terotrous: I've changed them all to be square.

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
```diff
@@ -0,0 +1,130 @@
...
+				var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+				foreach (var target in targets.LimitBySDR(caster, skill))
+				{
+					var skillHitResult = SCR_SkillHit(caster, target, skill);
+					target.TakeDamage(skillHitResult.Damage, caster);
+
+					var skillHit = new SkillHitInfo(caster, target, skill, skillHitResult, damageDelay1, skillHitDelay);
+					skillHit.HitEffect = HitEffect.Impact;
+					hits.Add(skillHit);
```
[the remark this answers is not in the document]
Terotrous: Yep, this is wrong, the send should be outside the foreach statement.

src/ZoneServer/Skills/Handlers/Highlander/CrossCut.cs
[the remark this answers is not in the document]
exectails: Our current understanding is that skill_bytool is the data set we want to reference for skill hit areas, as the data has been reliable several times already. According to the data for Cross Cut, the length should be 30 and the width 45. Note that 45 is half the width, which extends in both directions for a total of 90 internally. Do you think these parameters could work?

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
```diff
@@ -0,0 +1,139 @@
...
+	/// Handler for the Highlander skill SkyLiner.
+	/// </summary>
+	[SkillHandler(SkillId.Highlander_SkyLiner)]
+	public class SkyLiner : IGroundSkillHandler, IDynamicCasted
+	{
+		private const int TotalHits = 4;
+
+		public void StartDynamicCast(Skill skill, ICombatEntity caster)
+		{
+			if (caster is Character character)
```
exectails, line 29: Based on the name of this variable and how you're setting it, it seems like it might be better suited for a general purpose variable that would be set from the packet handlers. Is this a Sky Liner specific mechanic? If so, the variable name should reflect that. If not, we should probably look into establishing a clearer way to use it.

Terotrous: It's not skyliner specific, I think every dynamic cast skill would use this.  I don't quite know where you're suggesting you want to add it, we can probably discuss this more in the development chat.

exectails: If this is a general function of dynamically casted skills, we don't want to repeat this in every single dynamic skill handler. Instead, the packet handlers could set the variable and the skill handler could run a proper loop while the variable is set. Ideally this variable would also be an actual property, rather than a character variable, to "legitimize" it and make it easier and clearer what it is. And seeing how it would be a core part of combat, it should most likely be part of ICombatEntity.

src/ZoneServer/Skills/Handlers/Highlander/CrossCut.cs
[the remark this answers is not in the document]
Terotrous: I've switched it to the new parameters, they seem close enough.

src/ZoneServer/Skills/Handlers/Highlander/SkyLiner.cs
Terotrous: I've added it to the CombatComponent, similar to the AttackState.  I specifically tested to see if the client would send the "endDynamicCasting" packet even if you hold it to the maximum length (at which point it ends automatically) and it still does, so setting the casting state to false in endDynamicCasting is safe.

src/ZoneServer/Skills/Handlers/Highlander/CrossCut.cs
```diff
@@ -0,0 +1,106 @@
...
+		/// <param name="splashArea"></param>
+		private async void Attack(Skill skill, ICombatEntity caster, ISplashArea splashArea)
+		{
+			var hitDelay = TimeSpan.FromMilliseconds(30);
+			var damageDelay1 = TimeSpan.FromMilliseconds(50);
+			var damageDelay2 = TimeSpan.FromMilliseconds(80);
+			var delayBetweenHits = TimeSpan.FromMilliseconds(330);
+			var skillHitDelay = TimeSpan.Zero;
+
+			var pc = skill.Owner;
```
exectails, line 67: Is there a specific reason why you're using the skill owner instead of the caster here?

Terotrous: Haha, it's actually because you wrote that code snippet for me and I just pasted it verbatim.
https://discord.com/channels/273278107773960192/1128408008469266503/1185313478928506930

You're right though, it should be caster.  I tested it with caster and it still works fine.

PR #256, opened 2024-04-16
feat: ✨ Add new branch to work with Docker

gustavobarbosa-web:
Hi, I'm working in this project's Docker support. Hope anyone can better test it's compatibility with current client. Had to make a few changes in server files so that it's compatible with Linux + works with docker initialization. It would be better to keep this in a separate Docker branch for now. Thanks!


exectails:
Thanks for sharing your work. We wouldn't want to merge this into the main branch, but keeping it in a seperate one might be a nice idea. Though there are a couple changes I'm personally not happy with, such as merging all SQL files or removing the console command code entirely. I feel like we'd want to keep as much of the code the same as possible, without major differences between versions/branches.

gustavobarbosa-web:
> Thanks for sharing your work.

Thank you for your project, dedication and for taking the time to review this fork.
>We wouldn't want to merge this into the main branch, but keeping it in a seperate one might be a nice idea. 

Yeah, I think it's quite a change of scope to be merged straight to master.

> Though there are a couple changes I'm personally not happy with, such as merging all SQL files or removing the console command code entirely. 

I agree, I did the SQL merge because it's easier to setup the Docker entrypoint with one file, but there was no good reason to remove the original files. The console commands shouldn't have been removed either, so I've added a feel checks to skip incompatible commands on incompatible systems whille keeping it working as intented on Windows baremetal OS.

> I feel like we'd want to keep as much of the code the same as possible, without major differences between versions/branches.

Also agreed. I'll try to keep up!

stop
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+clear
```
exectails, line 2: Might just be, but I think I'd be a bit annoyed if the stop script cleared my screen every time. What if there was still something I needed?

system/conf/database.conf
```diff
@@ -2,9 +2,9 @@
 // Configuration file
 //----------------------------------------------------------------------------
 
-host     : 127.0.0.1
+host     : melia-database
 user     : root
-pass     : 
+pass     : 123456
```
exectails, line 7: Same as with the PHP options, you don't need to modify the system configs if you just create the respective user conf files.

exectails:
I believe if those couple things are improved upon, this is basically fine to be merged into a branch, or even master, if it manages to be general purpose enough.

y4my4my4m:
fyi, under windows, i had to add vim to the packges of the dockerfile and do `:set ff=unix` to the `start` file because it was copying the file to the docker with non-unix CLRF file ending

y4my4my4m:
https://github.com/gmargriff/melia-docker/pull/4
I've pushed this to @gmargriff 's branch (because im using the docker version for convenience) but thought the docker ver should get merged in the main repo too. If the docker doesnt get merged, maybe you'd still like to add my changes.

src/Shared/Configuration/Files/Web.cs
```diff
@@ -17,8 +18,12 @@ public class WebConfFile : ConfFile
 		public void Load(string filePath)
 		{
 			this.Include(filePath);
-
-			this.PhpCgiFilePath = this.GetString("php_cgi_bin", Path.Combine("user", "tools", "php", "php-cgi.exe"));
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
exectails, line 21: This won't really change anything, because the configuration value is read from the conf file regardless of the operating system. The only way you will get the usr bin path from this is if the conf option is missing.

src/ZoneServer/Commands/ChatCommands.Handlers.cs
```diff
@@ -63,7 +63,21 @@ public ChatCommands()
 			this.Add("item", "<item id> [amount]", "Spawns item.", this.HandleItem);
 			this.Add("silver", "<modifier>", "Spawns silver.", this.HandleSilver);
 			this.Add("spawn", "<monster id|class name> [amount=1] ['ai'=BasicMonster] ['tendency'=peaceful] ['hp'=amount]", "Spawns monster.", this.HandleSpawn);
-			this.Add("madhatter", "", "Spawns all headgears.", this.HandleGetAllHats);
+
+			this.Add("madhatter", "", "Spawns all hair accessories.", this.HandleGetAllHairAccessories);
```
exectails, line 67: This isn't related to the docker PR, is it? 🤔

src/ZoneServer/ZoneServer.cs
```diff
@@ -101,7 +101,11 @@ public override void Run(string[] args)
 			var title = string.Format("Zone ({0}, {1})", groupId, serverId);
 
 			ConsoleUtil.WriteHeader(ConsoleHeader.ProjectName, title, ConsoleColor.DarkGreen, ConsoleHeader.Logo, ConsoleHeader.Credits);
-			ConsoleUtil.LoadingTitle();
+			
+			// Skip the following command on incompatible systems
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
exectails, line 106: We since fixed the problem with `LoadingTitle` under Linux, and this should now work without any exceptions made in the code.

src/ZoneServer/ZoneServer.cs
```diff
@@ -124,8 +128,21 @@ public override void Run(string[] args)
 			this.StartCommunicator();
 			this.StartAcceptor();
 
-			ConsoleUtil.RunningTitle();
-			new ConsoleCommands().Wait();
+
+			// If running on Windows system, enable console inputs
+			// otherwise, start an event loop to keep server running until stopped
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
exectails, line 134: This seems like a suboptimal thing to do, because if I'm running the server in a console under Linux, I might want console commands. In the same vein, I might run the server without a console on Windows. It seems to me like a check for an interactive environment would be more appropriate.

We actually have a property for that on `ConsoleUtil`, "`IsUserInteractive`", though I'll admit this property hasn't been used or confirmed to still be working in a while.

system/conf/web.conf
```diff
@@ -5,6 +5,6 @@
 // Path to the PHP CGI executable
 // If the file doesn't exist, and the configured path is in user/tools/,
 // PHP is downloaded automatically on Windows.
-php_cgi_bin: user/tools/php/php-cgi.exe
+php_cgi_exe: /usr/bin/php-cgi
```
exectails, line 8: You seem to have reverted a recent change here, where the name of the option was fixed. It was supposed to be `_bin`. Additionally, it would be better if, instead of changing default values in conf files in the system folder, you would dynamically make your changes in the user folder for Docker.

user/conf/database-example-for-docker.conf
```diff
@@ -0,0 +1,8 @@
+// Melia
```
exectails, line 1: Do we need an example db conf file for Docker? Will this not be created during the setup process?

README.md
```diff
@@ -41,7 +35,116 @@ Specifically, some of the major features that are working are as follows:
 - Monster spawns
 - Quests
 
-Requirements
+Installation with Docker
```
exectails, line 38: I'd prefer you revert the changes to the readme, which is supposed to be both concise and general purpose. Long explanations about the setup are better suited for the docs folder.

sql/updates/update_2024-08-11_1.sql
```diff
@@ -1,18 +1,17 @@
-CREATE TABLE `character_etc_properties` (
-  `propertyId` bigint(20) NOT NULL,
+CREATE TABLE IF NOT EXISTS `character_etc_properties` (
```
exectails, line 1: Can you tell me what this change is about? I feel like we'd want this to fail if the table already exists, because that's an indicator for something not being right.

exectails (changes requested):
Thanks for the continued work on this PR. I'd like to merge it in the near future, but unfortunately there are still some issues and potential improvements, as well as comments that weren't addressed yet.

gmargriff:
Hey, @exectails and @y4my4my4m , thanks for your work in this PR, I've done just enough changes to make autosync available again on Github, but I'm not directly working on this right now since my PC has been down since may/june, just changed the PR to draft to avoid confusion. Through my phone I've been just keeping the repo synced. I've read through the comments here and it all seems reasonable, as soon as I can get back I'll be working on this. Thanks again!

PR #260, opened 2024-06-08
Peltasta Tier 1

Terotrous:
Adds the Tier 1 skills for Peltasta.  There's a fair bit of complexity here as Peltasta's skills have a number of interactions with each other.  For example, enemies that are swashbuckled take 10% more damage from Peltasta skills.

As such, I've introduced the concept of skill modifiers, which is a class that you can create in your skill handler to assign some properties that influence damage calculation.  A fair number of other skills are currently calculating damage somewhat incorrectly and should eventually be adapted to use this system once it is implemented (one example would be Smite, which multiplies the final damage by 1.5 if the target has a certain race, this calculation should be occurring before defense calculation, as this effectively multiplies the effect of enemy defense by 1.5).  Later skills will probably need more properties than this, but they can easily be added to the class without influencing existing skills.


src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
```diff
@@ -0,0 +1,213 @@
...
+			List<ICombatEntity> hitTargets = new List<ICombatEntity>();
+			var sr = (int)skill.Properties.GetFloat(PropertyName.SkillSR);
+			if (ZoneServer.Instance.Conf.World.DisableSDR)
+				sr = int.MaxValue;
+
+			while (!cancelToken.IsCancellationRequested)
+			{
+				var splashArea = new Circle(shieldMonster.Position, 40);
+
+				var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
```
exectails, line 144: Would this skill benefit from `Mobs` supporting trigger areas? It seems like that would be easier than running a custom collision check.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
```diff
@@ -0,0 +1,213 @@
...
+		/// <param name="caster"></param>
+		/// <param name="targets"></param>
+		private void Attack(Skill skill, ICombatEntity caster, List<ICombatEntity> targets)
+		{
+			var damageDelay = TimeSpan.FromMilliseconds(100);
+			var skillHitDelay = TimeSpan.Zero;
+
+			var bonusPatk = 0f;
+
+			if (caster.IsBuffActive(BuffId.HighGuard_Abil_Buff) && caster is Character character)
```
exectails, line 178: Again, `TryGetBuff` may simplify this code.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/SwashBuckling.cs
```diff
@@ -0,0 +1,70 @@
...
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, caster.Position, null);
+
+			// This buff is mentioned in the skill file, but it seems like it was tied to abilities that have been removed
```
exectails, line 42: I would argue that we should just add the respective checks and buff applications instead of only commenting about them. If the ability doesn't appear in the wild anymore that's fine, but if someone wanted to add it, it will immediately work, which is always nice.

src/ZoneServer/World/WorldManager.cs
```diff
@@ -24,6 +24,12 @@ public class WorldManager
 		// Unique handles for buffs
 		private int _buffhandles = 0;
 
+		// Unique handles for skills
```
exectails, line 27: I assume you tested this and it worked, but just to be sure, do these two use their own id ranges or should they perhaps be part of the entities?

exectails (changes requested):

src/Shared/Network/NormalOp.cs
```diff
@@ -44,11 +44,14 @@ public static class Zone
 			public const int AccountProperties = 0x4D;
 			public const int UnkDynamicCastStart = 0x4F;
 			public const int UnkDynamicCastEnd = 0x50;
-			public const int Skill_59 = 0x59;
+			public const int SkillRunScript = 0x59;
```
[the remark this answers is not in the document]
Terotrous: Just verified on latest steam version, still up to date.

src/ZoneServer/Buffs/Handlers/Swordsman/Peltasta/SwashBuckling_Debuff.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+
+namespace Melia.Zone.Buffs.Handlers.Swordsman.Peltasta
+{
+	/// <summary>
+	/// Handle for the Swashbuckling Debuff
+	/// Note that the ingame tooltip implies that this reduction in Block / Dodge rate is only to attacks from
```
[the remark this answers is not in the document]
Terotrous: I've moved it into the class itself.

src/ZoneServer/Network/Send.Normal.cs
```diff
@@ -358,6 +359,127 @@ public static void UnkDynamicCastEnd(Character character, SkillId skillId, float
...
+			/// <param name="caster"></param>
+			/// <param name="skill"></param>
+			/// <param name="packetStringId"></param>
+			/// <param name="position"></param>
+			/// <param name="direction"></param>
+			/// <param name="f1"></param>
+			/// <param name="f2"></param>
+			/// <param name="skillHandle"></param>
+			/// <param name="f3"></param>
+			public static void SkillPad(ICombatEntity caster, Skill skill, int packetStringId,
```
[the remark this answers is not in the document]
Terotrous: I'm going to be honest with you, I copied this code from Akinue's implementation.  I think you're saying you don't want the overload where you pass in the id instead of the packetstring?  I've removed it.

src/ZoneServer/Network/Send.Normal.cs
```diff
@@ -358,6 +359,127 @@ public static void UnkDynamicCastEnd(Character character, SkillId skillId, float
...
+			public static void SkillPad(ICombatEntity caster, Skill skill, string animationName,
+				Position position, Direction direction, float f1, float f2, int skillHandle,
+				float f3, bool isVisible = true)
+			{
+				if (ZoneServer.Instance.Data.PacketStringDb.TryFind(animationName, out var packetString))
+					SkillPad(caster, skill, packetString.Id, position, direction, f1, f2, skillHandle, f3, isVisible);
+			}
+
+			/// <summary>
+			/// Creates a pad in client (MONSKL_CRE_PAD)
```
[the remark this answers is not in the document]
Terotrous: I also expanded on this a bit.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -214,6 +214,14 @@ private void IncreaseNearbyHate(TimeSpan elapsed, IEnumerable<ICombatEntity> pot
 		/// <param name="amount"></param>
 		protected void IncreaseHate(ICombatEntity entity, float amount)
 		{
+			// This debuff halts all hate accumulation from anything other than the caster, unless the caster is gone
+			if (this.Entity.IsBuffActive(BuffId.ProvocationImmunity_Debuff))
```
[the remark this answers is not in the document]
Terotrous: Good call, I've moved it to CanAccumulateHate.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
```diff
@@ -0,0 +1,213 @@
...
+				var shield = character.Inventory.GetItem(EquipSlot.LeftHand).Data;
+				Send.ZC_NORMAL.SetPadModel(shieldMonster, "warrior_f_", shield.Id);
+			}
+			Send.ZC_NORMAL.Skill_ItemRotate(shieldMonster, 90);
+			Send.ZC_NORMAL.SpinObject(shieldMonster);
+			Send.ZC_NORMAL.Skill_SetActorHeight(shieldMonster, skillHandle, 22);
+			Send.ZC_NORMAL.AttachEffect(shieldMonster, "I_light004_violet", 1.5f);
+
+			// Begin the shield throw, throwing shield outwards to the far point
+			var movement = shieldMonster.Components.Get<MovementComponent>();
```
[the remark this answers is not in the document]
Terotrous: This one was a challenge to be sure.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
Terotrous: It's been replaced with TryGetBuff, this was pretty old code.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
Terotrous: I don't know, I like that this is somewhat similar to how most skills do their collission check.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/SwashBuckling.cs
Terotrous: I would probably have done so, but at the time I was not sure what these buffs did or what ability applied them.  When I go back to add the rest of Peltasta's other skills I'll probably try to find that information, I've been adding it to my more recent skill PRs.

src/ZoneServer/World/WorldManager.cs
Terotrous: I don't know the answer to this one.

SalmanTKhan: The correct answer is most likely part of entities. What I thought were skill handles turned out to be "pad" handles, but since our pads are NPCs with triggers, you can probably avoid using them.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -358,6 +375,14 @@ protected ICombatEntity GetMostHated()
 		/// <returns></returns>
 		protected bool IsHating(ICombatEntity entity)
 		{
+			// Always hating the person that casted this buff
+			if (this.Entity.IsBuffActive(BuffId.ProvocationImmunity_Debuff))
```
exectails, line 379: I hope I don't need to mention that there are other cases still where you might want to use `TryGetBuff` for more concise code^^

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
exectails: What do you mean by that? For clarification, trigger areas are essentially just shapes that move around the map with the entity, and an event is called when they're triggered, aka if an entity enters them. That way you could just let em rip and react once something happens.

Terotrous: I mean the collision here is handled by defining a splashArea and checking for targets within that splasharea, just like most other skills.  I'm sure it could be done with events, I'm just not sure it would be simpler.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
exectails: Has it? I'm still seeing `IsBuffActive` and a component get on my side =o

Terotrous: It looks like the one I fixed was a separate instance where TryGetBuff was not used, I've also now fixed this one.

src/ZoneServer/World/WorldManager.cs
exectails: Do we know if pads are their own thing? They are mentioned time and time again in the client data, but I don't think I've seen anything yet that would clarify what they are exactly. Are they monsters as well? A different type of actor?

If these skill handles, that I've also been wondering about, are pad handles, I would agree that they're probably some form of entity and should go into that pool. Or rather, the handle will just be the handle of the Monster, right? And the effects probably too? I guess we're actually already handling those that way, like in Heal.

src/ZoneServer/Skills/Handlers/Swordsman/Peltasta/ShieldLob.cs
exectails: > not sure it would be simpler

Wanna do a bet?^^

Terotrous: Of course, if you'd like to recode parts of the skill you're welcome to, I just don't think the collision check is that complicated.

src/ZoneServer/Scripting/AI/AiScript.cs
Terotrous: I also removed this one.

exectails:
Going to close this PR because a version of it has been merged by now. Thanks for the contribution!

PR #261, opened 2024-06-21
Block and critical formulas implementation

kenedos:
This PR implements:

- [NEW] Block Rate
- [NEW] Block Penetration
- Block Chance Formula
- Block damage reduction (50% less damage)
- [NEW] Critical Defense Rate
- Improved Critical Chance Formula
- Improved Critical Damage to deal 1.5x damage
- Adds Block/BlockPen and Critical Attack/Rate/Defense stats for monsters

This PR *DOES NOT* implement:

- Active blocking (holding 'C' to block)
- Equipment block rates in items.txt database

Videos:

- Player blocking attacks: https://www.youtube.com/watch?v=tGhd0X7mmpM
- Monster blocking attacks: https://www.youtube.com/watch?v=IcUwUen4OZA

Based on: https://treeofsavior.com/page/news/view.php?n=951
Credits to @SalmanTKhan for base implementation

### DIFF
--- a/system/scripts/zone/core/calc_monster.cs
+++ b/system/scripts/zone/core/calc_monster.cs
@@ -334,4 +334,140 @@ public class MonsterCalculationsFunctionsScript : GeneralScript
 
 		return (int)Math.Max(1, value);
 	}
+
+	/// <summary>
+	/// Returns the monster's Critical Hit Rate.
+	/// </summary>
+	/// <param name="monster"></param>
+	/// <returns></returns>
+	[ScriptableFunction]
+	public float SCR_Get_MON_CRTHR(Mob monster)
+	{
+		if (monster.Properties.Overrides.TryGetFloat(PropertyName.CRTHR, out var fixValue))
+			return fixValue;
+		
+		var value = (float)monster.Data.CritHitRate;
+		
+		var byBuffs = 0f;
+		byBuffs += monster.Properties.GetFloat(PropertyName.CRTHR_BM, 0);
+		
+		var byRateBuffs = 0f;
+		byRateBuffs += monster.Properties.GetFloat(PropertyName.CRTHR_RATE_BM);
+		byRateBuffs = value * byRateBuffs;
+		
+		value = value + byBuffs + byRateBuffs;
+		
+		return (int)Math.Max(0, value);
+	}
+	
+	/// <summary>
+	/// Returns the monster's Critical Dodge Rate (Critical resistance).
+	/// </summary>
+	/// <param name="monster"></param>
+	/// <returns></returns>
+	[ScriptableFunction]
+	public float SCR_Get_MON_CRTDR(Mob monster)
+	{
+		if (monster.Properties.Overrides.TryGetFloat(PropertyName.CRTDR, out var fixValue))
+			return fixValue;
+		
+		var value = (float)monster.Data.CritDodgeRate;
+		
+		var byBuffs = 0f;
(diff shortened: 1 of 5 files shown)


system/scripts/zone/core/calc_monster.cs
```diff
@@ -334,4 +334,140 @@ public float SCR_Get_MON_SDR(Mob monster)
...
+		if (monster.Properties.Overrides.TryGetFloat(PropertyName.CRTATK, out var fixValue))
+			return fixValue;
+
+		var value = (float)monster.Data.CritAttack;
+		
+		var byBuffs = 0f;
+		byBuffs += monster.Properties.GetFloat(PropertyName.CRTATK_BM, 0);
+		
+		var byRateBuffs = 0f;
+		//byRateBuffs += monster.Properties.GetFloat(PropertyName.CRTATK_RATE_BM);
```
exectails, line 412: This property doesn't appear to exist. Could these lines not be removed then?

system/scripts/zone/core/calc_monster.cs
```diff
@@ -334,4 +334,140 @@ public float SCR_Get_MON_SDR(Mob monster)
...
+		
+		var properties = monster.Properties;
+		
+		var baseValue = (float)monster.Data.BlockRate;
+		var value = baseValue;
+		
+		var byBuffs = properties.GetFloat(PropertyName.BLK_BM);
+		
+		var byRateBuffs = 0f;
+		byRateBuffs += properties.GetFloat(PropertyName.BLK_RATE_BM);
```
exectails, line 439: Same as the other one, this property does not seem to exist for monsters.

system/scripts/zone/core/calc_monster.cs
```diff
@@ -334,4 +334,140 @@ public float SCR_Get_MON_SDR(Mob monster)
...
+		
+		var properties = monster.Properties;
+		
+		var baseValue = (float)monster.Data.BlockBreakRate;
+		var value = baseValue;
+		
+		var byBuffs = properties.GetFloat(PropertyName.BLK_BREAK_BM);
+		
+		var byRateBuffs = 0f;
+		byRateBuffs += properties.GetFloat(PropertyName.BLK_BREAK_RATE_BM);
```
exectails, line 466: And this one doesn't exist for monsters either. I checked some older data as well, and it doesn't seem like these properties ever existed for monsters. They didn't even exist for players in older versions. I wonder if we should leave them, in case they're ever added, or if we should rather remove them to avoid confusion.

kenedos: Maybe I'll leave them commented for now so we remember they exist but they're not actually implemented?
Or should I just remove them entirely?
What would be the best solution in your opinion?

exectails: I'm one the fence. They don't exist for monsters currently, so including them in any way could lead to confusion. Either because people don't see them anywhere, or, if they're commented, because one might think they just weren't implemented by us yet. I believe it would be best to remove them until they're actually added for monsters, if ever.

kenedos:
I've removed the nonexisting monster properties mentioned:
- CRTATK_RATE_BM
- BLK_RATE_BM
- BLK_BREAK_RATE_BM

I've also removed the `(int)Math.Max(0, value)` from return of those monster_calc functions, as it seems to make more sense leaving those able to go negative.

Please let me know if there's anything else needed in regards to this PR.

exectails:
Looks good I think. Thanks for the PR.

PR #263, opened 2024-06-29
Adding Swordsman's removed skills

Terotrous:
This PR adds the 4 removed skills back to Swordman.  It also moves the Buff Handlers for Swordsman to their appropriate folders.  Note that the Skill Modifier functionality from Peltasta is also included here (and there's another modifier, for skills that ignore defense).


src/ZoneServer/Skills/Handlers/Swordsman/PommelBeat.cs
```diff
@@ -0,0 +1,108 @@
...
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				SkillModifier modifier = new SkillModifier();
+
+				if (target.IsBuffActive(BuffId.Stun))
+				{
+					modifier.DamageMultiplier = StunDamageMultiplier;
+				}
+				var targetSize = SizeType.M; // non-mobs are always size M
```
exectails, line 82: While I see where you're going with this comment, it's another case of a comment that doesn't contribute much. You're setting M as the default and change it for mobs. That tells us that M is the intended default, and that's really all there is to it. All non-mobs are M as a result. The consequence of wanting to clarify this would mean that you'd have to do this every time you set M as the default somewhere.

src/ZoneServer/Skills/Handlers/Swordsman/PommelBeat.cs
```diff
@@ -0,0 +1,108 @@
...
+				if (target is Mob mob)
```
exectails, line 83: With the goal of making skill handlers as target type agnostic as possible in mind, it might be better to create an extension methods that handles this, returning either the mob data or Mas the default for a combat entity. Or whatever else might be necessary, in case there's a way for players to change their size.

exectails: Another thing to potentially keep in mind, which I just learned about, is that players actually aren't size `M`. They have their own size, called `PC`. Though in certain situations they are equivalent to size `M`, as these two share a "size rank". That's presumably what we want to query for at some point in the future. See `shape.ies` for some more information.

src/ZoneServer/Skills/Handlers/Swordsman/Restrain.cs
```diff
@@ -0,0 +1,51 @@
...
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			var target = caster;
+
+			var duration = TimeSpan.FromSeconds(30 + 3 * skill.Level);
+			// num2 is stun chance on normal attack
```
exectails, line 45: See previous comment about comments on buff parameters. Curiously you used two different names to describe these parameters, and neither matches the actual name, `NumArg2`, adding to the potential confusion.

system/scripts/zone/core/calc_combat.cs
```diff
@@ -89,6 +89,23 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
 
 		var damage = SCR_GetRandomAtk(attacker, target, skill, modifier, skillHitResult);
 
+		// Add extra attack damage based on concentrate
+		// TODO move this to a buff handler
+		if (attacker.Components.Get<BuffComponent>().TryGet(BuffId.Concentrate_Buff, out var concentrateBuff))
```
exectails, line 94: Since you already mentioned moving this to a buff handler, what do you think about putting it in a static method on the buff handlers, similar to what I did with Link? That way the buff-relevant code is all grouped together and this function is easier to read and maintain while we don't have a hook system yet.

system/scripts/zone/core/calc_combat.cs
```diff
@@ -192,6 +218,19 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
 		if (skillHitResult.Result == HitResultType.Crit)
 			damage *= 1.5f;
 
+		// Added effects
+		// Check for Restrain's stun chance, this only activates on normal attacks,
+		// which are skills with ID < 1001
+		if ((int)skill.Id <= 1000 && attacker.TryGetBuff(BuffId.Restrain_Buff, out var restrainBuff))
```
exectails, line 224: For a check like this I would personally rather see a method/property on `Skill`, like `IsNormalAttack`, so we can easily change it down the line if necessary. Because checking a range could break the moment they put some kind of common skill nonsense into it that does damage but isn't considered a normal attack, and then we'd have to change this check everywhere it's done. Additionally it's easier to read and doesn't require a comment.

exectails (changes requested):

src/ZoneServer/Buffs/Handlers/Swordsman/Bear.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+
+namespace Melia.Zone.Buffs.Handlers
+{
+	/// <summary>
+	/// Handle for the Bear Buff, which reduces damage taken
+	/// This class's effect is handled in calc_combat, but this handler
+	/// needs to remain so its duration can be saved on logout.
```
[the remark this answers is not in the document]
Terotrous: Strange, the first time I tested it, it seemed like removing the Buffhandler made the buff not save, but this time it worked fine.  In that case I guess the Buffhandler can be removed, since it has no effect.

src/ZoneServer/Buffs/Handlers/Swordsman/Restrain.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+
+namespace Melia.Zone.Buffs.Handlers
+{
+	/// <summary>
+	/// Handle for the Restrain Buff, which reduces maximum HP
```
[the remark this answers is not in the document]
Terotrous: Okay, I mentioned the other effects of the buff.

src/ZoneServer/Skills/Handlers/Swordsman/Concentrate.cs
```diff
@@ -0,0 +1,52 @@
...
+				return;
+			}
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			var target = caster;
+
+			var duration = TimeSpan.FromSeconds(45);
+			// val2 is the flat damage bonus
```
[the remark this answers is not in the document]
Terotrous: You're right, that is much cleaner.

src/ZoneServer/Skills/Handlers/Swordsman/PommelBeat.cs
Terotrous: Essentially I mainly wanted to make it clear that this wasn't simply an oversight where this skill had only been considered for PVE, but if you prefer I can remove the comment.

src/ZoneServer/Skills/Handlers/Swordsman/PommelBeat.cs
Terotrous: So should we create a SizeType.PC for players to have?  I feel like maybe it's best to leave the implementation of this one up to you.

src/ZoneServer/Skills/Handlers/Swordsman/Restrain.cs
Terotrous: It all makes perfect sense in my head.  I've changed the format to match Concentrate.

src/ZoneServer/Buffs/Handlers/Swordsman/Restrain.cs
```diff
@@ -0,0 +1,36 @@
...
+		public override void OnEnd(Buff buff)
+		{
+			if (buff.Vars.TryGetFloat(VarName, out var penalty))
+				buff.Target.Properties.Modify(PropertyName.MHP_BM, penalty);
+		}
+
+		private float GetMaxHPPenalty(Buff buff)
+		{
+			var skillLevel = buff.NumArg1;
+			return MaxHPDropBase + skillLevel * MaxHPDropPerLevel;
```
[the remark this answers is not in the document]
Terotrous: If you ever want someone to reimplement the buff, you know who to call.

src/ZoneServer/Skills/Handlers/Swordsman/PommelBeat.cs
exectails: Looking at the file mentioned, we probably do want a PC size, but yea, we/I can look into that another time.

system/scripts/zone/core/calc_combat.cs
exectails: I can understand that, I've been going back and forth on it as well. Though you could also do the check here.
```cs
if (attacker.IsBuffActive(BuffId.Concentrate_Buff))
    Concentrate.AddBonus(modifier);
```
Or you could name the method something like `TryAddBonus`, indicating that it's not guaranteed, but that it will happen if applicable.

system/scripts/zone/core/calc_combat.cs
Terotrous: I added one.

system/scripts/zone/core/calc_combat.cs
Terotrous: I guess it's fine for now, once we have a hooking system it won't matter, since only the methods that relevant will be hooked.

PR #264, opened 2024-06-29
Fixes Revealed Map Packets

Terotrous:
This is a small fix that makes saving and loading your map completion to the database work again.

### DIFF
--- a/src/ZoneServer/Network/PacketHandler.cs
+++ b/src/ZoneServer/Network/PacketHandler.cs
@@ -1612,6 +1612,7 @@ namespace Melia.Zone.Network
 		{
 			var mapId = packet.GetInt();
 			var visible = packet.GetBin(128);
+			var percentage = packet.GetFloat();
 
 			// Check if the map exists
 			var mapData = ZoneServer.Instance.Data.MapDb.Find(mapId);
@@ -1632,12 +1633,20 @@ namespace Melia.Zone.Network
 				return;
 			}
 
-			var revealedMap = new RevealedMap(mapId, visible, 0);
+			// Check the percentage for validity
+			if (percentage < 0 || percentage > 100)
+			{
+				Log.Warning("CZ_MAP_SEARCH_INFO: User '{0}' tried to update the visibility for map '{1}' beyond an acceptable percentage.", conn.Account.Name, mapId);
+				return;
+			}
+
+			var revealedMap = new RevealedMap(mapId, visible, percentage);
 			conn.Account.AddRevealedMap(revealedMap);
 		}
 
 		/// <summary>
 		/// Reports to the server a percentage of the map that has been explored.
+		/// This packet is seemingly no longer used, it's been combined with the above
 		/// </summary>
 		/// <param name="conn"></param>
 		/// <param name="packet"></param>
--- a/src/ZoneServer/Network/Send.cs
+++ b/src/ZoneServer/Network/Send.cs
@@ -2031,6 +2031,8 @@ namespace Melia.Zone.Network
 			{
 				packet.PutInt(revealedMap.MapId);
 				packet.PutBin(revealedMap.Explored);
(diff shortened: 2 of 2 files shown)


SalmanTKhan (approved):
LGTM

src/ZoneServer/Network/PacketHandler.cs
```diff
@@ -1632,12 +1633,20 @@ public void CZ_MAP_REVEAL_INFO(IZoneConnection conn, Packet packet)
 				return;
 			}
 
-			var revealedMap = new RevealedMap(mapId, visible, 0);
+			// Check the percentage for validity
+			if (percentage < 0 || percentage > 100)
+			{
+				Log.Warning("CZ_MAP_SEARCH_INFO: User '{0}' tried to update the visibility for map '{1}' beyond an acceptable percentage.", conn.Account.Name, mapId);
```
exectails, line 1639: This should've been updated to say `CZ_MAP_REVEAL_INFO`.

src/ZoneServer/Network/PacketHandler.cs
```diff
@@ -1632,12 +1633,20 @@ public void CZ_MAP_REVEAL_INFO(IZoneConnection conn, Packet packet)
...
+				return;
+			}
+
+			var revealedMap = new RevealedMap(mapId, visible, percentage);
 			conn.Account.AddRevealedMap(revealedMap);
 		}
 
 		/// <summary>
 		/// Reports to the server a percentage of the map that has been explored.
+		/// This packet is seemingly no longer used, it's been combined with the above
```
exectails, line 1649: This should've probably rather been a remark than a part of the summary, as it's unrelated to the function of the method. The summary of `CZ_MAP_REVEAL_INFO` meanwhile could've used an update, since it's not correct anymore. Both of these summaries and any potential remarks should ideally also document this transition to a degree and reference each other, in support of future efforts to add backwards compatibility.

src/ZoneServer/Network/Send.cs
```diff
@@ -2031,6 +2031,8 @@ public static void ZC_MAP_REVEAL_LIST(IZoneConnection conn)
 			{
 				packet.PutInt(revealedMap.MapId);
 				packet.PutBin(revealedMap.Explored);
+				packet.PutFloat(revealedMap.Percentage);
+				packet.PutFloat(0);
```
exectails, line 2035: Have you perchance gained any new insights into what this second float does? Which we apparently documented in the binary templates, but never gave a proper name.

Speaking of binary templates, since the identity of the first float appears clear now, it might be a good idea to also update the templates to reflect this, to improve maintainability and debuggability. Again, not something I would enforce, but a nice to have.

src/ZoneServer/Network/PacketHandler.cs
```diff
@@ -1612,6 +1612,7 @@ public void CZ_MAP_REVEAL_INFO(IZoneConnection conn, Packet packet)
 		{
 			var mapId = packet.GetInt();
 			var visible = packet.GetBin(128);
+			var percentage = packet.GetFloat();
```
exectails, line 1615: It seems like we use the same struct for reading and sending this data in our binary templates, suggesting that there might be a second float after the percentage, just like in the sending part. Have you looked into that by any chance? Is a float missing here or are our templates incorrect?

src/ZoneServer/Network/PacketHandler.cs
Terotrous: My bad, I just copied it from the other method.

src/ZoneServer/Network/Send.cs
Terotrous: I wasn't aware that we had templates, you'll have to let me know where they are some time.

As for the second Float, I have no idea what it does, it might even just be a spacer.

PR #265, opened 2024-06-30
Minor fix to saving buffs

Terotrous:
Previously, the system did not End buffs that were discarded upon map change / logout due to having save: false, causing their effects to become permanent as the buffed properties would not be removed.  This also means that buffs must be saved before properties since ending a buff usually modifies the user's properties.


exectails:
I'm not sure this issue should be fixed by ending the buffs in the middle of the saving process, because I would personally not expect a save call on the database to make any changes to the character's state. It would seem clearer to me to end the buffs before calling into the database. To make the data saving ready before going through with it. What do you think?

Though the issue might get even more complicated than that if we ever wanted to save during runtime, which is by no means unreasonable. Perhaps we should consider adding a modification property collection, similar to the override properties for monsters, where we could set temporary modifiers. Done right, we could potentially even get rid of the whole variable thing, where we manually save the modifiers to later reset them.

Terotrous:
I think you're right where the most sensible way to do this would be prior to a logout / zone change, not during the save method.  Looking at the code, such a method to do this kind of pre-logout cleanup doesn't exist.  Any thoughts on where it should be?  Character.PrepareForLogout perhaps?

exectails:
The saving is currently a bit wonky, taking place in one of two places depending on whether a warp was initiated or not. I'd say let's add a `StopTempBuffs` method to the buffs component, call it from those two locations, and think about it again when we clean the saving up.

PR #267, opened 2024-07-02
Scout fixes

Terotrous:
Some small fixes for Scout.

Dagger Slash's damage buff is now implemented and lasts the correct duration.
Cloaking's damage debuff is now implemented properly.


system/scripts/zone/core/calc_combat.cs
```diff
@@ -73,19 +75,42 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
...
 		}
 
-		var damage = SCR_GetRandomAtk(attacker, target, skill, skillHitResult);
+		var damage = SCR_GetRandomAtk(attacker, target, skill, modifier, skillHitResult);
+
+		// Check for Concentrate
+		if (attacker.Components.Get<BuffComponent>().TryGet(BuffId.Concentrate_Buff, out var concentrateBuff))
+		{
+			var bonusDamage = concentrateBuff.NumArg2;
+			var variableName = "Melia.HitsLeft";
```
exectails, line 91: Would it perhaps make more sense to make this a public const on the buff, so we don't have to copy around a string identifier that is easy to mess up? We could also potentially move the entire logic to a static method on the buff, so all the buff relevant code is in one place, except for where its effects take place. (For now.)

system/scripts/zone/core/calc_combat.cs
```diff
@@ -73,19 +75,42 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
...
+		if (attacker.Components.Get<BuffComponent>().TryGet(BuffId.DaggerSlash_Buff, out var daggerSlashBuff))
+		{
+			modifier.DamageMultiplier += daggerSlashBuff.OverbuffCounter * 0.07f;
+		}
 
 		var skillFactor = skill.Properties.GetFloat(PropertyName.SkillFactor);
 		var skillAtkAdd = skill.Properties.GetFloat(PropertyName.SkillAtkAdd);
 		damage *= skillFactor / 100f;
-		damage += skillAtkAdd;
+		damage += skillAtkAdd + modifier.BonusDamage;
```
exectails, line 112: Note: Kinda wish the skill modifier had been its own PR, as that would've made it easier to review.

src/ZoneServer/Skills/Combat/SkillModifier.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using Melia.Shared.Game.Const;
+
+namespace Melia.Zone.Skills.Combat
+{
+	/// <summary>
+	/// A class for properties that can modify the damage calculation for a skill
+	/// </summary>
+	public class SkillModifier
```
exectails, line 9: The skill modifier class is a good idea, though I gotta admit that I dislike the amount of passing downs we have to do. I wonder if we could perhaps combine this with the skill or the skill hit result in some way, which we're already passing around anyway. Then again, I guess this _is_ most likely a one time change. Hm 🤔

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -328,6 +333,10 @@ protected bool IsHating(ICombatEntity entity)
 			if (!_hateLevels.TryGetValue(entity.Handle, out var hate))
 				return false;
 
+			// Cloaking blocks hate while it's active
```
exectails, line 336: In addition to the current code that disables accumulating more hate, you added a check for excluding cloaked enemies from gaining hate, as well as skipping them among the most hated entities. Effectively this means hate no longer increases and the monsters won't consider you a target for the duration of cloaking, but they will remember and target you again immediately once your cloaking status disappears. Is that the intended behavior or should the hate be cleared?

system/scripts/zone/core/calc_combat.cs
Terotrous: You're right, but this is actually a copy-paste error, Concentrate is part of the Swordman PR and shouldn't be here at all (the buffhandler is not present in this PR).  I've removed it.

src/ZoneServer/Skills/Combat/SkillModifier.cs
Terotrous: What do you mean by passing down?  You mean the number of methods that have to be adjusted to allow for it?

Something I was able to do here is that the skillmodifier is an overload, so if a method doesn't use it, they don't have to include it (a blank skillmodifier will be created that doesn't modify the skill's behaviour in any way).

system/scripts/zone/core/calc_combat.cs
Terotrous: I can see the argument for it, though then there wouldn't be any examples of its usage.

src/ZoneServer/Scripting/AI/AiScript.cs
Terotrous: It is indeed the intended behavior for the hate to remain and for the monster to remain hating you the moment you lose cloak, which could be almost immediately since it's disabled on attack.

src/ZoneServer/Buffs/Handlers/Scout/DoubleAttack_Buff.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using Melia.Shared.Game.Const;
 using Melia.Zone.Buffs.Base;
 
-namespace Melia.Zone.Buffs.Handlers
+namespace Melia.Zone.Buffs.Handlers.Scout
```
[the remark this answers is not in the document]
Terotrous: We can discuss this more, but I feel like it's inevitable that we need some kind of sorting because there's literally hundreds of buffs.  If we ever have to move a buff, it's at least quite simple to do, due to the way the handlers work skills applying the buff don't need to care about its namespace, so moving it from one folder to another is as simple as a drag and drop.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -304,11 +304,16 @@ protected ICombatEntity GetMostHated()
 				var handle = entry.Key;
 				var hate = entry.Value;
 
-				if (hate > highestHate)
-				{
-					highestHate = hate;
-					mostHated = this.Entity.Map.GetCombatEntity(handle);
-				}
+				if (hate < highestHate)
```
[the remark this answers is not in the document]
Terotrous: You're right, when I looked at it I didn't think it would matter, but it is true that it results in checking the buffs more often, so I've changed it to <=.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -304,11 +304,16 @@ protected ICombatEntity GetMostHated()
...
+					continue;
+
+				var entity = this.Entity.Map.GetCombatEntity(handle);
```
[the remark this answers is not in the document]
Terotrous: I've added a null check.  Also, this method is now refactored as below.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -304,11 +304,16 @@ protected ICombatEntity GetMostHated()
...
+
+				if (entity.IsBuffActive(BuffId.Cloaking_Buff))
```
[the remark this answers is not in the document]
Terotrous: I've added a "CanTarget" method, which I think most appropriately sums up what this property does (for example, a cloaked entity, while untargettable, can absolutely still be hit by AOEs).

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
```diff
@@ -44,7 +44,7 @@ public void Handle(Skill skill, ICombatEntity caster, Position originPos, Positi
 			var splashParam = skill.GetSplashParameters(caster, originPos, farPos, length: 55, width: 30, angle: 0);
 			var splashArea = skill.GetSplashArea(SplashType.Square, splashParam);
 
-			var buffDuration = TimeSpan.FromSeconds(2);
+			var buffDuration = TimeSpan.FromSeconds(10);
```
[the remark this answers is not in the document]
Terotrous: I'll add a comment if you wish, though I feel like this could become messy if they make a lot of such changes.

src/ZoneServer/Buffs/Handlers/Scout/DoubleAttack_Buff.cs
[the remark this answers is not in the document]
exectails: There's actually _thousands_ of buffs, but yea, I don't disagree^^

src/ZoneServer/Skills/Handlers/Scout/DaggerSlash.cs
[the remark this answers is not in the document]
exectails: It will most definitely become messy, but that's the name of the game if you want to keep track of what's happening^^

system/scripts/zone/core/calc_combat.cs
```diff
@@ -132,37 +142,43 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
...
 		if (hitCountMultiplier != 1)
 		{
 			damage *= hitCountMultiplier;
 			skillHitResult.HitCount = (int)Math.Round(skillHitResult.HitCount * hitCountMultiplier);
 		}
 
+		if (target.Components.Get<BuffComponent>().TryGet(BuffId.Cloaking_Buff, out var cloakingBuff))
+		{
+			// Cloaking reduces damage by 25%
+			damage = Math.Max(1, damage - damage * 0.75f);
```
exectails, line 179: I'm kind of tired right now, but aren't you reducing the damage by 75% here? Or is that part correct and I'm misunderstanding the comment?

Terotrous: Nope, you're right, that does accidentally reduce it by 75%.  Whoops.

system/scripts/zone/core/calc_combat.cs
exectails: You don't really need an example for this, it's all in the code. The example is just setting a variable and passing the modifier^^

PR #268, opened 2024-07-03
Assassin Tier 1

Terotrous:
Implements Assassin's 3 tier 1 skills.


exectails:
I'm going to hold off on reviewing your other modifier-including PRs until master is merged in, to make it easier to review them.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Behead_Debuff.cs
```diff
@@ -0,0 +1,28 @@
...
+	/// Buff handler for Behead: Bleeding, which deals damage
+	/// every 1.5 seconds
+	/// </summary>
+	[BuffHandler(BuffId.Behead_Debuff)]
+	public class Behead_Debuff : BuffHandler
+	{
+		public override void WhileActive(Buff buff)
+		{
+			if (!buff.Target.IsDead) { 
+				var skill = new Skill(buff.Caster, SkillId.Assassin_Behead_DOT, (int)buff.NumArg1);
```
exectails, line 20: It feels kinda iffy to create a new throwaway Skill object every tick, doesn't it? Especially because we know this would quickly get copied to all buffs and skills with similar effects. Do officials actually send `Assassin_Behead_DOT`?

exectails, line 20: It seems like this skill's max level is 1, so there's presumably no point in setting its level.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Reflection.Metadata;
+using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+
+namespace Melia.Zone.Buffs.Handlers.Scout.Assassin
+{
+	/// <summary>
+	/// Handle for the Bear Buff, which increases the target's defense.
```
exectails, line 8: This summary appears to be copy-pasted from another buff.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
```diff
@@ -0,0 +1,41 @@
...
+	/// </summary>
+	[BuffHandler(BuffId.Hasisas_Buff)]
+	public class Hasisas : BuffHandler
+	{
+		private const string aspdVarName = "Melia.AttackSpeedBonus";
```
exectails, line 13: Our coding conventions dictate that consts, statics, and public fields should be formatted in `PascalCase`.

src/ZoneServer/Skills/Handlers/Scout/Assassin/Behead.cs
```diff
@@ -0,0 +1,110 @@
...
+				var skillHitResult = SCR_SkillHit(caster, target, skill, modifier);
+				target.TakeDamage(skillHitResult.Damage, caster);
+
+				var skillHit = new SkillHitInfo(caster, target, skill, skillHitResult, damageDelay1, skillHitDelay);
+				skillHit.HitEffect = HitEffect.Impact;
+				hits.Add(skillHit);
+				Send.ZC_SKILL_HIT_INFO(caster, hits);
+			}
+
+			await Task.Delay(delayBetweenHits);
```
exectails, line 87: I believe I saw you guys talk about this skill on Discord, and as such I assume you did some research on it, but just for confirmation: Is this how it's supposed to work, with 3 multi hits split up across two skill hits?

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
```diff
@@ -0,0 +1,41 @@
...
+			buff.Vars.SetFloat(aspdVarName, aspdBonus);
+
+			var critBonus = critBonusBase + critBonusPerLevel * buff.NumArg1;
+			buff.Vars.SetFloat(critVarName, critBonus);
+
+			buff.Target.Properties.Modify(PropertyName.ASPD_BM, aspdBonus);
+			buff.Target.Properties.Modify(PropertyName.CRTATK_BM, critBonus);
+		}
+
+		public override void OnEnd(Buff buff)
```
exectails, line 32: Similarly to the skill no longer saying that it consumes a Hasisas Potion, it seems like this also used to
> continuously consume your HP as a side effect

Is this also no longer a thing? Though I did check the buff description, and there the health loss is mentioned.

src/ZoneServer/Skills/Handlers/Scout/Assassin/InstantAcceleration.cs
```diff
@@ -0,0 +1,91 @@
...
+
+			await Task.Delay(hitTime);
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+			var hits = new List<SkillHitInfo>();
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				SkillModifier modifier = new SkillModifier();
+				modifier.HitCount = 4;
```
exectails, line 79: Am I missing something, or should this be a 1-hit skill? Neither the description nor the bytool data appears to mention 4 hits.

exectails (changes requested):

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Behead_Debuff.cs
Terotrous: I added an overload to ZC_Hit_Info that takes a SkillId instead of a Skill object (it only uses the SkillId anyway), which allows this to work without creating a skill object for every tick.

Terotrous: This is no longer necessary, though I disagree in general.  Even if the skill max level is 1 in the skillDb, levelling it up still works normally if the skill level is increased to be more than 1 (even things like factors display and work properly), so we should probably still handle skill levels correctly.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
Terotrous: You got me.  I fixed it.
(I can't believe I did this after fixing so many of them myself in other PRs).

src/ZoneServer/Skills/Handlers/Scout/Assassin/InstantAcceleration.cs
Terotrous: It's in the skill tooltip, I posted a screenshot of it in the chat.  Hopefully this links to it?
https://discord.com/channels/273278107773960192/1128408008469266503/1260631576925175840

src/ZoneServer/Skills/Handlers/Scout/Assassin/Behead.cs
Terotrous: Confirmed by Kenedos that it does work this way.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
Terotrous: Definitely seems like this was removed, I note in the old buff description, it also increased your crit damage by proportion to your missing health, but that whole part of the description is gone.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Behead_Debuff.cs
exectails: While I fully agree with you that we should generally accommodate levels past the maximums, I don't believe this applies to `Assassin_Behead_DOT`, because it's a dummy skill without any additional effects. It doesn't do anything on its own and has no scaling factors whatsoever. As such, I would argue setting a level may cause confusion, making someone think there actually is more to it, when in reality there isn't and likely never will be.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
exectails: Just checked the latest description of the buff once more to be absolute certain, and it does still state that it uses HP.
> Increases basic attack speed and critical damage. Reduces HP over time

Unless that effect is not present on officials, we probably want to add it to the buff, right? That will also make it less weird that you no longer need an item to use it.

src/ZoneServer/Skills/Handlers/Scout/Assassin/InstantAcceleration.cs
exectails: Well, apparently I'm blind. Nevermind me^^

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Hasisas.cs
kenedos: Checked on official and seems it does not consume hp overtime or upon cast at all.
Buff tooltip still says it does, however.

src/ZoneServer/Buffs/Handlers/Scout/Assassin/Behead_Debuff.cs
Terotrous: True, though we're not doing it anymore since we're no longer creating the skill object with the new overload.

exectails: True, but it's something we can keep in mind going forward^^

PR #272, opened 2024-07-08
Update to templates based on Map Changes

Terotrous:
As mentioned in the previous PR for this feature, there were some documentation changes requested. Note that MapRevealData does not include a second float, attempting to get another float gives the end of buffer error.  It seems the extra float sent while sending the data to the client is just a spacer.

### DIFF
--- a/doc/bt/ZC_MAP_REVEAL_LIST.bt
+++ b/doc/bt/ZC_MAP_REVEAL_LIST.bt
@@ -20,4 +20,5 @@ int count;
 if (!checkCompression())
 {
     mapRevealData dataSets[count];
+    float spacer;
 }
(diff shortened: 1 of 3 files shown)


doc/bt/ZC_MAP_REVEAL_LIST.bt
```diff
@@ -20,4 +20,5 @@ int count;
 if (!checkCompression())
 {
     mapRevealData dataSets[count];
+    float spacer;
```
exectails, line 23: I believe this can't be correct, because based on my latest logs, the second float is part of every entry in `ZC_MAP_REVEAL_LIST`. There's not only one at the end of the list. Though is seems like the second float is always the same as the first.

Terotrous: What would you like the variable to be called?

exectails: The name is not the problem, that it doesn't exist is^^" I'll take care of this.

exectails:
I implemented the changes and updated the template. Thanks for the reminder^^

PR #279, opened 2024-07-19
Pathfinding

kenedos:
- Implements pathfinding using A* algorithm.
- Algorithm uses a dynamic grid for path nodes, allowing the grid to expand and retract depending on distance to target.
- Algorithm implements collision checking for entities based on their sizes.
- All entities are assumed to be circular in collision detection (this seems to be correct).
- Uses ground "height check" to check for collision.

Current Limitations:
- The algoritm _can_ find paths for destinations really far away (i.e. across the map) but it is currently limited to not do so due to it's collision check.
- Using height checks for collision is not ideal. There are situations where terrain has ground data but is only traversable in one direction (i.e. going up cliffs).
- Does not handle collision with other entities, such as monsters with OBBs (object bounding boxes).
- There are certain cells in which pathfinding should not be done due to terrain complexity. The default behaviour in these cases is to make the monster "return to home". This is not implemented in this PR.


kenedos:
@exectails regarding the dictionary I created, `_entitySizeRadius` , I'm sure you won't like it being there, but I wasn't sure where to place it exactly.

This size is not the same as the monster's bounding box (the "obb") and it is circular in shape instead of squared. It is based on data from shape.ies and the monster size itself.

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
```diff
@@ -125,28 +127,79 @@ private TimeSpan MoveToConditional(Position destination, bool executeMove)
...
+		/// Sets the target of entity to the next position
+		/// in its path list.
+		/// </summary>
+		private void MoveToNextPosition()
+		{
+			if (_path.Count == 0)
+				return;
+
+			// Remove completed path node
+			_path.RemoveAt(0);
```
exectails, line 170: Is there a particular reason you aren't using a Queue for the path?

src/ZoneServer/World/Maps/Pathfinder.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Linq;
+using Melia.Shared.World;
+using Melia.Shared.Game.Const;
+using Yggdrasil.Util;
+
+namespace Melia.Zone.World.Maps
+{
+	public class Pathfinder
+	{
+		private readonly int _maxNodeExpand = 500;
```
exectails, line 12: There's nothing really wrong about these "consts", including the dictionary, though it would be more appropriate to make them `readonly static` and `PascalCase`, as they never change.

src/ZoneServer/World/Maps/Pathfinder.cs
```diff
@@ -0,0 +1,266 @@
...
+			}
+
+			// No path found
+			return path;
+		}
+
+		/// <summary>
+		/// Compares two float and position tuples.
+		/// </summary>
+		private class PositionComparer : IComparer<(float F, Position Pos)>
```
exectails, line 147: Is this used anywhere or is it obsolete?

src/ZoneServer/World/Maps/Pathfinder.cs
```diff
@@ -0,0 +1,266 @@
...
+						neighbors.Add(neighbor);
+					}
+				}
+			}
+
+			// No adjacent neighbors found
+			if (neighbors.Count == 0)
+			{
+				// Try again recursively with lower scale
+				return this.GetNeighbors(pos, entitySize, gridScale/2);
```
exectails, line 260: It would be preferential if you used auto formatting on your code, as code that is not formatted based on our conventions will create unnecessary whitespace changes the next time someone auto formats a file. Like here with the missing spaces between `gridScale/2`. It's technically a minor thing of course, but it can make reading PRs or double checking commits a whole lot more annoying when you have to find the actual changes first.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -137,7 +142,10 @@ private void Load()
 
 			// A few maps don't seem to have any ground data.
 			if (ZoneServer.Instance.Data.GroundDb.TryFind(this.ClassName, out var groundData))
+			{
 				this.Ground.Load(groundData);
+				this.Pathfinder.Load(this.Ground);
```
exectails, line 147: Would it perhaps make more sense to pass the whole map to the pathfinder? Decoupling the two would be nice, but the pathfinder will presumably need to know about monsters, their bounding boxes, and things like platforms down the line, in addition to the ground. Probably easiest to just give it everything it will need. Though we could also add an interface with methods for providing the ground and potential obstacles for maps or other consumers.

src/ZoneServer/World/Maps/Pathfinder.cs
```diff
@@ -0,0 +1,266 @@
...
+		/// Finds a path from start position to the goal position using
+		/// A* algorithm. Returns List of valid positions to goal.
+		/// Last element is always the closest position to goal.
+		/// Returns empty list if no path can be found.
+		/// </summary>
+		/// <param name="start"></param>
+		/// <param name="goal"></param>
+		/// <param name="entitySize"></param>
+		/// <returns></returns>
+		public List<Position> FindPath(Position start, Position goal, SizeType entitySize = SizeType.M)
```
exectails, line 44: The entity size parameter should be sanity checked, as you're using it as the index for accessing the dictionary. You're expecting it to be `>= None <= XXL`, but we know those aren't all of them, and passing in an unexpected value would cause exceptions.

exectails (changes requested):
Thanks for the PR. I'm surprised by the simplicity of this implementation, kudos for that. Though I have two requests to make this a bit more flexible.

1) I believe we should create an interface for a general purpose path finder. This will allow us to quickly swap out the path finder or use alternative ones, should we need or want to switch to something else down the line.
2) Separate the pathfinding movement from the normal `MoveTo` behavior. This is so we still have access to both if necessary and to not break existing code. It will also make it easier to merge the combat entity movement component with the more general purpose actor movement component I introduced for pads. The goal will be to reduce redundancies without adding too many conditions inside the movement code. The `MoveTo` and `MovePath` (?) could technically merge into the same movement code afterward, just that one may ignore obstacles and only ever sets up one destination.

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
kenedos: Yes - I'm stupid :)

Initially I was using a list and iterating over it, figured out that wasn't a good idea and didn't realize I basically had a Queue.

kenedos:
- Movement Component now has straightline and/or pathfinding support
- Can calculate path and request to move separately
- Added IPathfinder interface
- Pathfinder Load() receives Map
- Changed _path to Queue
- Sanity Check entitySize
- Improved heuristic for height check
- Added angle subdivision on GetNeighbors()
- Formatting fixes
- Removed unecessary code

exectails (commented):
The monster movement using your path finding is somewhat peculiar. That they sway left and right now and then should probably feel more natural, but somehow it seems weird^^" Especially if they have a clear path to their target and should want to get there as quickly as possible on a straight line. Idle movement, too, seems a bit erratic. Instead of walking straight to a random position, they often move for a bit and then make 90° turns. I think we'll probably want to refine this a little over time, but it's a decent starting point. Thanks!

kenedos:
The "random swaying" was intentionally added to the heuristic to simulate monster behaviour in official. 

If you have 5 monsters after you and they always take the shortest path to you (that is, no random to heuristic), they always "clump up" together becoming one on top of eachother.

Official does seem to have some randomness to their heuristic, but on top of that, they don't always move exactly to their target. They sometimes stop on positions adjacent to you and may even walk past you before starting to attack. That change, however, would be mostly on the mob AI we have right now so I decided to leave it out of scope for this PR.

If you want to completely take away this "random swaying" of them, you can remove the randomFactor variable from this line:
`fScore[neighbor] = gScore[neighbor] + this.Heuristic(neighbor, goal) + randomFactor;`

As for the random movement mobs do while idle, they shouldn't really be using pathfinding for that in first place, it's a rather unecessary calculation. Ideally we should change the monster AI itself to not use pathfinding while random idle moving, and again I felt that ran out of scope for this PR.

exectails:
Thanks for the explanation. I agree that `MoveRandom` shouldn't really use pathfinding and changed how it works when I went over the PR. Personally I would've like to see this "fixed", since the PR affected it in an arguably negative way. The chase behavior I'm not sure about though. A bit of randomness should probably be a good thing, like I said, but to me personally it currently feels a bit off. Granted, I haven't done a detailed analysis of officials yet, but I don't recall ever seeing this behavior. At least not when a single monster was chasing me. Could this instead be a result of the monsters going out of each other's way to a degree, to prevent them converging on the same exact path?

exectails:
I took a brief look at the monster chase behavior on officials, and given an obstacle free path, single monsters appear to move perfectly smooth and head straight for their target, as I would expect. Though I did notice one detail you mentioned, where they sometimes seem to almost pass you before attacking. It feels like they might be targeting a position on a circle around you, but not necessarily the one that is the absolute closest between your two positions.

kenedos:
We can remove the randomFactor altogether if you prefer. This is what the pathing look like when it's not added to the heuristic:

https://www.youtube.com/watch?v=6jH9kzDYjto&feature=youtu.be

So no random 'swaying' at all.

As for the AI, I will open another PR to implement that and fix the broken idle behaviour you mentioned. Apologies for breaking it ^^;

kenedos:
I don't know how to explain it, but in official they don't seem to clump up together as hard as without the randomFactor added. I agree it may be too high at the moment, maybe we should divide it by 10 or something like that, to add _just_ a little bit of randomness. Here's the reference behaviour in official:

https://www.youtube.com/watch?v=ZqBbV5pvNXU

kenedos:
In fact, after watching it a couple times, I think I know what's going on here: 

They're not calculating pathfinding per Update() in official. When the mob reaches the destination, it calculates pathfinding again if it doesn't find an attack target.

The mobs target a position adjacent to you and start moving there, thus, they don't seem to "clump up" as much. If mobs are super slow (like the Zignuts), they still clump up because they move too slow anyways.

I think we can completely remove the randomFactor, then, and work on improving the AI to better simulate that official behaviour.

kenedos:
Although what I said wouldn't explain 100% the behaviour. They still do turn around to chase me before they reach their destination, so there is some kind of trigger that makes them calculate pathfinding again pre-emptively as well.

exectails:
> They're not calculating pathfinding per Update() in official. When the mob reaches the destination, it calculates pathfinding again if it doesn't find an attack target.

I feel like the official pathfinding might work slightly different from what we have currently implemented. I can't say for sure, of course, but they're using a standard pathfinding library that uses navmeshes, and even though you once mentioned somewhere that the official pathfinding is not deterministic, I'm rather certain that it is, with some exceptions sprinkled in to avoid crowds converging on one location. And I believe they are recalculating paths on Update, though potentially with an exception if the monster's target didn't move, for performance reasons.

> Notice how the mobs in R1 always turn around a bit slower than in Melia

Not entirely sure what you're referring to here, but it is very possible that officials have a higher "FPS" count, which would affect how quickly the AI can react and reevaluate. We're currently updating the world only every 250ms, though I've been considering lowering that to 100ms.

PR #282, opened 2024-07-20
Hoplite Skills

Terotrous:
This adds all 8 skills for hoplite, including the removed skill Long Stride.


src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
```diff
@@ -0,0 +1,198 @@
...
+		private void ExplosionAttack(Skill skill, ICombatEntity caster, ISplashArea splashArea)
+		{
+			var damageDelay = TimeSpan.FromMilliseconds(50);
+			var skillHitDelay = TimeSpan.Zero;
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+			var hits = new List<SkillHitInfo>();
+
+			// The explosion has its own SR which is separate from the skill
+			var sr = 8;
```
exectails, line 171: I think you should stop calling this S(D)R, because that's not what it is. Also, if this is an inherent feature we need for pads, to retrieve the amount of actors limited by concurrent use, we should add that instead.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
```diff
@@ -0,0 +1,198 @@
...
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="splashArea"></param>
+		private void ExplosionAttack(Skill skill, ICombatEntity caster, ISplashArea splashArea)
+		{
+			var damageDelay = TimeSpan.FromMilliseconds(50);
+			var skillHitDelay = TimeSpan.Zero;
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
```
exectails, line 167: Instead of keep writing custom solutions, I believe it would be preferable to utilize the pad more. You define it with the trigger area, and it tracks the actors inside already. You don't need a second call into `Map`. And if you feel like the pad should be able to return X random attackable actors from inside, why not add that method and see how it feels?

src/ZoneServer/Skills/Combat/SkillModifier.cs
```diff
@@ -63,6 +63,14 @@ public class SkillModifier
 		/// </remarks>
 		public int HitCount { get; set; } = 1;
 
+		/// <summary>
+		/// Gets or sets forced critical status.
+		/// </summary>
+		/// <remarks>
+		/// If this is true, the attack always deals a critical hit.
+		/// </remarks>
+		public bool ForcedCritical { get; set; }
```
exectails, line 72: Why add forced critical but not forced block, in preparation for when we need to move the code and will need it?

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,132 @@
...
+		/// <param name="farPos"></param>
+		public void Handle(Skill skill, ICombatEntity caster, Position originPos, Position farPos, ICombatEntity target)
+		{
+			if (!skill.Vars.TryGet<Position>("Melia.ToolGroundPos", out var targetPos))
+			{
+				caster.ServerMessage(Localization.Get("No target location specified."));
+				return;
+			}
+
+			if (!caster.Map.Ground.IsValidPosition(targetPos))
```
exectails, line 58: Don't forget to add a check for collisions, because you were not able to jump over obstacles originally, an option to turn that off, as well as a check for the distance, to makes sure players don't jump wherever they want by modifying the client.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,132 @@
...
+			{
+				caster.ServerMessage(Localization.Get("Not enough SP."));
+				return;
+			}
+			
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			Send.ZC_SKILL_READY(caster, skill, originPos, targetPos);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, targetPos, null);
```
exectails, line 74: Random thought: Perhaps `ZC_SKILL_MELEE_GROUND` would be used for the sending of hits if there's no delay between them?

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,132 @@
...
+
+				var skillHit = new SkillHitInfo(caster, target, skill, skillHitResult, damageDelay, skillHitDelay);
+
+				skillHit.KnockBackInfo = new KnockBackInfo(caster.Position, target.Position, skill);
+				skillHit.HitInfo.Type = skill.Data.KnockDownHitType;
+				target.Position = skillHit.KnockBackInfo.ToPosition;
+
+				hits.Add(skillHit);
+
+				if (caster.Components.TryGet<AbilityComponent>(out var abilities) && abilities.IsActive(AbilityId.Hoplite30))
```
exectails, line 119: `IsAbilityActive`.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,132 @@
...
+				{
+					target.StartBuff(BuffId.Stun, skill.Level, 0, TimeSpan.FromSeconds(2), caster);
+				}
+			}
+
+			Send.ZC_SKILL_HIT_INFO(caster, hits);
+
+			// This skill requires an additional cast cancel at the end
```
exectails, line 127: Does it? What happens otherwise? I don't remember any issues when I tested the skill without this yesterday. Either way, it would be better if you told others _why_ something is the way it is.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
```diff
@@ -0,0 +1,112 @@
...
+			await Task.Delay(hitDelay);
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+			var hits = new List<SkillHitInfo>();
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				var hitCount = 4;
+
+				if (!Feature.IsEnabled("PierceNoSizeEffect")) {
```
exectails, line 75: Your code is often times not formatted correctly, like the curly brace on the same line as the `if` here. Have you considered using an auto formatter? That may make life easier.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
```diff
@@ -0,0 +1,112 @@
...
+						hitCount = 3;
+					if (targetSize == SizeType.XL || targetSize == SizeType.XXL)
+						hitCount = 4;
+				}
+
+				var modifier = SkillModifier.MultiHit(hitCount);
+
+				var skillHitResult = SCR_SkillHit(caster, target, skill, modifier);
+
+				if (skillHitResult.Result == HitResultType.Crit && caster.Components.TryGet<SkillComponent>(out var skills)
```
exectails, line 89: Code lines should not be broken up, as per our coding conventions.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
exectails, line 75: Structures like this tend to make the code harder to read and such simple assignment operators are prime candidates for being put into the own little methods, where can also easily be modified without having to sift through unrelated code.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/SpearLunge.cs
```diff
@@ -0,0 +1,124 @@
...
+			var hits = new List<SkillHitInfo>();
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				var modifier = SkillModifier.Default;
+				modifier.DefensePenetrationRate = 0.15f;
+
+				var skillHitResult = SCR_SkillHit(caster, target, skill, modifier);
+
+				if (skillHitResult.Result == HitResultType.Crit && caster.Components.TryGet<SkillComponent>(out var skills)
```
exectails, line 77: If you notice yourself doing the same cumbersome operation over and over, like getting a skill via a component, you might want to consider adding helper methods to make said operation easier. Especially if there's already precedents for such helpers, like there are for buffs and abilities.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/SpearLunge.cs
```diff
@@ -0,0 +1,124 @@
...
+			await Task.Delay(delayBetweenHits);
+			hits.Clear();
+			targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				var modifier = SkillModifier.Default;
+				modifier.DefensePenetrationRate = 0.15f;
+
+				var skillHitResult2 = SCR_SkillHit(caster, target, skill, modifier);
```
exectails, line 102: There isn't really any reason to append a `2` to this variable name, as the two results are both in their own scope. Generally you would do this if they're in the same scope, which means this naming convention conveys something that isn't the case.

src/ZoneServer/Pads/Handlers/Swordsman/Hoplite/ThrowingSpear_Hoplite33.cs
```diff
@@ -0,0 +1,64 @@
...
+	/// </summary>
+	[PadHandler("ThrouwingSpear_Hoplite33_Pad")]
+	public class ThrouwingSpear_Hoplite33_Pad : ICreatePadHandler, IUpdatePadHandler, IDestroyPadHandler
+	{
+		/// <summary>
+		/// Called when the pad is created.
+		/// </summary>
+		/// <param name="sender"></param>
+		/// <param name="args"></param>
+		public void Created(object sender, TriggerArgs args)
```
exectails, line 27: I'd appreciate it if you could update your pad handling for the latest version of master, where handlers not use `PadTriggerArgs` and subscriptions should be done via the `Subscribe` method in most cases. Once this is done, you also don't need to check the types anymore, as you currently are below.

system/db/features.txt
```diff
@@ -105,6 +105,9 @@
 		
 		// Changes Hasisas to not base crit bonus on HP lost
 		{ name: "HasisasNoHpCritBonus", enabled: true },
+
+		// Changes Pierce to not base its hit count on target size
```
exectails, line 109: Just for clarification, this was an official change that occurred some time after Re:Build was added?

exectails (changes requested):

src/ZoneServer/Pads/Handlers/Swordsman/Hoplite/ThrowingSpear_Hoplite33.cs
```diff
@@ -0,0 +1,64 @@
...
+using Melia.Zone.World.Actors.Pads;
+using Yggdrasil.Logging;
+
+namespace Melia.Zone.Pads.Handlers.Swordsman.Hoplite
+{
+	/// <summary>
+	/// Handler for the ThrouwingSpear_Hoplite33_Pad,
+	/// which displays an explosion effect in a given position
+	/// </summary>
+	[PadHandler("ThrouwingSpear_Hoplite33_Pad")]
```
exectails, line 19: You should consider using the static `PadName` class, as it may prevent accidental typos.

src/ZoneServer/Network/Send.Normal.cs
```diff
@@ -245,6 +246,50 @@ public static void SkillProjectile(ICombatEntity entity, string packetString1, T
...
+			/// <param name="animationId"></param>
+			/// <param name="scale"></param>
+			/// <param name="tossScale"></param>
+			/// <param name="hangScale"></param>
+			/// <param name="speed"></param>
+			/// <param name="f5"></param>
+			/// <param name="f6"></param>
+			/// <param name="f7"></param>
+			/// <param name="itemScale"></param>
+			/// <param name="itemAppearOnFloorDuration"></param>
```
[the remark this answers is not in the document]
Terotrous: I changed it to itemStayOnGroundSeconds, which I think is more descriptive and also slightly shorter.  I actually don't really like the idea of using a timespan here, because we just convert it right back to a float, it seems like unnecessary overhead.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
```diff
@@ -0,0 +1,198 @@
...
+
+			skill.IncreaseOverheat();
+			caster.SetAttackState(true);
+
+			Send.ZC_SKILL_READY(caster, skill, originPos, farPos);
+			Send.ZC_SKILL_MELEE_GROUND(caster, skill, farPos, null);
+
+			if (caster is Character character)
+			{ 
+				Send.ZC_NORMAL.PadSetModel(caster, "warrior_f_", character.Inventory.GetItem(EquipSlot.RightHand).Id);
```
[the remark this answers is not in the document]
Terotrous: It was Akinue code.  You're right, this is not required.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
```diff
@@ -0,0 +1,198 @@
...
+				Send.ZC_NORMAL.Skill_ItemToss(caster, "warrior_f_", "RH", farPos, "F_smoke177", 3, 0.2f, 0, 600, 1, 435, 485, 0, 3);
```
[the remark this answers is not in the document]
Terotrous: This was also Akinue code.  I can see that the values up until the 435 match skill_bytool, so they are correct, but the 435 and 485 do not.  Upon fiddling with these values, these actually control the item's rotation, so I've corrected them so the spear is thrown in the correct direction.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
```diff
@@ -0,0 +1,198 @@
...
+			}
+
+			this.Attack(skill, caster, new Circle(farPos, 50));
+
+			// This skill has a follow-up that uses a pad if Hoplite33 is active
+			if (caster.Components.TryGet<AbilityComponent>(out var abilities) && abilities.IsActive(AbilityId.Hoplite33))
```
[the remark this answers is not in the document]
Terotrous: That's handy.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
Terotrous: I'm confused, to me this does seem to be SDR.  Isn't SDR the maximum target count of a skill?

src/ZoneServer/Skills/Combat/SkillModifier.cs
Terotrous: I actually don't think a skill can add forced block, since block is something the defender does, not the attacker.  Forced Block comes from buffs.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
Terotrous: For implementations with potentially wide-ranging impacts like this it's always my preference to let you add them.  For example, I don't really know if that method should have the code to override the value if ignoreSDR is set, or if the pad should use its maxActors value to limit how many values it returns.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
Terotrous: Do we know you couldn't jump over obstacles originally?  I note that the "splash" part of the animation will display in the selected location even if it crosses out of bounds, so if you cut the jump short by checking for invalid terrain it messes up the animation.
I did add the max distance check, though.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
Terotrous: I don't quite understand the comment, there definitely is a delay here.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
Terotrous: Done.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
Terotrous: There's a significant freeze period after the skill animation ends where you can't move.  It's possible this could be due to values in skills.txt not being specified properly, though I am hesitant to modify them since they're pulled from an IES file.
I've amended the comment to make this more clear.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
Terotrous: I fixed that one, but I don't know if I'd say "often".

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
Terotrous: Okay, I've changed this to be one line every place it occurs (which is a lot).

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
Terotrous: Done.  There was actually also a small mistake in this code in that the hitcount wasn't being set to 1 if none of those size modifiers applied.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/SpearLunge.cs
Terotrous: In this case adding the new method is simple enough (and there's already an existing example) that I did add it.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/SpearLunge.cs
Terotrous: True.

src/ZoneServer/Pads/Handlers/Swordsman/Hoplite/ThrowingSpear_Hoplite33.cs
Terotrous: Okay, hopefully this is right.

system/db/features.txt
Terotrous: I believe so.  I can't find the exact patch where they changed it, but as of 2019 the patch notes still list it doing 2-4 hits.

src/ZoneServer/Pads/Handlers/Swordsman/Hoplite/ThrowingSpear_Hoplite33.cs
Terotrous: I expected it would be that, but I actually based it off your implementation for the Shield Lob pad, which is specified using a string.

src/ZoneServer/Network/Send.Normal.cs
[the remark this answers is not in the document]
exectails: That's better. Though I do want to say, while it undoubtedly takes cycles to turn the TimeSpan into a float, it shouldn't be understated how much it improves usability. You don't use it because it's faster, but because there are no questions about units and no magic numbers, it's clear and to the point. And I would argue that's generally much more important than saving fractions of a nanosecond on the cast to and from a TimeSpan^^

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
exectails: Not exactly, even though you may view it like that for simplicity's sake. It is our current understanding that a character's/skill's `SR` affects the number of targets it can hit, but the actual limiting factor is the targets' `SDR`, which differs between players and monsters and is affected by things like equipment, buffs, sizes, etc. There is a clear distinction between this and a simple target counter. If you call this `SR`, I'm starting to wonder why you aren't using the target's `SDR`, or why the attacker's actual `SR` isn't taken into consideration. The answer to that, as well as the question why the number 8 you have presumably found in the pad_skill_list doesn't match the skill's `SR`, is that that's not what his is. It's the "concurrent use" of the pad. Something separate from the skill's `SR`, as you rightfully concluded.

src/ZoneServer/Skills/Combat/SkillModifier.cs
exectails: I think you mentioned somewhere that you understood what I was going for with this suggestion after seeing the combat event interfaces, right?^^

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
exectails: Actually, nevermind. I thought I had found proof that you couldn't jump over obstacles, but after doing some more research it actually seems like you _were_ able to. Honestly did not expect that^^

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
exectails: No, but there was an open question why it is used in some skills, and I was wondering whether these other skills might not have a delay, and then send their skill hits immediately with this packet instead of separately.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/Pierce.cs
exectails: As the one who usually just fixes them, I can say they occur often enough for me to notice^^

src/ZoneServer/Pads/Handlers/Swordsman/Hoplite/ThrowingSpear_Hoplite33.cs
exectails: Shame on me, forgot to update Shield Lob^^

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/ThrowingSpear.cs
Terotrous: I understand what you're getting at, but I still don't know how we should handle it in this case.

src/ZoneServer/Skills/Combat/SkillModifier.cs
Terotrous: Yes, I actually added it to the modifier class already.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,144 @@
...
+			await Task.Delay(hitDelay);
+
+			var targets = caster.Map.GetAttackableEntitiesIn(caster, splashArea);
+			var hits = new List<SkillHitInfo>();
+
+			foreach (var target in targets.LimitBySDR(caster, skill))
+			{
+				var skillHitResult = SCR_SkillHit(caster, target, skill);
+
+				if (skillHitResult.Result == HitResultType.Crit && caster.TryGetSkill(SkillId.Hoplite_SharpSpear, out var sharpSpear))
```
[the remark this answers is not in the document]
Terotrous: I considered that, but it only applies to specific hoplite skills, so you'd need to check the skill ID in the code.  It could definitely be done using a case statement or some such.

src/ZoneServer/Skills/Handlers/Swordsman/Hoplite/LongStride.cs
```diff
@@ -0,0 +1,144 @@
...
+				{
+					skillHitResult.Damage += skillHitResult.Damage *= (0.1f + sharpSpear.Level * 0.02f);
```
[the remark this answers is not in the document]
Terotrous: It's definitely not, haha.

src/ZoneServer/Buffs/Handlers/Skill_MomentaryBlock_Buff.cs
```diff
@@ -0,0 +1,65 @@
...
+		/// <param name="attacker"></param>
+		/// <param name="target"></param>
+		/// <param name="skill"></param>
+		/// <param name="modifier"></param>
+		/// <param name="skillHitResult"></param>
+		public void OnDefenseBeforeCalc(Buff buff, ICombatEntity attacker, ICombatEntity target, Skill skill, SkillModifier modifier, SkillHitResult skillHitResult)
+		{
+			Log.Warning("Checking forced block");
+			// Magic is unblockable, so it doesn't register a block in this case
+			if (skill.Data.ClassType != SkillClassType.Magic)
```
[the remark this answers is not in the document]
Terotrous: Oops, I see that I left a debug statement in there, I was sure I took those out.

Anyway, the main reason for the check is that if you block a hit with momentary block, it saves a variable into the buff that some skills later check to deal bonus damage.  It needs to know that taking a hit from Magic doesn't constitute a valid block, since you can't block magic.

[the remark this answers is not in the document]
exectails: Oh, so the hit isn't considered valid for activating the block? Right, that makes sense.

PR #283, opened 2024-07-21
Rare Monsters Update

kenedos:
Rewrote part of the code for blinking monsters because I needed it to scale linearly according to the values listed in conf files.

I tried to remain as faithful as possible to official drops, basing my observations on these two examples:

(Silver) Green Griba  - 25k Silver, 4 Dilgele
https://www.youtube.com/watch?v=SrjSUuTFRRg

(Gold) Rambear - 136k Silver, 6 Rambear Bones
https://www.youtube.com/watch?v=TTNs3swHRq4

### DIFF
--- a/src/ZoneServer/World/Actors/Monsters/Mob.cs
+++ b/src/ZoneServer/World/Actors/Monsters/Mob.cs
@@ -443,8 +443,6 @@ namespace Melia.Zone.World.Actors.Monsters
 			var autolootChance = killer?.Variables.Temp.Get("Autoloot", 0) ?? 0;
 			var worldConf = ZoneServer.Instance.Conf.World;
 
-			var superDropLevel = this.GetSuperDropLevel();
-
 			foreach (var dropItemData in this.Data.Drops)
 			{
 				if (!ZoneServer.Instance.Data.ItemDb.TryFind(dropItemData.ItemId, out var itemData))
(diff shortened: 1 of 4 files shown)


src/ZoneServer/World/Actors/Monsters/Mob.cs
```diff
@@ -462,23 +460,38 @@ private List<DropStack> GenerateDropStacks(Character killer)
...
-					var dropSuccess = rnd.NextDouble() < superAdjustedDropChance / 100f;
-					if (!dropSuccess)
-						continue;
+					switch (superMobType)
+					{
+						case SuperMobType.Silver:
+							superMobRerolls = worldConf.SilverJackpotRolls;
+							superMobGuaranteedItemDrop = originalDropChance > worldConf.SilverJackpotGuaranteedItemThreshold;
+							superMobIncreaseMoney = worldConf.SilverJackpotRolls;
+							superMobMoneyStacks = RandomProvider.Next(40, 50);
```
exectails, line 477: I would generally recommend using `RandomProvider.Get().Next` over `RandomProvider.Next`, because the latter needs to lock a `Random` instance to produce a number. The former creates one instance per thread that can then be used unobstructed, and any additional calls from the same thread will return that already existing instance.

Though in this specific case you already have a `Random` instance in the variable `rnd`, which you could use directly.

PR #291, opened 2024-08-03
Added buff handlers for missing skills

kenedos:


exectails:
Whoops. Good catch, thanks for the fix!

PR #300, opened 2024-08-16
Fix combat modifiers hooks

kenedos:
This PR fixes missing hooks for passive skills and abilities on AfterCalc, BeforeBonuses and AfterBonuses.

For some reason the current implementation only had it on BeforeCalc.


exectails:
Huh, terrible oversight. Thanks for the PR!

PR #302, opened 2024-08-16
Sadhu skills

felipecastagnarodecarvalho:
All Sadhu Skills:
 - Enira
 - Moksha
 - Patati
 - Possession
 - Prakriti
 - SpiritExpert
 - Tanoti
 
 Also there was an implementation of the art Spirit Expert: Wandering Soul which creates a dummy that acts by himself.


src/ZoneServer/World/Actors/Characters/Character.cs
```diff
@@ -594,6 +609,9 @@ public void Warp(int mapId, Position pos)
 			if (!ZoneServer.Instance.Data.MapDb.TryFind(mapId, out var map))
 				throw new ArgumentException("Map '" + mapId + "' not found in data.");
 
+			if (this.IsOutOfBody())
```
exectails, line 612: I feel like this is a weird place to check for out of body to prevent warps, because what if we do want to warp? This special behavior also isn't documented anywhere, so anyone who might try to warp someone will have to dig deep into the warp code to figure out why it's not working.

Additionally, creating explicit checks for very specific conditions within general purpose functions is not ideal, because it's not extensible. An alternative idea that I'm just going to throw out there might be to add a warp lock, which you can set for these buffs you're checking in `IsOutOfBody`, which will generalize the whole approach, make it somewhat self-documenting, save you from having to do that manual buff check, and have it be reusable.

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
```diff
@@ -404,12 +404,21 @@ private void CheckWarp()
...
 				this.MoveSpeedType = type;
 				this.Entity.Properties.Invalidate(PropertyName.MSPD);
 
 				Send.ZC_MSPD(this.Entity);
+			} else if (this.Entity is Character character && character.IsDummy)
+			{
+				this.MoveSpeedType = type;
+				character.Properties.Invalidate(PropertyName.MSPD);
+
+				Send.ZC_MSPD(character.Owner, character);
```
exectails, line 421: Can you  tell me what the exact purpose of this `else if` is? It seems like you could just pass the character into `ZC_MSPD`, which would allow you to leave it as one if.

system/scripts/zone/ais/sadhu_dummy.cs
```diff
@@ -0,0 +1,188 @@
...
+	protected IEnumerable Idle()
+	{
+		var movement = this.Entity.Components.Get<MovementComponent>();
+
+		movement.SetMoveSpeedType(MoveSpeedType.Run);
+		
+		if (this.Entity is Character entityCharacter && entityCharacter.IsDummy)
+		{
+			SetRunning(true);
+			entityCharacter.Properties.Modify(PropertyName.MSPD_BM, 55);
```
exectails, line 45: This hack is not ideal. `SetRunning` technically does something similar to what you wanted to do, but it's limited to mobs because characters don't differentiate between walking and running. Seeing that and copying the code into the AI script just causes unnecessary redundancies though.

It's also not clear what `SetRunning` and this are supposed to do together. One might think the former is still there in case this AI is used for a monster, but then it wouldn't belong in this if. And even if it did, you're also calling `SetMoveSpeedType`, which is the very method `SetRunning` is calling as well.

What I believe you probably want to do here, is call `SetFixedMoveSpeed`. That works for both characters and monsters, saves you from doing any of this manually, and you can simply set the entity's speed to that of the owner, plus maybe a little on top.

system/scripts/zone/core/calc_combat.cs
```diff
@@ -197,6 +197,17 @@ public float SCR_CalculateDamage(ICombatEntity attacker, ICombatEntity target, S
 
 		SCR_Combat_AfterCalc(attacker, target, skill, modifier, skillHitResult);
 
+		// TODO: There may be a better place to check this
```
exectails, line 200: The better place is the buff handlers, from where you can hook into all this without having to add the code here^^ Notice the `SCR_Combat_AfterCalc` call just above, which makes the server call the handlers of all active buffs to see if they'd like to change anything about this skill hit.

system/scripts/zone/other/character_adv.cs
```diff
@@ -710,7 +710,6 @@ private static void GrantDefaults(Character character, JobId jobId)
 			{
 				LearnSkill(character, SkillId.Hammer_Attack);
 				LearnSkill(character, SkillId.Hammer_Attack_TH);
-				LearnSkill(character, SkillId.Sadhu_OutofBodyCancel);
```
exectails, line 713: Can you tell me why you removed this? It's part of the official data, which there probably is a reason for.

exectails (changes requested):
Thanks for the PR. I took a quick glance and I noticed that some of it still seems a bit prototypical, and I also have some questions and concerns before we continue. Please refer to my comments.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -43,6 +41,13 @@ public class Map : IUpdateable
 		/// </summary>
 		private readonly Dictionary<int, Character> _characters = new();
 
+		/// <summary>
+		/// Collection of dummy characters.
+		/// <para>Key: <see cref="Character.Handle"/></para>
+		/// <para>Value: <see cref="Character"/></para>
+		/// </summary>
+		private readonly Dictionary<int, Character> _dummies = new Dictionary<int, Character>();
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Some of Melia methods takes in consideration that a character on the _characters dictionary have a connection and dummies don't have it. That should be the main difference. Well, I am out of ideas on how to improve it. I don't think it's should be overcomplicated: is it a dummy? Instead of adding to the _characters dictionary add it to the _dummies. 

Broadcasts is an example that we should be threating characters differently from dummies:

```
		/// <summary>
		/// Broadcasts packet to all characters on map.
		/// </summary>
		/// <param name="packet"></param>
		public virtual void Broadcast(Packet packet)
		{
			lock (_characters)
			{
				foreach (var character in _characters.Values)
					character.Connection.Send(packet);
			}
		}
```

```
		/// <summary>
		/// Broadcasts packet to all characters on map, that are within
		/// visible range of source.
		/// </summary>
		/// <param name="packet">Packet to broadcast.</param>
		/// <param name="source">Reference actor for visible range to broadcast in.</param>
		/// <param name="includeSource">Send to source as well?</param>
		public virtual void Broadcast(Packet packet, IActor source, bool includeSource = true)
		{
			lock (_characters)
			{
				foreach (var character in _characters.Values.Where(a => (includeSource || a != source) && a.Position.InRange2D(source.Position, VisibleRange)))
					character.Connection.Send(packet);
			}
		}
```

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
```diff
@@ -404,12 +404,21 @@ private void CheckWarp()
...
+			if (this.MoveSpeedType == type)
+				return;
+
+			if (this.Entity is Mob mob)
 			{
 				this.MoveSpeedType = type;
 				this.Entity.Properties.Invalidate(PropertyName.MSPD);
 
 				Send.ZC_MSPD(this.Entity);
+			} else if (this.Entity is Character character && character.IsDummy)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: gotcha, I will try my best to slowly improve it.

src/ZoneServer/Network/Send.cs
```diff
@@ -148,13 +151,13 @@ public static void ZC_ENTER_PC(IZoneConnection conn, Character character)
...
 			packet.PutByte(0);
-			packet.PutShort(0);
+			packet.PutShort(isDummy ? 5 : 0);
 			packet.PutInt(-1); // titleAchievmentId
 			packet.PutInt(0);
 			packet.PutByte(0);
 			packet.AddAppearancePc(character);
 			packet.PutInt(0);
-			packet.PutFloat(405494.3f);
+			packet.PutFloat(isDummy ? 330012.6f : 405494.3f);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: yeah, this seems to be unnecessary (at least for now).

src/ZoneServer/Network/Send.cs
```diff
@@ -124,7 +125,9 @@ public static void ZC_MYPC_ENTER(Character character)
 		/// </summary>
 		/// <param name="conn"></param>
 		/// <param name="character"></param>
-		public static void ZC_ENTER_PC(IZoneConnection conn, Character character)
+		/// <param name="equipIds"></param>
+		/// <param name="isDummy"></param>
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: indeed!

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -66,6 +67,16 @@ internal void InitFor(ICombatEntity combatEntity)
 			if (combatEntity is Mob mob)
 				this.SetTendency(mob.Tendency);
 
+			if (combatEntity is Character character && character.IsDummy)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: gotcha!

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/SadhuBaseSkill.cs
```diff
@@ -0,0 +1,274 @@
...
+			dummyCharacter.InitProperties();
+			dummyCharacter.Properties.Stamina = (int)casterCharacter.Properties.GetFloat(PropertyName.MaxSta);
+			dummyCharacter.UpdateStance();
+			dummyCharacter.ModifyHpSafe(casterCharacter.MaxHp, out var hp, out var priority);
+
+			dummyCharacter.Owner = casterCharacter;
+
+			casterCharacter.Map.AddDummyCharacter(dummyCharacter);
+
+			Log.Info("[SpawnDummyClone] casterCharacter.IsDummy: {0}", casterCharacter.IsDummy);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Forgot to remove those!

src/ZoneServer/Skills/Skill.cs
```diff
@@ -199,7 +199,8 @@ public bool IncreaseOverheat(int overheatMaxCount)
 
 			// Update the overheat after the max was checked so we reset it
 			// to 0 if we went into cooldown
-			Send.ZC_OVERHEAT_CHANGED(character, this);
+			if (!character.IsDummy)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: yeah, I don't know the reason I put this check, it seems unnecessary!

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -326,7 +326,7 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
 				var teleportDistance = minDistance * 4;
 				var distance = followTarget.Position.Get2DDistance(this.Entity.Position);
 
-				if (distance > teleportDistance)
+				if (teleport && distance > teleportDistance)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: it doesn't seems like, it only lasts 3 seconds but I assume it doesn't.

src/ZoneServer/Network/Send.cs
```diff
@@ -3180,20 +3220,34 @@ public static void ZC_TRUST_INFO(IZoneConnection conn)
...
 			packet.PutByte(0);
 			packet.PutFloat(0);
 			packet.PutFloat(1);
 
-			// [i373230] Maybe added earlier
+			// [i373230] Maybe added earlier [Updated ib 28/07/24]
 			{
-				packet.PutByte(0);
-				packet.PutByte(0);
+				packet.PutShort(unknowBoolean ? 248 : 0);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Making a short makes the life easier since I could directly sign it to 248 if needed.
But it's part of the process, trial and error. I can't magically identify what's it's purpose so I left as a unknow.

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/SadhuBaseSkill.cs
```diff
@@ -0,0 +1,274 @@
...
+			var moveSpeedBonus = this.GetMoveSpeedBonus(skill);
+
+			var dummyCharacter = this.SpawnDummyClone(casterCharacter, farPos);
+
+			Send.ZC_PLAY_SOUND(caster, "skl_eff_yuchae_start_2");
+			Send.ZC_GROUND_EFFECT(casterCharacter, "I_only_quest_smoke013_blue_smoke", farPos, 1, 0.7f, 0, 0, 0);
+			Send.ZC_GROUND_EFFECT(casterCharacter, "I_only_quest_smoke013_blue_smoke", dummyCharacter.Position, 1.5f, 0.3f, 0, 0, 0);
+			Send.ZC_GROUND_EFFECT(casterCharacter, "I_only_quest_smoke058_blue", farPos, 3f, 0.5f, 0, 0, 0);
+			Send.ZC_GROUND_EFFECT(casterCharacter, "I_only_quest_smoke058_blue", dummyCharacter.Position, 3, 0.5f, 0, 0, 0);
+			Send.ZC_SET_POS(casterCharacter, dummyCharacter.Handle, farPos);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: No! The dummy character has no connection so I can't send packets through him...

src/ZoneServer/Buffs/Handlers/Clerics/Sadhu/OOBE_Anila_Buff.cs
```diff
@@ -0,0 +1,182 @@
...
+			this.ReturnToBody(casterCharacter, (int)buff.NumArg2);
+		}
+
+		/// <summary>
+		/// Remove the dummy character from the map
+		/// </summary>
+		/// <param name="character"></param>
+		private void RemoveDummyCharacter(Character character)
+		{
+			if (character.Owner is Character ownerCharacter)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: This may be useful for pets or blue orb monsters but I will change it anyway.

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
felipecastagnarodecarvalho: This is needed because DummyCharacters don't have connection so I need to pass the Owner to the other method.

system/scripts/zone/other/character_adv.cs
felipecastagnarodecarvalho: This skill is not being used anymore and has no purposes. Sadhu has been remake I think.

exectails: This is based on official data though, and removing it creates a dissonance between server and client. Unless this causes issues, it should stay this way.

src/ZoneServer/World/Maps/Map.cs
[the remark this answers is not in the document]
exectails: Well, you're saying,
> it's should be overcomplicated: is it a dummy? Instead of adding to the _characters dictionary add it to the _dummies.

but while duplicating dozens of lines of code is technically simple, it makes future maintenance and understanding the code much more complicated. And in a project that many more people will work on over many more years, that is far from ideal.

Instead, I would argue that an overall less complicated way is to view dummies as that which they are: characters without a connection. One way to handle that could be to check the character type before sending packets. But another, even more opaque way would be to give dummy characters a dummy connection. Simply a connection type that doesn't do anything when `Send` is called. That way it doesn't matter if you ever do send a packet for any reason, as it will effectively be ignored.

I would implement this, but looking at the current buff handlers, it's hard to tell how this would and should affect the existing code, so you would have to do it.

src/ZoneServer/Network/Send.cs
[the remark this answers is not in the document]
exectails: > I can't magically identify what's it's purpose

And yet you found the need to change this, so I must assume that you did "magically" figure out _something_^^ The alternative is that you changed this for no reason. Either way, it doesn't make sense, and it's unrelated to anyone understanding what these two bytes do in actuality. You didn't tell us why this was changed, what the change was supposed to accomplish, and as I described, you opened up a lot of potential questions, which is the real problem.

src/ZoneServer/Network/Send.cs
```diff
@@ -3236,15 +3275,23 @@ public static void ZC_PCBANG_POINT(IZoneConnection conn)
...
+		/// <param name="entity"></param>
 		public static void ZC_MSPD(ICombatEntity entity)
+			=> ZC_MSPD(entity, entity);
+
+		/// <summary>
+		/// Updates entity movement speed for another entity that has connection.
+		/// </summary>
+		/// <param name="entity"></param>
+		/// <param name="target"></param>
+		public static void ZC_MSPD(ICombatEntity entity, ICombatEntity target)
```
exectails, line 3289: Seeing all these sender overloads, I'm starting to wonder whether dummies perhaps _should_ be able to receive packets. If they were forwarded to their owner's client, you would potentially not need any of these.

src/ZoneServer/Network/Send.cs
```diff
@@ -4384,6 +4431,39 @@ public static void ZC_KNOCKDOWN_INFO(ICombatEntity entity, ICombatEntity target,
...
+		/// <param name="packetString"></param>
+		/// <param name="position"></param>
+		/// <param name="f1"></param>
+		/// <param name="f2"></param>
+		/// <param name="f3"></param>
+		/// <param name="f4"></param>
+		/// <param name="f5"></param>
+		public static void ZC_GROUND_EFFECT(ICombatEntity entity, string packetString, Position position, float f1, float f2, float f3, float f4, float f5)
+		{
+			if (!ZoneServer.Instance.Data.PacketStringDb.TryFind(packetString, out var packetStringData))
```
exectails, line 4447: Reminder that we have a helper extension for packet strings now, `Packet.AddStringId`. That way you don't need to get the packet data and its id yourself anymore.

src/ZoneServer/Pads/Handlers/Cleric/Sadhu/Sadhu_Anila_Effect_Pad.cs
```diff
@@ -0,0 +1,67 @@
...
+
+			this.MovePad(pad, args.Creator);
+		}
+
+		/// <summary>
+		/// Makes pad moves forwards
+		/// </summary>
+		/// <param name="pad"></param>
+		/// <param name="creator"></param>
+		private async void MovePad(Pad pad, ICombatEntity creator)
```
exectails, line 55: Just like we replaced the the async methods in the skill handlers, you shouldn't use `async void` anywhere else either, because these kinds of methods are unsafe. Any exception that occurs within will bring the whole server down. Instead, use `async Task` and use our `CallSafe` helper method to call them.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -276,7 +276,7 @@ protected IEnumerable Animation(string packetString)
 		/// <param name="minDistance">The minimum distance to the target the AI attempts to stay in.</param>
 		/// <param name="matchSpeed">If true, the entity's speed will be changed to match the target's.</param>
 		/// <returns></returns>
-		protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50, bool matchSpeed = false)
+		protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50, bool matchSpeed = false, bool teleport = true)
```
exectails, line 279: When adding or removing parameters, please remember to update the doc comments.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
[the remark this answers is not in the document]
exectails: The clone only lasts 3 seconds? In that case this seems like an unnecessary change, doesn't it? I mean, it's fine to have to option, so it's not a problem per se, but any change less also lessens to amount of time we're spending on reading and discussing changes within PRs^^

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/Sadhu_Anila.cs
```diff
@@ -0,0 +1,27 @@
...
+		/// Handles skill, makes the character out of body.
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="originPos"></param>
+		/// <param name="farPos"></param>
+		/// <param name="target"></param>
+		public void Handle(Skill skill, ICombatEntity caster, Position originPos, Position farPos, ICombatEntity target)
+		{
+			Handle(skill, caster, originPos, farPos, target, BuffId.OOBE_Anila_Buff);
```
exectails, line 24: Missing `this.`, coding conventions, yada yada^^

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/Sadhu_Skill_Base.cs
```diff
@@ -0,0 +1,271 @@
...
+		/// <summary>
+		/// Handles a sadhu skill for the given BuffId
+		/// </summary>
+		/// <param name="skill"></param>
+		/// <param name="caster"></param>
+		/// <param name="originPos"></param>
+		/// <param name="farPos"></param>
+		/// <param name="target"></param>
+		/// <param name="buffId"></param>
+		public void Handle(Skill skill, ICombatEntity caster, Position originPos, Position farPos, ICombatEntity target, BuffId buffId)
```
exectails, line 31: It's unfortunately not very clear what's going on here if you're not deep into this job already, made worse by all the duplicated code between `Handle` and `CreateClone`. Like with the buff handlers, I'd like to see a bit more structure that is easier to follow here.

src/ZoneServer/World/Actors/Characters/Character.cs
exectails: You marked this as resolved, but the issues are still there 🤔

system/scripts/zone/ais/sadhu_dummy.cs
```diff
@@ -0,0 +1,193 @@
...
+	protected IEnumerable Idle()
+	{
+		var movement = this.Entity.Components.Get<MovementComponent>();
+
+		movement.SetMoveSpeedType(MoveSpeedType.Run);
+		
+		if (this.Entity is Character entityCharacter && entityCharacter is DummyCharacter dummyCharacter)
+		{
+			SetFixedMoveSpeed(55);
+			entityCharacter.Properties.Modify(PropertyName.MSPD_BM, 55);
```
exectails, line 53: This code is redundant, as it's essentially what `SetFixedMoveSpeed` does.

system/scripts/zone/ais/sadhu_dummy.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		if (Case(80))
+		{
+			yield return MoveRandom();
+		}
+	}
+
+	protected IEnumerable Attack()
+	{		
+		// Remove the dummy character if the master is gone
```
exectails, line 76: You should not send raw packets from AI scripts unless there's absolutely no other way to accomplish what you need to do, because scripts, by their nature, should be simple to understand for users who are less experienced.

A cleaner solution would be to call some type of despawn method, which then internally figures out how to remove the dummy or whatever other actor the AI might be controlling.

exectails (changes requested):

system/scripts/zone/other/character_adv.cs
felipecastagnarodecarvalho: This WAS from the official data, but it got outdated, since the last changes on sadhu, the skill is not being used anymore.

src/ZoneServer/Buffs/Handlers/Clerics/Sadhu/OOBE_Anila_Buff.cs
```diff
@@ -0,0 +1,131 @@
...
+
+			var dummyCharacter = casterCharacter.Map.GetDummyCharacter((int)buff.NumArg2);
+
+			if (dummyCharacter != null)
+			{
+				dummyCharacter.Died += this.OnDummyDied;
+			}
+		}
+
+		public override void OnEnd(Buff buff)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I have added more comments and fixed a few missing function descriptions.
Also tried to clean up some of the code for buff handlers.

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/Sadhu_Skill_Base.cs
```diff
@@ -0,0 +1,271 @@
...
+		private float GetMoveSpeedBonus(Skill skill)
+		{
+			return skill.Id == SkillId.Sadhu_Prakriti ? 18 - (skill.Level + 6) : 18;
+		}
+
+		/// <summary>
+		/// Spawns a dummy character that will looks like the original character
+		/// </summary>
+		/// <param name="casterCharacter"></param>
+		private Character SpawnDummyClone(Character casterCharacter, Position position)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Good hint!

src/ZoneServer/Skills/Handlers/Clerics/Sadhu/Sadhu_Skill_Base.cs
felipecastagnarodecarvalho: Refactored some of the base skill sadhu class...

src/ZoneServer/Network/Send.cs
felipecastagnarodecarvalho: Dummies receiving packets? Does that make sense? It's not a client, who will be receiving it?
I need those because it has a different handle then the original character.

felipecastagnarodecarvalho:
@exectails this PR is ready to be reviewed again.
I have put some effort into improving the code.

src/ZoneServer/Pads/Handlers/Cleric/Sadhu/Sadhu_Anila_Effect_Pad.cs
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		/// <summary>
+		/// Makes pad moves forwards
+		/// </summary>
+		/// <param name="pad"></param>
+		/// <param name="creator"></param>
+		private async Task MovePad(Pad pad, ICombatEntity creator)
+		{
+			// Forward and back, hovering a moment in between.
```
exectails, line 58: The comment says forward _and back_ but that's not what's happening here. Could you clarify whether the error is in the code or the comment? If it's the comment that's incorrect, and the pad only moves forward, this comment could safely be removed, as the method appears clear-cut and is commented appropriately.

src/ZoneServer/World/Actors/Characters/DummyCharacter.cs
```diff
@@ -0,0 +1,37 @@
...
+	{
+		/// <summary>
+		/// Returns reference to the character's owner (In case of being a dummy).
+		/// </summary>
+		public Character Owner { get; set; }
+
+		/// <summary>
+		/// Returns true if the DummyCharacter has Owner
+		/// </summary>
+		public bool hasOwner => Owner != null;
```
exectails, line 24: There are still some convention violations, such as curly braces on the same line as a `foreach` or this camelCase property. If you're using Visual Studio, auto-formatting should solve some of these issues and the style checker should inform you about the rest.

src/ZoneServer/World/Actors/CombatEntities/Components/CooldownComponent.cs
```diff
@@ -49,7 +49,7 @@ public void Start(CooldownId cooldownId, TimeSpan duration)
 				}
 			}
 
-			if (this.Entity is Character character)
+			if (this.Entity is Character character && character is not DummyCharacter)
```
exectails, line 52: You're mixing conventions of checking for dummy characters before calling senders and inside them, as well as mixing type and null checks. I do understand what led to this, but it would be better to stay consistent. Better yet would be to not need this distinction.

src/ZoneServer/World/Actors/CombatEntities/Components/MovementComponent.cs
```diff
@@ -404,13 +404,23 @@ private void CheckWarp()
...
+				return;
+
+			if (this.Entity is Mob mob)
 			{
 				this.MoveSpeedType = type;
 				this.Entity.Properties.Invalidate(PropertyName.MSPD);
 
 				Send.ZC_MSPD(this.Entity);
 			}
+			else if (this.Entity is Character character && character is DummyCharacter dummyCharacter)
```
exectails, line 417: I would presume this check to be either incorrect or redundant. Am I right in that assessment? You're checking whether the entity is a character _and_ a dummy character, which would mean the code applies only to dummy characters, as only they are both. Though in that case it would be enough to check for the dummy character type. And if you wanted it to apply to both, a check for character would be enough, as the dummy inherits character iirc.

src/ZoneServer/World/Maps/Map.cs
```diff
@@ -794,7 +794,7 @@ public virtual void Broadcast(Packet packet)
 			lock (_characters)
 			{
 				foreach (var character in _characters.Values)
-					character.Connection.Send(packet);
+					character.Connection?.Send(packet);
```
exectails, line 797: I'm not a huge fan of this change, as it introduces a new type of character into the system that did not exist prior and might not be handled correctly in all cases: a character without connection. Before this, all characters could be presumed to have a connection during their entire session, from when they connect to when they disconnect. If that's no longer the case, we'd have to make sure that any access to the connection property, as well as information behind that, such as accounts, are properly checked before they're accessed.

While I recognize that this only applies to the Sadhu dummy characters right now, and that everything may very well work as intended for the moment, we will inevitably use the cloning methods you introduced for other purposes and other systems are going to have to work around these dummy characters, regardless of where they came from.

With that in mind, have you made sure that this radical change won't cause any issues in the existing code?

system/scripts/zone/other/character_adv.cs
exectails: That may be true, but it is in fact still part of the job data as of the last time I checked, which this script is based on.

Removing this makes updating the data more challenging, because you have to remember not to add this skill. Potential automated tools for this purpose would also have to be changed not to re-add it. And if it is actually still used for something without our knowledge, we would rather not remove it. Not to mention that any kind of backwards compatibility we might want to introduce would potentially require this line to be re-added.

For these reasons I asked whether there are any issues with leaving it as is, because that is the simplest solution for my concerns, unless we _have_ to conditionally remove it for some reason.

exectails (changes requested):
I appreciate your efforts for improving the code, though some concerns have not been cleared yet. Still, I might consider merging the PR soon, assuming you're still around and able to maintain the changes, fixing issues in a timely manner should they arise.

Given the nature of this class, the code is unfortunately not very straightforward to follow in its current form, and I suspect potential changes or improvements to it would be simpler to accomplish if one were to rewrite the skills. As their developer, I assume you would have a better overview of them and their behavior.

src/ZoneServer/Network/Send.cs
```diff
@@ -1318,6 +1329,22 @@ public static void ZC_UPDATED_PCAPPEARANCE(Character character)
 			character.Map.Broadcast(packet, character);
 		}
 
+		/// <summary>
+		/// Broadcasts ZC_UPDATED_PCAPPEARANCE of a targetCharacter
+		/// in range of character, updating their appearance.
+		/// </summary>
+		/// <param name="character"></param>
+		/// <param name="targetCharacter"></param>
+		public static void ZC_UPDATED_PCAPPEARANCE(Character character, Character targetCharacter)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: resolved.

src/ZoneServer/Network/Send.cs
```diff
@@ -1543,6 +1580,10 @@ public static void ZC_OBJECT_PROPERTY(IZoneConnection conn, IPropertyObject obj,
 		/// <param name="propertyList"></param>
 		public static void ZC_OBJECT_PROPERTY(IZoneConnection conn, long objectId, PropertyList propertyList)
 		{
+			// TODO: Find a better way to check this - This was done to avoid sending packets for dummies (while receiving buffs)
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: I don't think we would need a fake connection for dummies.
A connection means that a client exists behind it, right?
For the other side, we can send packets to the dummy map:
```
dummy.Map.Broadcast(packet);
```
If in the future the dummy plays an animation (when the dummy is selling equipament improvements and it needs to play the refinement animation, for example) we can use this without problem.

-----------

I have move this check to InitEvents() method  on CharacterProperties.cs:
```
		/// <summary>
		/// Sets up event subscriptions, to react to actions of the
		/// character with property updates.
		/// </summary>
		private void InitEvents()
		{
			// This was done to avoid sending packets for dummies (while receiving buffs)
			if (this.Character.Connection == null)
				return;
			
			[...]
		}
```

src/ZoneServer/Network/Send.cs
```diff
@@ -3254,26 +3309,35 @@ public static void ZC_TRUST_INFO(IZoneConnection conn)
...
 			packet.AddStringId(animationName);
 			packet.PutByte(stopOnLastFrame);
 			packet.PutByte(0);
 			packet.PutFloat(0);
 			packet.PutFloat(1);
 
-			// [i373230] Maybe added earlier
+			// [i373230] Maybe added earlier [Updated at 28/07/24]
 			{
-				packet.PutByte(0);
```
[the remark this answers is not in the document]
felipecastagnarodecarvalho: Changes reverted.

src/ZoneServer/Pads/Handlers/Cleric/Sadhu/Sadhu_Anila_Effect_Pad.cs
felipecastagnarodecarvalho: wrong commentary. Fix applied.

src/ZoneServer/World/Actors/Characters/DummyCharacter.cs
felipecastagnarodecarvalho: Done.

src/ZoneServer/World/Maps/Map.cs
felipecastagnarodecarvalho: This is a bit confused for me.
Last time we talked you said that it would be better to keep using the on going Characters instead of creating a new dictionary:
```
private readonly Dictionary<int, Character> _dummiesCharacter = new();
```

That what I did last time and I have refactored to keep using the current dictionary.
I am not sure if this approach will work still with vending dummies character but it should!
It would require a few changes but mainly it's the same.

felipecastagnarodecarvalho:
@exectails I have update this PR with:
  - Correct OOBE animation for the dummy.
  - Improvements on the AI for the cloned character (can be used if you have [Arts] Spirit Expert: Wandering Soul).
  - Fixed all mentioned commentaries.
  - Fixed a bug that the monster wouldn't be alerted by the cloned character when taking damage.

PR #306, opened 2024-08-18
Pathfinding follower fix

kenedos:
- Removal of randomFactor from DynamicGridPathfinding heuristic (so entities move in deterministic way)
- Fix for follower pathfinding
- Changed ai.conf to use pathfinding by default

### DIFF
--- a/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
+++ b/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
@@ -291,7 +291,7 @@ namespace Melia.Zone.Scripting.AI
 
 				if (catchUp)
 				{
-					var closePos = this.Entity.Position.GetRelative(followTarget.Position, 50);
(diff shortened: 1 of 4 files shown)


src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -291,7 +291,7 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
 
 				if (catchUp)
 				{
-					var closePos = this.Entity.Position.GetRelative(followTarget.Position, 50);
+					var closePos = followTarget.Position;
```
kenedos, line 294: GetRelative() is not correct when we use pathfinding now, because the closest position may no longer be within a straight line to the master/target.

Furthermore, isTargetInRange is already finding the distance to target/master so the follower stops moving at 50 distance either way, so we don't need any additional code change.

PR #309, opened 2024-08-19
Fix for threat calculation

kenedos:
In AiScript.cs, the hate given per hit can be exploited if user has Liberate buff and monster has less than _minAggroHateLevel amount of hate towards the player.

If the player with Liberate hits a monster with zero hate, they will immediately build 500 hate towards them. If they hit a monster with 100 or more hate towards them, there is no issue as that hate value will be multiplied by _overHateRate.

This PR addresses this issue by adding a simple check at the end of IncreaseHate() and multiplying any threat amount past the limit by the _overHateRate.

### DIFF
--- a/src/ZoneServer/Scripting/AI/AiScript.cs
+++ b/src/ZoneServer/Scripting/AI/AiScript.cs
@@ -235,6 +235,14 @@ namespace Melia.Zone.Scripting.AI
 			if (entity.Components.Get<BuffComponent>().Has(BuffId.Liberate_Buff))
 				amount *= 5;
 
+			if (amount > _minAggroHateLevel)
+			{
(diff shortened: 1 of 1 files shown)


exectails:
Good catch, but I believe the proposed solution doesn't fix the issue entirely. It might for very large amounts, but not necessarily all, right? You're checking the amount to be added against the min aggro level, but if you were at, say, 99 hate, and we added 100 from somewhere to cause aggro, we'd then be at 199 despite the check. That's not quite as bad as with liberate, but still more than it should be if we go by the logic that everything past 100 should be adjusted. I guess we might also consider adding a method that just pushes the AI over the edge, regardless of the current hate 🤔 But that's a different issue. For now I'm going push a fix for this. Thanks for bringing it up 👍

PR #310, opened 2024-08-22
Added AI changing targets according to hate levels

kenedos:
This PR implements monsters changing targets according to their hate levels in GetMostHated().

[Video](https://www.youtube.com/watch?v=jLmEwBT2P8c)

### DIFF
--- a/system/scripts/zone/ais/basic_monster.cs
+++ b/system/scripts/zone/ais/basic_monster.cs
@@ -14,6 +14,7 @@ public class BasicMonsterAiScript : AiScript
 	protected override void Setup()
 	{
 		During("Idle", CheckEnemies);
+		During("Attack", CheckEnemies);
 		During("Attack", CheckTarget);
 		During("Attack", CheckMaster);
 	}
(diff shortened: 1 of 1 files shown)


system/scripts/zone/ais/basic_monster.cs
```diff
@@ -89,7 +90,7 @@ protected IEnumerable StopAndAttack()
 	private void CheckEnemies()
 	{
 		var mostHated = GetMostHated();
-		if (mostHated != null)
+		if (mostHated != null && (target == null || mostHated != target))
```
exectails, line 93: Quick question, wouldn't the check in `CheckTarget` be enough? We're calling that function anyway, even the name matches, and we could just do this, no?
```text
if target != GetMostHated
    changeTarget
```

system/scripts/zone/ais/basic_monster.cs
```diff
@@ -14,6 +14,7 @@ public class BasicMonsterAiScript : AiScript
 	protected override void Setup()
 	{
 		During("Idle", CheckEnemies);
+		During("Attack", CheckEnemies);
```
exectails, line 17: By adding this event callback, as well as a call for `CheckEnemies` to the `CheckTarget` function, it would be called twice right now. Probably not an issue, but unnecessary.

system/scripts/zone/ais/basic_monster.cs
exectails: I changed the code a little to make it more straight forward, with a single check in `CheckTarget`, where we're already handling the current target and simply change it if applicable. Let me know if I missed something in your implementation.

PR #312, opened 2024-08-24
Companions - WIP

SalmanTKhan:
* Base Implementation
 - Loading/Saving both in Barracks and Zone Server
 - Able to mount companions with proper class (Falcon/Flying pets don't work yet).
* Known Issues
 - Changing map while riding, you have to remount the companion.
 - Companion list in barracks works for 1 companion, after that it breaks.
 - Companion Buff Logic has to be refactored.


exectails:
Thanks for the PR! I'm going to convert this to a draft for now, because of the known issues, which sound like this isn't _quite_ ready to be pulled into master. The barracks issue in particular sounds like the packet(s) aren't correct yet. Did you base your implementation on logs or is the structure guessed?

SalmanTKhan:
> Thanks for the PR! I'm going to convert this to a draft for now, because of the known issues, which sound like this isn't _quite_ ready to be pulled into master. The barracks issue in particular sounds like the packet(s) aren't correct yet. Did you base your implementation on logs or is the structure guessed?

Sounds good.

As always based-off packet logs and then guessed, because you know there are certain nuances like a byte + 3 gap/filler bytes, that are common in ToS packet structure, that I tend to miss. There used to be Pet Weapon/Armor slots in the Barrack PetInfo packet but as you know IMC removed them and now I have no clue what's in there. The structure seems fine in the packet template but when applying it to the actual in-game, it doesn't work.

exectails:
Based on the discussion on Discord, it seems like the second known issue is fixed now, and I also see a commit addressing buff issues. Does that mean only the first known issue (remounting) is left?

SalmanTKhan:
> Based on the discussion on Discord, it seems like the second known issue is fixed now, and I also see a commit addressing buff issues. Does that mean only the first known issue (remounting) is left?

That is correct

system/scripts/zone/core/calc_companion.cs
```diff
@@ -0,0 +1,394 @@
...
+using Melia.Zone.Scripting;
+using Melia.Zone.World.Actors;
+using Melia.Zone.World.Actors.Monsters;
+using Melia.Zone.World.Maps;
+
+public class CompanionCalculationsScript : GeneralScript
+{
+	private const float PET_STAT_BY_OWNER_RATE = 0.5f;
+
+	public float PET_STAT_BY_OWNER(Companion companion, string statName)
```
exectails: I assume that you're imitating official names with the constant and this function, but generally speaking, using upper case here goes entirely against our normal naming conventions, and we presumably also won't need these for reference. As such, the normal PascalCase would be more appropriate.

system/scripts/zone/core/calc_companion.cs
```diff
@@ -0,0 +1,394 @@
...
+	/// </summary>
+	/// <param name="companion"></param>
+	/// <returns></returns>
+	[ScriptableFunction]
+	public float SCR_Get_Companion_DEX(Companion companion)
+	{
+		var ownerProperties = companion.Owner.Properties;
+
+		var baseValue = companion.Data.DEX;
+		var ownerValue = ownerProperties.GetFloat(PropertyName.DEX) * PET_STAT_BY_OWNER_RATE;
```
exectails: Am I missing something or should these use `PET_STAT_BY_OWNER`?

src/BarracksServer/Database/Account.cs
```diff
@@ -330,5 +335,47 @@ public void Save()
...
+
+		/// <summary>
+		/// Adds companion to account object and assigns index.
+		/// </summary>
+		/// <param name="companion"></param>
+		private void AddCompanion(Companion companion)
+		{
+			lock (_companions)
+			{
+				for (byte i = 1; i <= byte.MaxValue; ++i)
```
exectails: While almost certainly not an issue in this case, C# conventions and our guidelines advocate for not using non-`int`s if not necessary for interoperability. `var` is also to be preferred.

src/BarracksServer/Database/BarracksDb.cs
```diff
@@ -571,5 +573,85 @@ public void SaveItem(Character character, long itemId)
...
+						cmd.Set("characterId", null);
+
+					cmd.Execute();
+				}
+				trans.Commit();
+			}
+		}
+
+		/// <summary>
+		/// Deletes a companion.
```
exectails: While the purpose of this return value can be guessed, potentially ambiguous values should be clarified.

src/BarracksServer/Network/Helpers/BarrackPetHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+	/// Contains extensions for writing barracks character data to packets.
+	/// </summary>
+	public static class BarrackPetHelper
+	{
+		/// <summary>
+		/// Writes the character's data to the packet.
+		/// </summary>
+		/// <param name="packet"></param>
+		/// <param name="companion"></param>
+		/// <exception cref="InvalidOperationException"></exception>
```
exectails: It doesn't seem like this function would throw this exception, right?

src/BarracksServer/Network/PacketHandler.cs
```diff
@@ -692,7 +705,25 @@ public void CB_PET_COMMAND(IBarracksConnection conn, Packet packet)
...
+			{
+				Log.Warning("CB_PET_COMMAND: Companion not found by id '{0}' received from '{1}'.", companionId, conn.Account.Name);
+				return;
+			}
+
+			companion.CharacterDbId = characterId;
+			switch (command)
+			{
+				// Delete
+				case 1:
```
exectails: When turning types into actionable values, it's generally better to create an enum for better readability and type safety. It would also be useful to log unknown values in the default case or as result of an `Enum.IsDefined` check, so we know when new, previously unknown values, come in.

src/Shared/Data/Database/Companions.cs
```diff
@@ -0,0 +1,59 @@
...
+			// Values: id, className, name, rideMSPD, premium, pet, ride, endRideOnHit, [foodGroup], [buff]
+			entry.AssertNotMissing("id", "className");
+
+			var data = new CompanionData();
+
+			data.Id = entry.ReadInt("id");
+			data.ClassName = entry.ReadString("className");
+			data.Name = entry.ReadString("name");
+			if (entry.ContainsKey("price"))
+				data.Price = entry.ReadInt("price");
```
exectails: Is there a reason you're explicitly checking for the existence of keys instead of utilizing the default value overload of the read methods?

src/Shared/Data/Database/Companions.cs
```diff
@@ -0,0 +1,59 @@
...
+		public bool TryFindByClassName(string className, out CompanionData data)
+		{
+			data = this.Entries.Values.FirstOrDefault(a => a.ClassName == className);
+			return data != null;
+		}
+
+		protected override void ReadEntry(JObject entry)
+		{
+			// Values: id, className, name, rideMSPD, premium, pet, ride, endRideOnHit, [foodGroup], [buff]
+			entry.AssertNotMissing("id", "className");
```
exectails: Based on the comment above and your `Contains` calls belows, it seems like you're expecting quite a few more values to be defined. Should these not be part of this assert?

src/Shared/Data/Database/Monsters.cs
```diff
@@ -30,6 +30,12 @@ public class MonsterData
 		public int Exp { get; set; }
 		public int JobExp { get; set; }
 
+		public int STR { get; set; }
```
exectails: We generally try to avoid the upper case naming-scheme unless there's a good reason for it, as PascalCase is more idiomatic to C#. We're also already using `Str`, `Con`, etc. in other places, which makes these stand out a little.

src/ZoneServer/Commands/ChatCommands.Handlers.cs
```diff
@@ -327,6 +330,118 @@ private CommandResult HandleJump(Character sender, Character target, string mess
...
+			if (!int.TryParse(args.Get(0), out var petShopId))
+				return CommandResult.InvalidArgument;
+
+			if (!ZoneServer.Instance.Data.CompanionDb.TryFind(petShopId, out var data))
+				return CommandResult.InvalidArgument;
+
+			if (!ZoneServer.Instance.Data.MonsterDb.TryFind(data.ClassName, out var monData))
+				return CommandResult.InvalidArgument;
+
+			var targetNpcName = "Companion Trader";
```
exectails: Correct me if I'm wrong, but won't this check fall apart if the NPC used is not explicitly named `Companion Trader`? It seems like this would be a bit of issue. Not only in cases where the admin might rename NPCs, but also when translations are applied.

src/ZoneServer/Commands/ChatCommands.Handlers.cs
```diff
@@ -327,6 +330,118 @@ private CommandResult HandleJump(Character sender, Character target, string mess
...
+			if (string.IsNullOrEmpty(currentNpcName) || !currentNpcName.Contains(targetNpcName))
+				return CommandResult.InvalidArgument;
+
+			//TODO: Decide if we sell pets that don't have a price defined.
+			if (data.Price == 0)
+				return CommandResult.Okay;
+
+			if (sender.Inventory.CountItem(ItemId.Silver) < data.Price)
+			{
+				sender.SystemMessage("OwnerDontHaveSilver");
```
exectails: Since we decided not to rely on client-side translation, messages should be written out and wrapped in localization calls, instead of using system messages.

src/ZoneServer/Commands/ChatCommands.Handlers.cs
```diff
@@ -327,6 +330,118 @@ private CommandResult HandleJump(Character sender, Character target, string mess
...
+				var propertyName = "Monster_Stat_" + args.Get(1);
+
+				if (companion != null && PropertyTable.Exists("Monster", propertyName)
+					&& int.TryParse(args.Get(2), out var modifierValue))
+				{
+					var baseCost = propertyName == PropertyName.Stat_DEF ? 600 : 300;
+					var totalCost = 0;
+					var currentValue = companion.Properties.GetFloat(propertyName) - 1;
+
+					for (var i = currentValue; i < (currentValue + modifierValue); i++)
```
exectails: This feels like it might be a good candidate for a scriptable function to allow for easy customization. What do you think?

src/ZoneServer/Database/ZoneDb.cs
```diff
@@ -1443,5 +1446,100 @@ private void LoadCollections(Character character)
...
+					cmd.Set("stamina", companion.Stamina);
+					cmd.Set("exp", companion.Exp);
+					cmd.Set("adoptTime", companion.AdoptTime);
+					cmd.Set("active", companion.IsActivated ? 1 : 0);
+
+					cmd.Execute();
+					companion.DbId = cmd.LastId;
+				}
+
+				trans.Commit();
```
exectails: While not an issue per se, you don't need a transaction if there's only one query.

src/ZoneServer/World/Actors/Monsters/Companion.cs
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		/// <summary>
+		/// Grants exp to companion.
+		/// </summary>
+		/// <param name="exp"></param>
+		/// <param name="monster"></param>
+		public void GiveExp(long exp, IMonsterBase monster)
+		{
+			// Base EXP
```
exectails: Comments like this are generally somewhat pointless, because they don't add any information over the code. Common wisdom: Describe the "why," not the "what."

❌
```cs
// Add base EXP
this.Exp += exp;
```

✔
```cs
this.Exp += exp;
```

✔
```cs
// Add EXP here instead of there because reason X.
this.Exp += exp;
```

src/ZoneServer/World/Actors/Monsters/Monster.cs
```diff
@@ -54,6 +54,16 @@ public interface IMonsterBase : IActor, IMonsterAppearance, IMonsterAppearanceBa
...
 		/// </summary>
 		Properties Properties { get; }
+
+		/// <summary>
+		/// Returns the owner's handle.
+		/// </summary>
+		int OwnerHandle { get; }
+
+		/// <summary>
+		/// Returns an associated handle.
```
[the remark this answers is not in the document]
SalmanTKhan: I don't really have a better comment here. I'm open to suggestions; usually, it's the same as Owner Handle. Owner Handle + Associated Handle is generally used for Mobs (Turrets, Summons, Attacking Traps) while Npcs (Portals for Sage Skill) use Associated Handle only.

src/ZoneServer/Commands/ChatCommands.Handlers.cs
SalmanTKhan: You are correct, but it was an attempt to prevent abuse of the /pethire command without using the "Npc". I can remove the check for now.

src/ZoneServer/World/Actors/Monsters/Monster.cs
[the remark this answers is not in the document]
exectails: That^^ That's what should be documented, an explanation of what this is. And if it seems like a lengthy explanation might be overkill for the summary, you can add it as a remark. Just so anyone who looks at this method can understand what an associated handle may be, without having to do a reference search and check where and how it's used manually.

PR #314, opened 2024-08-26
Pathfinding improvements

kenedos:
- Removal of randomFactor from DynamicGridPathfinding heuristic (so entities move in deterministic way)
- Fix for follower pathfinding
- Changed ai.conf to use pathfinding by default
- added merging of path nodes
- AI improvement to only calculate pathfinding if player moves
- trace path to positions adjacent to player (not just exact player position)

### DIFF
--- a/src/Shared/World/Position.cs
+++ b/src/Shared/World/Position.cs
@@ -112,6 +112,20 @@ namespace Melia.Shared.World
 		}
 
 		/// <summary>
+		/// Returns true if the given position is inside a circle defined by
+		/// a center position and radius in 2D space.
+		/// </summary>
+		/// <param name="center"></param>
+		/// <param name="radius"></param>
+		/// <returns></returns>
+		public readonly bool InCircle2D(Position center, float radius)
+		{
+			var dx = X - center.X;
+			var dz = Z - center.Z;
+			return (dx * dx + dz * dz) <= (radius * radius);
+		}
+
+		/// <summary>
 		/// Returns random position around this position,
 		/// not nearer than min, and not further than max.
 		/// </summary>
--- a/src/Shared/World/Position.cs
+++ b/src/Shared/World/Position.cs
@@ -112,20 +112,6 @@ namespace Melia.Shared.World
 		}
 
 		/// <summary>
-		/// Returns true if the given position is inside a circle defined by
-		/// a center position and radius in 2D space.
-		/// </summary>
-		/// <param name="center"></param>
-		/// <param name="radius"></param>
-		/// <returns></returns>
-		public readonly bool InCircle2D(Position center, float radius)
-		{
-			var dx = X - center.X;
-			var dz = Z - center.Z;
-			return (dx * dx + dz * dz) <= (radius * radius);
-		}
-
-		/// <summary>
 		/// Returns random position around this position,
 		/// not nearer than min, and not further than max.
 		/// </summary>
--- a/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
+++ b/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
@@ -291,7 +291,7 @@ namespace Melia.Zone.Scripting.AI
 
 				if (catchUp)
 				{
-					var closePos = this.Entity.Position.GetRelative(followTarget.Position, 50);
+					var closePos = followTarget.Position;
 					yield return this.MoveTo(closePos, false);
 				}
 				else if (movement.IsMoving)
--- a/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
+++ b/src/ZoneServer/Scripting/AI/AiScript.Routines.cs
@@ -71,6 +71,28 @@ namespace Melia.Zone.Scripting.AI
 		}
 
 		/// <summary>
+		/// Moves the entity to an attack position around the given target.
+		/// </summary>
+		/// <param name="target"></param>
+		/// <param name="attackRange"></param>
+		/// <returns></returns>
+		protected IEnumerable MoveToAttackPosition(ICombatEntity target, float attackRange)
+		{
+			while (!this.InRangeOf(target, attackRange))
+			{
+				if (_lastPathfindingGoal == Position.Zero || !target.Position.InRange2D(_lastPathfindingGoal, 10))
+				{
+					_lastPathfindingGoal = this.GetAdjacentValidPosition(target, attackRange);
+					yield return this.MoveTo(_lastPathfindingGoal, wait: false);
+				}
+				else
+				{
+					yield return this.Wait(100);
+				}
+			}
+		}
+
+		/// <summary>
 		/// Moves entity to the given destination in a straight line.
 		/// </summary>
 		/// <param name="destination"></param>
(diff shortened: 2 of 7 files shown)


src/ZoneServer/Scripting/AI/AiScript.Routines.cs
```diff
@@ -291,7 +291,7 @@ protected IEnumerable Follow(ICombatEntity followTarget, float minDistance = 50,
 
 				if (catchUp)
 				{
-					var closePos = this.Entity.Position.GetRelative(followTarget.Position, 50);
+					var closePos = followTarget.Position;
```
exectails, line 294: Will this not make the following actor always run to the target's exact position, and potentially glue companions and other followers to their masters?

system/scripts/zone/ais/basic_monster.cs
```diff
@@ -50,24 +52,29 @@ protected IEnumerable Idle()
...
 			if (!TryGetRandomSkill(out var skill))
 			{
 				yield return Wait(2000);
 				continue;
 			}
-
-			while (!InRangeOf(target, skill.GetAttackRange()))
-				yield return MoveTo(target.Position, wait: false);
-
+			var attackRange = skill.GetAttackRange();
```
exectails, line 62: I feel like this might be a bit much to put into the AI script, especially if it would potentially have to be copied to other AIs. Perhaps it would be better to move it to `AiScript`, as a routine to get into attacking position? Something for us to think about.

system/scripts/zone/ais/basic_monster.cs
```diff
@@ -50,24 +52,29 @@ protected IEnumerable Idle()
 	protected IEnumerable Attack()
 	{
 		SetRunning(true);
-
```
exectails, line 53: Just for my understanding, why are you removing all sensible spacing and grouping in this method?^^

exectails (commented):
Thanks for the PR. It looks good generally, just got a few questions for my better understanding.

src/Shared/World/Position.cs
```diff
@@ -111,6 +111,20 @@ public bool InPolygon2D(Position[] points)
...
 		}
 
+		/// <summary>
+		/// Returns true if the given position is inside a circle defined by
+		/// a center position and radius in 2D space.
+		/// </summary>
+		/// <param name="center"></param>
+		/// <param name="radius"></param>
+		/// <returns></returns>
+		public readonly bool InCircle2D(Position center, float radius)
```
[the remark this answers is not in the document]
kenedos: Yup, you're right. Slight oversight by me.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
kenedos: This doesn't happen because when they get close enough `isTargetInRange` will return true then the follower will stop catching up.

Also getting the relative position was not exactly correct because that position could be out of bounds and not exactly the position the follower should be in relation to the master.

system/scripts/zone/ais/basic_monster.cs
exectails: You're right in saying that this might not become a one-size-fits-all method, because different types of mobs will need other behavior and will then get their own AIs, but don't forget that this will also not stay the only melee AI script for normal monsters.

Not only are there hundreds of official AIs we might eventually want/need to reproduce, it was also always the intention to facilitate the creation of more varied custom AIs, to give monsters more of an identity, so they don't all act the exact same. And anyone who will write a custom AI in the future, ourselves included, might appreciate a simple to use method, or a collection of them, to easily get their monster into a certain attacking position.

What do you think?

kenedos: Sounds good, i'll create a new method for it

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
kenedos: What exactly is concerning you or you think we could improve here?

system/scripts/zone/ais/basic_monster.cs
exectails: Oh, also, I think it would be nice if we kept the scripts as simple and readable as possible, as they're typically intended to be customizable by beginners. `MoveRandom` for example is only used once in the basic AI as well, and who knows if other AIs are going to use it as is, but having that raw code inside the script would make that process much less straightforward.

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
exectails: Don't worry about it, I'm overthinking unlikely potential issues. Like, a follower _could_ move fast enough to get close to the target before stopping for example. But in practice that will probably never happen.

system/scripts/zone/ais/basic_monster.cs
kenedos: I've added MoveToAttackPosition() method to simplify the script.

src/ZoneServer/Scripting/AI/AiScript.cs
```diff
@@ -626,6 +629,27 @@ protected bool EntityGone(ICombatEntity entity)
...
+		/// Gets a valid position adjacent to target within given range.
+		/// Returns entity's own position if not found.
+		/// </summary>
+		/// <param name="entity"></param>
+		/// <returns></returns>
+		protected Position GetAdjacentValidPosition(ICombatEntity entity, float range)
+		{
+			var ground = entity.Map.Ground;
+
+			for (var i = 0; i < 10; i++)
```
exectails, line 642: Just one more question, do we really think that monsters choose an entirely random position _somewhere_ around their target? From what I've seen, they absolutely do choose an adjacent position at times, but to me it seemed to always be between the attacker and the target, and not like they would potentially decide to pass you entirely and attack from behind. If that happened on officials, I would rather assume that the AI overshot their target accidentally 🤔

src/ZoneServer/Scripting/AI/AiScript.Routines.cs
kenedos: Well what we could try to do here is a GetRelative between the target's position and the _lastPathfindingGoal, since we know the pathfinding calculated position would be valid and on the right angle towards the target. It would remain the possibility of running out of bounds if not using pathfinding, though.

exectails: I don't know if that's actually a concern, because we naturally want pathfinding to be on, and the option to disable it will most likely get removed eventually. Additionally, if we wanted to to keep it, we would then of course have to make sure monsters don't run through walls.

src/ZoneServer/Scripting/AI/AiScript.cs
exectails: I'm currently writing an alternative to this function, but for reference I wanted to mention that I think there's another potential issue with this method as the source for attack positions, in that monsters could randomly run back and forth as the player moves, alternating attacking them from the front or back, because a random position is chosen every time the player moves at least 10 units. It would probably not happen a lot in practice, but it might look weird when it does happen.

PR #315, opened 2024-08-26
Cataphract Skills

Terotrous:
Adding all skills for Cataphract.  These all need a mount to function.

There are two extra packets in Send.cs that I had implemented as a test a long time ago.  I believe they are correct, but I can also remove them if they are undesired, this class doesn't need them (this is just the first time I've had to commit my Send.cs for a while).


exectails:
Thanks for the PR. Since it essentially depends on #312, and that one is still a draft and will presumably require changes, I'm going to make this one a draft for the time being as well, to indicate that it's not ready to be merged just yet. I'd like to ask you to go over it again once we have companions merged, to make sure everything is still working, and afterwards we'll continue with the usual review.

PR #320, opened 2024-09-15
Basic monster attack cancelling

Terotrous:
Adds attack cancelling for monsters based on locks.  Nothing for players yet.

Also adds the Knockdown locktype for Doppelsoldner_Punish.

### DIFF
--- a/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
+++ b/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
@@ -103,6 +103,10 @@ namespace Melia.Zone.Skills.Combat
 			target.Position = this.KnockBackInfo.ToPosition;
 
 			target.AddState(StateType.KnockedBack, this.KnockBackInfo.Time);
+
+			// Currently we consider knockdowns to also be knockbacks
+			// so any knockback-specific functionality also applies to them.
+			// Note that some skills check for knockdowns specifically.
 			if (this.HitInfo.Type == HitType.KnockDown)
 				target.AddState(StateType.KnockedDown, this.KnockBackInfo.Time);
 		}
--- a/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
+++ b/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
@@ -99,6 +99,13 @@ namespace Melia.Zone.Skills.Combat
 			if (this.KnockBackInfo == null)
 				throw new InvalidOperationException("Knock back info is not set.");
 
+			// Cannot suffer additional knockback while knocked down
+			if (target.IsStateActive(StateType.KnockedDown))
+			{
+				this.KnockBackInfo = null;
+				return;
+			}
+
 			this.HitInfo.Type = this.KnockBackInfo.HitType;
 			target.Position = this.KnockBackInfo.ToPosition;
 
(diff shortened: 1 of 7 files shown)


src/ZoneServer/World/Actors/Components/StateLockComponent.cs
```diff
@@ -50,6 +50,7 @@ public StateLockComponent(IActor owner)
 
 			this.RegisterState(new(StateType.Stunned, [LockType.Movement, LockType.Attack]));
 			this.RegisterState(new(StateType.KnockedBack, [LockType.Movement, LockType.Attack]));
+			this.RegisterState(new(StateType.KnockedDown, [LockType.Movement, LockType.Attack]));
```
exectails, line 53: There was no `KnockDown` lock state yet because knock backs and knock downs are conceptually the same in terms of locks in most games. Did you find a reason to differentiate them here?

src/ZoneServer/Skills/Combat/SkillHitInfo.cs
```diff
@@ -99,10 +99,12 @@ public void ApplyKnockBack(ICombatEntity target)
 			if (this.KnockBackInfo == null)
 				throw new InvalidOperationException("Knock back info is not set.");
 
-			this.HitInfo.Type = this.Skill.Data.KnockDownHitType;
+			this.HitInfo.Type = this.KnockBackInfo.HitType;
 			target.Position = this.KnockBackInfo.ToPosition;
 
 			target.AddState(StateType.KnockedBack, this.KnockBackInfo.Time);
+			if (this.HitInfo.Type == HitType.KnockDown)
```
exectails, line 106: If you wanted to add a knock down lock state, presumably to differentiate them, why add both states here?

exectails (commented):
Thanks for the PR. I guess most of these changes are pretty general things and could just be merged, without any additional research into what cancellations look like on officials. Just got a couple questions.

src/ZoneServer/Skills/Combat/SkillHitInfo.cs
Terotrous: My interpretation is that knockdown is kind of like a more severe version of knockback.  I think it makes sense that if certain skill or something needs to check if you're in knockback state, it should also accept a knockdown, but not vice-versa.

exectails: I won't argue that your interpretation doesn't make sense, but it does go against the grain a little, because games often don't care about the difference between a KB and a KD. Especially when it comes to locks, which these calls are all about. The reason this bothers me is that your code _suggests_ that there is a difference, which we technically don't know yet afaik, As such, I'd suggest to comment it, to clarify why we're doing what we're doing here.

src/ZoneServer/World/Actors/Components/StateLockComponent.cs
exectails: Alright, good to know!

src/ZoneServer/Skills/Combat/SkillHitInfo.cs
Terotrous: I added a comment.  If you think it doesn't make sense to do it this way you can also change it.  It's true that currently, we're not aware of anything that needs this distinction.

exectails:
I'm going to merge this for now, since the changes are relatively straight forward. We could leave it open until we know even more about the requirements and have a more proper implementation of skill cancellation, but we can just open a new PR then.

PR #322, opened 2024-09-17
No knockback during knockdown

Terotrous:
Very simple change based on Amazon's testing.

### DIFF
--- a/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
+++ b/src/ZoneServer/Skills/Combat/SkillHitInfo.cs
@@ -99,6 +99,13 @@ namespace Melia.Zone.Skills.Combat
 			if (this.KnockBackInfo == null)
 				throw new InvalidOperationException("Knock back info is not set.");
 
+			// Cannot suffer knockback while knocked down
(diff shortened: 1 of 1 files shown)


src/ZoneServer/Skills/Combat/SkillHitInfo.cs
```diff
@@ -99,6 +99,13 @@ public void ApplyKnockBack(ICombatEntity target)
 			if (this.KnockBackInfo == null)
 				throw new InvalidOperationException("Knock back info is not set.");
 
+			// Cannot suffer knockback while knocked down
```
exectails, line 102: The placement of this check seems pretty good, but I think we might rather want to check for locks instead of states, in case we'll have other situations where we want to prevent knock backs/downs, like buffs. events, or what have you. For this purpose we would introduce a lock type like `GetKnockedDown` for example, similar to the already existing `GetHit`.

Also, the lack of additional checks implies that while you can't KB during KD, you can KD during KB, KD during KD, and KB during KB. Is that correct? Our analysis seemed to indicate that the triple KB we saw was reduced to a single KB server-side, didn't it?

exectails (changes requested):

PR #323, opened 2024-09-27
Minor fix to skill tags

Terotrous:
Super small fix to skill tags, the correct tag is "NormalSkill", not "NormalAttack".


SalmanTKhan:
`public bool IsNormalAttack => this.Data.Tags.Has(SkillTag.NormalSkill);`
While you're at it, might as well use the consts added for SkillTags

exectails:
Woops. Thanks for the PR!

PR #324, opened 2024-09-28
Minor Archer Fixes

Terotrous:
Fulldraw now plays correct audio for female characters.
Oblique Shot now sends the bounce shot's damage packet correctly.


exectails:
Thanks for the PR 👍

PR #330, opened 2024-10-10
Initial Card System Support

SalmanTKhan:
* Currently cards can be equipped/unequipped and synthesized to increase card level.
* Loading and Saving Cards works via a Cards Table.
* Added Item Properties Table (Save Card Level, Card Exp).

### DIFF
--- a/src/ZoneServer/Network/Send.cs
+++ b/src/ZoneServer/Network/Send.cs
@@ -1971,7 +1971,7 @@ namespace Melia.Zone.Network
 		/// <param name="msg"></param>
 		/// <param name="argNum"></param>
 		/// <param name="argStr"></param>
-		public static void ZC_ADDON_MSG(Character character, string msg, int argNum, string argStr)
+		public static void ZC_ADDON_MSG(Character character, string msg, int argNum = 0, string argStr = "")
 		{
 			var packet = new Packet(Op.ZC_ADDON_MSG);
 
@@ -3604,13 +3604,23 @@ namespace Melia.Zone.Network
 		}
 
 		/// <summary>
-		/// Unknown Purpose.
+		/// Update the client with current cards equipped.
 		/// </summary>
 		/// <param name="character"></param>
 		public static void ZC_EQUIP_CARD_INFO(Character character)
 		{
+			var cards = character.Inventory.GetCards();
+
 			var packet = new Packet(Op.ZC_EQUIP_CARD_INFO);
-			packet.PutShort(0); // Count?
+			packet.PutShort(cards.Count);
+
+			foreach (var card in cards)
+			{
+				packet.PutLpString(card.Value.Data.ClassName);
+				packet.PutLong((long)card.Value.Properties.GetFloat(PropertyName.ItemExp));
+				packet.PutInt(card.Key);
+				packet.PutShort((short)card.Value.Properties.GetFloat(PropertyName.CardLevel));
+			}
 
 			character.Connection.Send(packet);
 		}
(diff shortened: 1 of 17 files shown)


system/scripts/zone/items/exp_cards.cs
```diff
@@ -53,4 +57,91 @@ public DialogTxResult MULTIPLE_USE_XPCARD(Character character, DialogTxArgs args
...
+		}
+
+		var synthesisTime = 5;
+		Send.ZC_ADDON_MSG(character, AddonMessage.ITEM_EXP_START, synthesisTime, "");
+		Send.ZC_CANCEL_MOUSE_MOVE(character);
+
+		if (!character.Components.TryGet<TimeActionComponent>(out var timeAction))
+			return DialogTxResult.Fail;
+
+		var timeActionResult = timeAction.StartAsync("ItemCraftProcess", "None", "UPGRADEGEM", TimeSpan.FromSeconds(synthesisTime)).Result;
```
exectails, line 102: Using `Result` on the task isn't ideal, because by executing this synchronously, you're putting the thread that is handling this request that came in via the network on pause until the time action returns. Granted, it's only a couple seconds, but we want to handle packets as quickly as possible.

I would recommend using `Start` and a callback instead, or create an additional `async Task` method that you call with `CallSafe` from this one,  to more properly implement this asynchronous operation.

src/ZoneServer/Network/Send.cs
```diff
@@ -4474,5 +4484,17 @@ public static void ZC_ACTION_PKS(IActor toActor, IActor fromActor, byte type, in
 
 			toActor.Map.Broadcast(packet);
 		}
+
+		/// Cancel "timed action" due to mouse movement (walking)
```
exectails, line 4488: There seems to be a `<summary>` missing here. Also, could you clarify what exactly this does? The way you're using it, it doesn't seem related to mouse movement, and it's unclear to me what `walking` refers to in this context.

system/scripts/zone/core/normal_tx.cs
```diff
@@ -185,4 +187,87 @@ public NormalTxResult SCR_TX_SKILL_UP(Character character, int[] numArgs)
...
+			return NormalTxResult.Fail;
+		}
+
+		if (!character.Inventory.TryGetItem(itemWorldId, out var card))
+		{
+			Log.Warning("SCR_TX_EQUIP_CARD_SLOT: User '{0}' no card found {1}.", character.Username, argList[0]);
+			return NormalTxResult.Fail;
+		}
+
+		if (character.Inventory.EquipCard(slot, itemWorldId) != InventoryResult.Success)
```
exectails, line 219: Unless I missed it, you don't seem to have a sanity check for the slot, right? This means users could technically send any slot id and equip `int.MaxValue` number of cards.

src/ZoneServer/World/Items/Item.cs
```diff
@@ -97,6 +97,21 @@ public int Amount
 		/// </summary>
 		public Properties Properties { get; } = new Properties("Item");
 
+		/// <summary>
+		/// Gets or sets an expiration date on the item
+		/// </summary>
+		public DateTime ExpirationDate { get; private set; } = DateTime.MaxValue;
```
exectails, line 103: This property doesn't appear to get used yet. Did you include it for future use and because of its use in `IsExpired`, which will never return true in the current code?

exectails (changes requested):
Thanks for the PR. It doesn't look half bad, but I have a few notes.

src/ZoneServer/World/Items/Item.cs
SalmanTKhan: So this was one of those partially cut out of the "card synthesis" code because I'm not sure if we want to go full property on the implementation or not. You can't use an expired material item to upgrade the card. I can remove the expiring item check till we decide what we want to do.

`
{ namespace: "Item", name: "LifeTime", type: "Number", id: 12614 },
{ namespace: "Item", name: "ItemLifeTimeOver", type: "Number", id: 31586 },
{ namespace: "Item", name: "ItemLifeTime", type: "String", id: 31587 },
`
LifeTime defines the duration (in seconds) (or if an item can expire)
ItemLifeTime is the "calculated" date of expiration
ItemLifeTimeOver is a (0 or 1) if the item is expired.

system/scripts/zone/core/normal_tx.cs
SalmanTKhan: I see what you mean I'll add a slot check within EquipCard.
`
if (slot < 1 && slot > 15)
	return InventoryResult.InvalidSlot;
`
Interestingly UnequipCard had the check, so 🤷 guess I forgot to add the check here.

PR #336, opened 2024-11-15
Overbuff Stacking Fix

kenedos:
This PR fixes a bug found in Oblique Fire from Scout, although it also applies to other buffs that provide property modifiers depending on Overbuff stacks.

The bug consists of the fact `AddPropertyModifier()` method in `BuffHandler.cs` is using `Properties.Modify` to change the value of the property, however, the `Modify` method doesn't just set the property to that value, it actually adds it on top of the previous value.

What happens in Oblique Fire is that the first time the buff is applied it will grant +3 movement speed, the second time, however, instead of providing +6 movement speed, it will provide (3 + 6) = 9 movement speed due to the fact the oldValue is being added to the property itself. This causes Oblique Fire to grant more movement speed than it removes on `RemovePropertyModifier`, making players effectively able to gain infinite amounts of movement speed.

### DIFF
--- a/src/ZoneServer/Buffs/Base/BuffHandler.cs
+++ b/src/ZoneServer/Buffs/Base/BuffHandler.cs
@@ -106,7 +106,7 @@ namespace Melia.Zone.Buffs.Base
 				value += oldValue;
 
 			buff.Vars.SetFloat(varName, value);
-			target.Properties.Modify(propertyName, value);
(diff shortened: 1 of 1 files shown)


exectails:
Appreciate the PR and report, but I'm going to do this a tiny bit differently, to better communicate what's happening.

PR #337, opened 2024-11-17
Quarrel Shooter's skills, except Pavise

Terotrous:
Pavise is so complex and affects so many classes that it will have to be a separate PR.

### DIFF
--- a/src/ZoneServer/Buffs/Handlers/Archers/QuarrelShooter/Archer_Kneelingshot.cs
+++ b/src/ZoneServer/Buffs/Handlers/Archers/QuarrelShooter/Archer_Kneelingshot.cs
@@ -1,0 +1,72 @@
+﻿using Melia.Shared.Game.Const;
+using Melia.Zone.Buffs.Base;
+using Melia.Zone.Network;
+using Melia.Zone.Skills;
+using Melia.Zone.Skills.Combat;
+using Melia.Zone.World.Actors;
+using Melia.Zone.World.Actors.Characters;
+using Melia.Zone.World.Actors.Components;
+
+namespace Melia.Zone.Buffs.Handlers.Archers.QuarrelShooter
+{
+	/// <summary>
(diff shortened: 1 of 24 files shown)


exectails:
Thanks for the PR. And it's absolutely fine to put Pavise into a separate PR if it's overly complex, but I looked it up out of curiosity and it appears to be a kind of barrier and potentially "taunt," do I have that right?

> **Deploy Pavise**
> Set up an upright shield to protect yourself from enemy attacks. Enemies without a target will attack the pavise instead.

Based on your wording, I guess the assumption is that we would include special handling in all (ranged?) skills for this? If so, that seems like a maintenance nightmare 🤔 Do you have any more information on the skill and its behavior? Unfortunately I couldn't find any videos exploring how the skill works in detail, but it seems like it could potentially be handled as a buff, modifier, or combat event, where damage is automatically redirected towards the presumed shield monster. Would that seem like a viable implementation?

Terotrous:
The Pavise essentially blocks attacks if it is between you and the source of the attack.  The way I believe this would work is something along the lines of "draw a line from the origin point of the attack's hitbox to the target.  If there is a Pavise in the way, the Pavise blocks the hit".  I had some ideas for how this could feasibly be implemented before but I've kind of forgotten them, maybe I wrote it in the chat history.  I also think that possibly Kenedos may have implemented this skill in Laima.

exectails:
In that case it would have to be incorporated into the target finding I guess. We do not want to have to implement this into every skill handler, and we already have a mostly streamlined method for finding targets. We'll just need to check for potential obstacles between the attacker and the target.

Terotrous:
For sure, I was instead thinking of the skills that would instead bypass this special logic, of which there will unquestionably be some.

exectails:
Do we have any examples of such cases yet? Though if we assume that a handful of skills will simply ignore obstacles, it should be fairly simple to handle these situations on a handler to handler basis, since they ultimately have the control over how they get their targets.

PR #340, opened 2024-12-10
feat: dotnet 9 and nuget packages updates

MUN1Z:
dotnet 9 and nuget packages updates


exectails:
Thank you for your PR. I appreciate the desire to keep Melia up-to-date with the latest .NET versions, though we'll stay on .NET 8 for a while longer, as to not force users to update overly frequently.

PR #341, opened 2025-02-03
Update BC_MESSAGE.bt binary template

celophi:
Adds minor clarification to the packet structure.


exectails:
Thanks for the PR 👍

